PR #1, opened 2018-07-18
Add New supportAction package, JavaActions class

MennaMagedm:
Adding replaceRegex method to JavaActions


MohabMohie-incorta:
Looks very good, Thank you Menna.
In the future we can overload this method and create replaceRegex (String text), which will have a list of predefined special characters which we can enrich as needed, so that the user wouldn't need to define the special characters himself
sample: []\^$.|?*+(){}

MennaMagedm:
Sound good (Y) next time isA :)

PR #4, opened 2018-07-19
Add overloaded method to replaceRegex()

MennaMagedm:
Implementing replaceRegex with one String parameter

### DIFF
--- a/src/main/java/com/shaftEngine/supportActionLibrary/JavaActions.java
+++ b/src/main/java/com/shaftEngine/supportActionLibrary/JavaActions.java
@@ -3,13 +3,34 @@ package com.shaftEngine.supportActionLibrary;
 public class JavaActions {
 	
 /**
- * replaceRegex method takes two parameters, one is the array of special characters that are needed to be replaced, and the text needed to be updated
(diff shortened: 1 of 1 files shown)


src/main/java/com/shaftEngine/supportActionLibrary/JavaActions.java
```diff
@@ -19,4 +22,22 @@ public static String replaceRegex(String specialCharactersArray[], String text)
...
+	
+	/**
+	 * Returns text after replaces its regular expressions which included in this set []^$.|?*+(){}
+	 * @param text
+	 * @return updated text after escaping its regular expressions
+	 */
+	
+	public static String replaceRegex(String text) {
+		String specialCharactersArray[]= {"[","]","^","$",".","|","?","*","+","(",")","{","}"};
+		String oldChar;
```
MohabMohie: here we should call
replaceRegex(specialCharactersArray[], text);
and the other function will do the loop.

MennaMagedm: I got it (Y)

MennaMagedm: comment is applied

PR #12, opened 2018-10-07
Add assertResponseContains to ResActions

MennaMagedm:


src/main/java/com/shaft/api/RestActions.java
```diff
@@ -304,6 +305,49 @@ public String getResponseJSONValue(Response response, String jsonPath) {
...
 
+	/**
+	 * Check that given response body contains certain key, it transform response
+	 * body to string & assert if It contains this value
+	 * 
+	 * @param response
+	 * @param key
+	 */
+
+	public void assertResponseContains(Response response, String key) {
```
MohabMohie-incorta, line 316: can we discuss these functions?
I believe that we shouldn't have assertions here, as we already handle them in the validations package, that's why I removed the functions that did the assertions before, and only kept the functions that extracted from the resulting body. So that the user can use the validations package at a higher level (POM class).

MennaMagedm: As per our meeting, I will close this PR & make some changes.

PR #37, opened 2019-06-17
Incorta

MennaMagedm:
- Create DBActions base class with following methods:

1. createConnection
2. executeSelectQuery
3. closeConnection

- Add mysql, sql server, postgresql dependencies to pom.xml
- Create Test class "Test_dbConnection.java"


src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+
+	Connection connection;
+	Statement statement;
+	ResultSet resultSet;
+	String dbType;
+	String dbServerIP;
+	String dbPort;
+	String dbName;
+	String username;
+	String password;
```
MohabMohie: it is a must that all the variables are private

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+		failAction(actionName, testData, null);
+	}
+
+	/**
+	 * Setup connection to database
+	 * 
+	 * @throws ClassNotFoundException
+	 * @throws SQLException
+	 */
+	public void createConnection() {
```
MohabMohie: it is a must for this method to be private, and to be called automatically at the start of the executeSelectQuery() method. and the javadocs should be updated as both the exceptions are now handled and not thrown

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+import org.testng.Assert;
+
+import com.shaft.io.ReportManager;
+
+public class DBActions {
+
+	Connection connection;
+	Statement statement;
+	ResultSet resultSet;
+	String dbType;
```
MohabMohie: it is recommended that the dbType should be an enum with a list of the supported databases

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+	String dbServerIP;
+	String dbPort;
+	String dbName;
+	String username;
+	String password;
+
+	public DBActions(String dbType, String dbServerIP, String dbPort, String dbName, String username, String password) {
```
MohabMohie: it is a must that the constructor docs should have more details for the supported database types and description (with example) for each of the variables.

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+	 * Setup connection to database
+	 * 
+	 * @throws ClassNotFoundException
+	 * @throws SQLException
+	 */
+	public void createConnection() {
+		String testData = dbServerIP + ":" + dbPort + "/" + dbName;
+
+		try {
+			switch (dbType) {
```
MohabMohie: it is recommended to add .toLowerCase() and .trim() when switching on a string to reduce the risk of human error

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+				Class.forName("org.postgresql.Driver");
+				connection = DriverManager.getConnection("jdbc:postgresql://" + dbServerIP + ":" + dbPort + "/" + dbName,
+						username, password);
+				break;
+			// create connection to oracle DB
+			case ("oracle"):
+				break;
+			}
+
+			statement = connection.createStatement();
```
MohabMohie: it is recommended to customize and expose some connection configuration parameters in the properties files. the first of which is the connection timeout parameter.

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	/**
+	 * execute DB query and return the result as ResultSet object
+	 * 
+	 * @param dbQuery
+	 * @return ResultSet
+	 * @throws SQLException
+	 */
+	public ResultSet executeSelectQuery(String dbQuery) {
```
MohabMohie: it is a must to return a regular String to be consistent with the rest of the methods that do not require further processing of the returned data via special methods.

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+		try {
+			resultSet = statement.executeQuery(dbQuery);
+			passAction("Execute DB query", dbQuery);
+		}
+
+		catch (SQLException e) {
+			ReportManager.log(e);
+		}
+		return resultSet;
```
MohabMohie: it is a must for this return value to be attached as an attachment/step in the report

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+	 * @throws SQLException
+	 */
+	public ResultSet executeSelectQuery(String dbQuery) {
+		try {
+			resultSet = statement.executeQuery(dbQuery);
+			passAction("Execute DB query", dbQuery);
+		}
+
+		catch (SQLException e) {
+			ReportManager.log(e);
```
MohabMohie: it is a must to fail the action when catching any exception that will break the execution.
The same applies for the other catch statements in this class.

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+		}
+		return resultSet;
+	}
+
+	/**
+	 * Close database connection
+	 */
+	public void closeTheConnection() {
```
MohabMohie: it is a must for this method to be private, and to be called automatically at the end of the executeSelectQuery() method.

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+		if (resultSet != null) {
+			try {
+				resultSet.close();
+			} catch (Exception e) {
```
MohabMohie: it is a must to never catch a generic exception object, or throwable object. if a certain exception is thrown then it should be handled. else nothing should be catched.
The same applies for the other catch statements in this class.

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+	 * @throws SQLException
+	 */
+	public void createConnection() {
+		String testData = dbServerIP + ":" + dbPort + "/" + dbName;
+
+		try {
+			switch (dbType) {
+			// create connection to mysql DB
+			case ("mysql"):
+				Class.forName("com.mysql.jdbc.Driver");
```
MohabMohie: I understand that this line can sometimes be used as a workaround for some broken java implementation, and that sometimes it has a .newInstance(); at the end of it. Please explain this case via in-code comments as it currently looks like an unneeded call for a class that is never used.

src/main/java/com/shaft/support/DBActions.java
```diff
@@ -0,0 +1,155 @@
...
+	public DBActions(String dbType, String dbServerIP, String dbPort, String dbName, String username, String password) {
+		this.dbType = dbType;
+		this.dbServerIP = dbServerIP;
+		this.dbPort = dbPort;
+		this.dbName = dbName;
+		this.username = username;
+		this.password = password;
+	}
+
+	private void passAction(String actionName, String testData, String log) {
```
MohabMohie: this method is the one that should be called after the query runs successfully, and the result string from the query execution should be passed to the log object.

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: It is added in executeSelectQuery but I passed the query for it instead of result because in this layer the result is object of ResultSet not string, thoughts ?

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: I don't understand what is the workaround for some broken java implementation. could you explain more please?

this line is used to specify to DriverManager which JDBC drivers to try to make Connections with.

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MohabMohie: This line is not mandatory, as per the below docs:
https://docs.oracle.com/javase/8/docs/api/java/sql/DriverManager.html#getConnection-java.lang.String-

Instead, you should use the register method, and then simply get connection.

Here is a sample for this discussion:
https://stackoverflow.com/questions/5484227/jdbc-class-forname-vs-drivermanager-registerdriver

This line is used to load a class that was not loaded, and this is a workaround in case the regular register / connect approach fails for some reason. But it's not the best practice.

I assumed that you faced this issue and therefor used this workaround, and therefor asked that you document the issue that you faced.

In case there was no issue using the regular register then get connection, please refactor the code accordingly.

src/main/java/com/shaft/support/DBActions.java
MohabMohie: in executeSelectQuery, the method used is this one [passAction("Execute DB query", dbQuery);] <- 2 parameters

I'm asking you to use this one instead [passAction("Execute DB query", dbQuery, resultSet);] <- 3 parameters

There's another comment advising that the returned result set should be parsed to string and not returned as is. In which case you can parse it, report that the step has passed, and then return the string.

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: This will be handled in POM class because the query result is not returned as string, you have to make some processing of ResultSet object to get your data which will be implemented on another layer as explained in the meeting

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: let's discuss this tomorrow insha'Allah to get more details about the request

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: let's discuss this tomorrow insha'Allah to get more details about the request

src/main/java/com/shaft/support/DBActions.java
MohabMohie: Please plan a meeting and let's discuss any missing points.
For this point, no extra processing should be required after using any SHAFT_Engine method. In the APIs we return the rest object, as SHAFT itself is used to process it and extract json or xml values from it. So if further processing is needed, helping methods that extract the values should be implemented in SHAFT and then used in the ObjectModel class. If not then SHAFT should return a string. Eitherway SHAFT should make an attachment with the returned result from the action. The same happens in the GUI/CLI/API sections.

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: Done, as per first link, I removed the method and I didn't face any issue:

Class.forName() is not directly related to JDBC at all. It simply loads a class.
Most JDBC Driver classes register themselves in their static initializers by calling registerDriver().
registerDriver() is the real call that you hardly ever need to call yourself (unless you write your own JDBC driver).

src/main/java/com/shaft/support/DBActions.java
MennaMagedm: since it is recommended, let's defer this to enhancement phase

src/main/java/com/shaft/support/DBActions.java
MohabMohie: to-be-done in a future commit as agreed

src/main/java/com/shaft/support/DBActions.java
MohabMohie: (Y)

src/main/java/com/shaft/support/DBActions.java
MohabMohie: agreed

src/main/java/com/shaft/support/DBActions.java
MohabMohie: will reopen comment on the latest commit.

src/main/java/com/shaft/support/DBActions.java
MohabMohie: will reopen on the latest commit.

src/main/java/com/shaft/support/DBActions.java
MohabMohie: will reopen on the latest commit.

src/main/java/com/shaft/database/DBActions.java
```diff
@@ -0,0 +1,190 @@
...
+ * @author mennamaged
+ *
+ */
+
+public class DBActions {
+
+	private Connection connection;
+	private Statement statement;
+	private ResultSet resultSet;
+	private String dbType;
```
MohabMohie, line 24: recommended to be enum, please open a jira ticket for this as a future enhancement and reply with the link.

src/main/java/com/shaft/database/DBActions.java
```diff
@@ -0,0 +1,190 @@
...
+				connectionString = "jdbc:postgresql://" + dbServerIP + ":" + dbPort + "/" + dbName;
+				break;
+
+			default:
+				ReportManager.log("Database not supported");
+				failAction("createConnection", dbType);
+				break;
+			}
+			connection = DriverManager.getConnection(connectionString, username, password);
+			passAction("createConnection", connectionString);
```
MohabMohie: it is a must to replace this "passAction" with discrete logging as discussed.

src/main/java/com/shaft/database/DBActions.java
```diff
@@ -0,0 +1,190 @@
...
+	/**
+	 * prepare statement for database connection to perform execute query
+	 * 
+	 * 
+	 */
+	
+	private void createStatement() {
+		try {
+			statement = connection.createStatement();
+			passAction("createStatement", connection.toString());
```
MohabMohie: it is a must to replace this "passAction" with discrete logging as discussed.

src/main/java/com/shaft/database/DBActions.java
```diff
@@ -0,0 +1,190 @@
...
+	public void closeConnection() {
+		try {
+			if (resultSet != null)
+				resultSet.close();
+			if (statement != null)
+				statement.close();
+			if (connection != null)
+				connection.close();
+			ReportManager.logDiscrete("closeConnection");
+			passAction("closeConnection", connection.toString());
```
MohabMohie: ReportManager.logDiscrete("closeConnection"); -> This is a log message so it should be more descriptive. For example it should say clearly that the connection has been closed successfully.
passAction("closeConnection", connection.toString()); -> This line should be removed.

src/main/java/com/shaft/database/DBActions.java
```diff
@@ -0,0 +1,190 @@
...
+	 * 
+	 * return ResultSet
+	 */
+	public ResultSet executeSelectQuery(String dbQuery) {
+		ResultSet resultSet = null;
+		try {
+			createConnection();
+			createStatement();
+			resultSet = statement.executeQuery(dbQuery);
+			passAction("executeSelectQuery", dbQuery);
```
MohabMohie, line 163: as discussed, it is a must to call the other method
	private void passAction(String actionName, String testData, String log) {
and find a way to return the ResultSet as a string so that it gets attached to the Report.

src/main/java/com/shaft/database/DBActions.java
```diff
@@ -0,0 +1,190 @@
...
+		} catch (SQLException e) {
+			ReportManager.log(e);
+			failAction("executeSelectQuery", dbQuery);
+		}
+		return resultSet;
```
MohabMohie, line 168: as discussed, it is a must to implement one of those two options here:
1. return a string value so that the assertions library could be used.
2. provide more methods to process the ResultSet object and extract string data from it, so that the assertions library could be used.

src/main/java/com/shaft/database/DBActions.java
MennaMagedm: done

src/main/java/com/shaft/database/DBActions.java
MennaMagedm: done

src/main/java/com/shaft/database/DBActions.java
MennaMagedm: done

MennaMagedm: but I got confused, as I understood ReportManager.logDiscrete("closeConnection") is used when method is private and this one is public (will discuss later why it is public not private)

src/main/java/com/shaft/database/DBActions.java
MennaMagedm: let's discuss this in the meeting isA

src/main/java/com/shaft/database/DBActions.java
MennaMagedm: let's discuss this in the meeting isA

src/main/java/com/shaft/database/DBActions.java
MennaMagedm: Done :
https://incorta.atlassian.net/browse/AUTO-32

src/main/java/com/shaft/database/DBActions.java
MohabMohie: as agreed, I will develop this part after merging your contribution to a separate feature branch.

src/main/java/com/shaft/database/DBActions.java
MohabMohie: as agreed, I will develop this part after merging your contribution to a separate feature branch.

MohabMohie:
as agreed, I will close this PR and create a separate feature branch for your contribution.

PR #41, opened 2019-07-01
Add retrieve methods and remove resultset from executeSelectQuery return

MennaMagedm:
Apply review comments of last PR:

- Remove ResultSet as return from executeSelectQuery
- Add two methods (retrieve single value, retrieve string list) to get data from resultSet
- Update Test_dbConnection.java to use new methods


MohabMohie:
We agreed to meet on Tuesday to revisit the output from the pair programming phase and plan on the next actions. I expected the meeting to happen before new changes are implemented. Kindly let's meet first to discuss.

PR #82, opened 2020-04-17
bug fixes and enhancements

### LINKED ISSUE #81: add list is not empty built in validation before attempting to select from a dropdown menu
**Is your feature request related to a problem? Please describe.**
When the user tries to select from a menu that has no options, the select action fails with no clear error.

**Describe the solution you'd like**
Checking (or waiting) until the list text is not empty should help solve that problem.

MohabMohie:
.ElementActions:
..added wait for element text to not be empty in case of selecting from a dropdown menu. resolves #81 

.execution.properties:
..changed openAllureReportAfterExecution to false until a fix is found for the execution remaining hanged

.ReportManager:
..fixed issue with the allure binaries being redownloaded while they are already present in the local maven repository

.pom.xml
..updated some dependencies


MohabMohie:
test failed due to TestNG 7.1.0 issue
@BeforeGroups(groups = {"browserBasedTests"})

try upgrading to TestNG 7.2.0 and re-execute.

PR #113, opened 2020-05-21
4.2.x

### LINKED ISSUE #112: allure report archive is broken (doesn't contain the allure binaries)
**Describe the bug**
When the user enables generateAllureReportArchive the generated archive doesn't contain the allure binaries and hence is not usable.

MohabMohie:
fixes #112


MohabMohie:
will ignore the current test failure due to SO_TIMEOUT on the Grid

PR #132, opened 2020-06-13
Fix isElementClickable issue #131

MamdouhMansour:
Fixed Cleaning allure results before execution and is Element clickable.

Regards cleaning allure reports before execution, only tested on Windows OS and it needs to terminate the running test first before executing it again from the IDE.

Issue #131, opened 2020-06-13
ElementActions-Is Element Clickable fails action and don't return false || pass the action while the element is not enabled .


MamdouhMansour:
**Describe the bug**
When using isElementClickable method of ElementAction class in case of element is not enabled or not clickable it fails action and does not return false value or pass the action while the element is not enabled so web driver waits to element to be clickable where it will not as it is disabled as per attached.

**To Reproduce**
1. 	private WebDriver driver;

	private By firstName = By.id("firstname");
	private By lastName = By.id("lastname");
	private By emailAddress = By.id("email");
	private By password = By.id("password");
	private By registerButton = By.xpath("//button[contains(@class, 'register')]");
	private By checkBox1 = By.xpath("//label[@for='checkbox1']");
       
       @Test
	public void testIsButtonNotClickable() {
		driver = BrowserFactory.getBrowser();
		BrowserActions.navigateToURL(driver, "https://www.gocardi.com/register");		
		ElementActions.type(driver, firstName, "nameOne");
		ElementActions.type(driver, lastName, "nameTwo");
		ElementActions.type(driver, emailAddress, "test@gmail.com");
		ElementActions.type(driver, password, "test123456");
		Assertions.assertTrue(ElementActions.isElementClickable(driver, registerButton), AssertionType.NEGATIVE);
	}
	
2. See error

**Expected behavior**
isElementClickable returns false if element is disabled or not clicakble.

**Screenshots**
![IsELementClickable_Bug](https://user-images.githubusercontent.com/39811691/84573114-d9e01200-ad9e-11ea-8280-01b52aef80c0.PNG)

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - Browser [chrome, firefox]
 - Version [83, 77]

MohabMohie:
Fixed by @MamdouhMansour

PR #136, opened 2020-06-15
Fixes isElementClickable #131 

MamdouhMansour:
Applied CR of #132


src/main/java/com/shaft/tools/io/ReportManager.java
```diff
@@ -692,6 +693,23 @@ public static void openAllureReportAfterExecution() {
             new TerminalActions().performTerminalCommand(commandToOpenAllureReport);
         }
     }
+    
+    // TO close running reports so allure results can be cleaned before execution
+ 	// That will work after running test terminated from IDE first.
+ 	public static void closeRunningAllureReport() {
```
MohabMohie: please remove this code from this PR as it is related to another issue and let's discuss this proposed fix more before we implement it.

MamdouhMansour: Ok great, You are more than wlecome :).

MamdouhMansour:
Thank you for such amazing project and always happy to contribute.

Issue #244, opened 2021-01-27
I request a feature to add capabilities instead of overriding shaft engine's


ShorouqElrakhawy:
**Is your feature request related to a problem? Please describe.**
The problem is I want to add a capability that disables SSL Certificates without overriding the shaft engine's
in the future, I may want to add more capabilities 

**Describe the solution you'd like**
A method I could use to add a capability without affecting shaft engine's

**Describe alternatives you've considered**
Covering all existing capabilities in Shaft properties files

MahmoudElSharkawy:
Created a PR for a hot fix. Let's wait the review :D 
https://github.com/MohabMohie/SHAFT_ENGINE/pull/245

MohabMohie:
Hotfix merged, awaiting minor investigation and then release.

PR #245, opened 2021-01-27
Hot Fix for Custom Browser Options (#244)

MahmoudElSharkawy:
This is a hot fix for Custom Browser Options for the following ticket (#244) to enable adding browser option(s) while still using the default options of SHAFT.
As this is a hot fix, there is a known issue that when adding option(s) that's already included in the default options of SHAFT, the SHAFT default option(s) will override the custom one(s) of the same type.


MahmoudElSharkawy:
Added a feature which the user can choose to use/append to the default browsers options or not by adding a boolean appendToDefaultOptions parameter to the getBrowser() methods and updated the setDriverOptions() method accordingly.

ShorouqElrakhawy:
@MohabMohie May I know the available time for this PR to be merged?

src/main/java/com/shaft/gui/browser/BrowserFactory.java
```diff
@@ -317,16 +321,19 @@ private static void setLoggingPrefrences() {
         logPrefs.enable(LogType.DRIVER, Level.ALL);
     }
 
-    private static void setDriverOptions(String browserName, MutableCapabilities customBrowserOptions) {
+    private static void setDriverOptions(String browserName, MutableCapabilities customBrowserOptions, Boolean appendToDefault) {
```
MohabMohie: I think this boolean variable is redundant, I'll share my detailed feedback offline

MohabMohie:
@ShorouqElrakhawy it's still ongoing, please hold.

MohabMohie (approved):

MohabMohie:
Merged with known issue:
Ordering for options is reversed due to bug in Selenium WebDriver options.merge() method which drops a lot if the settings.

For now if you define a custom property it will be respected.
But if it was already set by the default engine configuration it will be ignored and the default will supercede.

MohabMohie:
@MahmoudElSharkawy please follow this commit for the bug fix from Selenium WebDriver side

https://github.com/SeleniumHQ/selenium/commit/ec9c67b2fb1ec110f51934832fec3cef8a9bf746

Issue #249, opened 2021-02-04
[JSONFileManager] the reader gets empty each time you use the method getTestData() and we need to initialize it each time we use methods in the class


magdyheiba:
**Describe the bug**
While using JSONFileManager the reader gets empty each time you use the method getTestData() and we need to initialize it each time we use methods in the class
**To Reproduce**

1. initialize the class `JSONFileManager  JSONFileManager  = new JSONFileManager(FilePath);`
2. `JSONFileManager.getTestData("JSON path");`
3. another `JSONFileManager.getTestData("JSON path");`

**Expected behavior**
- The First line returns the value of JSONPath 
- The second line returns null although the JSONPath is correct 
**Screenshots**

**Desktop (please complete the following information):**
 - OS: [Windows10]
 - Browser [chrome]
<img width="591" alt="19" src="https://user-images.githubusercontent.com/38945724/106911381-c0635000-670a-11eb-825d-1826e8fa65e7.PNG">
<img width="564" alt="18" src="https://user-images.githubusercontent.com/38945724/106911383-c0fbe680-670a-11eb-954d-ebbe00e9c228.PNG">

**Additional context**

MohabMohie:
Thanks for reporting and fixing this issue @magdyheiba 🙏

PR #250, opened 2021-02-05
Bug fix#249

magdyheiba:
Fix Bug #249 
Now when we use JSONFileManager we don't need to create an object from the class each time we use it

After Fix :- 

Thanks

### DIFF
--- a/src/main/java/com/shaft/tools/io/JSONFileManager.java
+++ b/src/main/java/com/shaft/tools/io/JSONFileManager.java
@@ -112,6 +112,14 @@ public class JSONFileManager {
      */
     private Object getTestData(String jsonPath, DataType dataType) {
         Object testData = null;
+        FileReader reader = null;
+		try {
+			reader = new FileReader(jsonFilePath);
+		} catch (FileNotFoundException rootCauseException) {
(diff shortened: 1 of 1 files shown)


MohabMohie (changes requested):
Hi @magdyheiba it seems that the entire file was changed. This usually happens if you use automatic styling.
Please change only the needed lines, and don't use automated styling (Ctrl+Shift+F) on the entire file.

src/main/java/com/shaft/tools/io/JSONFileManager.java
```diff
@@ -13,124 +13,141 @@
...
-    }
+	/**
+	 * Creates a new instance of the test data json reader using the target json
+	 * file path
+	 *
+	 * @param jsonFilePath target test data json file path
+	 */
+	public JSONFileManager(String jsonFilePath) {
+		this.jsonFilePath = jsonFilePath;
+		try {
```
MohabMohie: let's move this entire try/catch block to a private method "initializeReader" so that we can reuse it below instead of duplicating the code.

MohabMohie (approved):
Thank you @magdyheiba 🙏

PR #262, opened 2021-02-18
Extracted all wait-related methods from ElementActions

MohabMohie:
Signed-off-by: Mohab Mohie <Mohab.MohieElDeen@outlook.com>


MohabMohie:
rejected.

Issue #76, opened 2020-04-10
Use enums in the validateElementAttribute method in the ValidationActions class


MahmoudElSharkawy:
### **The proposal**
Make the SHAFT_Engine more user friendly with it's verification methods.
Make use of the enums in the **validateElementAttribute** method in the **ValidationActions** class and Implement an overloaded methods to take these enums for the common attributes or a string for any other needed attribute.

![image](https://user-images.githubusercontent.com/46620378/78984881-a7f4ca00-7b27-11ea-9ba9-93bc2443c90a.png)

PR #298, opened 2021-04-24
Added ElementAttributeType enums for the ValidateElementAttribute methods (#76)

MahmoudElSharkawy:
**Changes**: 
* public enum ElementAttributeType in both Assertions and Verifications classes
*  Implement an overloaded methods to take these enums for the common attributes or a string for any other needed attribute.
* Added a unit testing case to test the ElementAttributeType method.


MohabMohie (approved):
Thank you for your valued contributions 🙏

Issue #299, opened 2021-04-24
Make use of enums for the validateBrowserAttribute method


MahmoudElSharkawy:
**The proposal**

Make the SHAFT_Engine more user friendly with it's validations methods.
Make use of the enums in the validateBrowserAttribute method

MohabMohie:
Closed by #300

PR #300, opened 2021-04-24
Add enums for the verify/assertBrowserAttribute method #299

MahmoudElSharkawy:
**Changes**
* Added public enum BrowserAttributeType in both Assertions and Verifications classes
* Implemented an overloaded methods to make use of these enums attribute and left the old methods unchanged.


MohabMohie (approved):
Thank you @MahmoudElSharkawy 
Please add the unit tests at a later stage 👍

PR #316, opened 2021-05-10
feat: add assertApiResponseEquals and verifyApiResponseEquals methods

hammad101088:
two assertion methods that can be used for all API Response actions. <br/> add assertApiResponseEquals in Assertions Class and verifyApiResponseEquals in Verifications Class


MohabMohie (changes requested):
There's an issue with this PR. I think you're reformatting the entire file, and hence it's showing me that thousands of lines of code are changed, therefore I can't review the changes.

Kindly do not reformat the file. Just add the new lines of code and don't worry about the formatting.

PR #352, opened 2021-08-11
Adding pinchToZoom method

samiir95:
pinchToZoomAllureReport

### DIFF
--- a/src/main/java/com/shaft/gui/element/TouchActions.java
+++ b/src/main/java/com/shaft/gui/element/TouchActions.java
@@ -636,7 +636,7 @@ public class TouchActions {
     	} catch(Exception rootCauseException) {
             WebDriverElementActions.failAction(driver, null, rootCauseException);
         }
-        WebDriverElementActions.passAction(driver, null);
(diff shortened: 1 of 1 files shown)


MohabMohie (changes requested):
Everything looks great thank you!

Here are my comments, all of which can be implemented in less than 15 minutes I think:
1. We need the return of the public method to be this and not void so that the fluent design doesn't get broken.
2. We need to report pass/fail status.
3. The variable name zoom, needs to be renamed to have the same name as the enum but starting with a lower case.
4. We need to add javadocs for the public method.

Here's an example for your reference for all the above points.
https://github.com/MohabMohie/SHAFT_ENGINE/blob/1dd001c9c6e6f621f3ee8f9cfac8849d1ad91a47/src/main/java/com/shaft/gui/element/TouchActions.java#L52

MohabMohie (approved):

MohabMohie:
Thanks a lot, I'm very grateful and I hope to see many more excellent contributions from you 🙏

PR #386, opened 2021-09-15
Implement PDF Validation of Downloaded PDF File

MamdouhMansour:
- Added PdfFileManager that validate the downloaded pdf file
- The implementation includes waiting for the file to be successfully downloaded, which will be determined by user
- Also included enum to check if the pdf file will be deleted or not after done getting the document text
- Added test sample pdf file that added in test data files


pom.xml
```diff
@@ -1,4 +1,5 @@
-<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
MohabMohie: Kindly do not auto format the file from eclipse IDE as it registers false changes

src/main/java/com/shaft/tools/io/PdfFileManager.java
```diff
@@ -0,0 +1,173 @@
...
+	 * 
+	 * @param url                the path of the downloaded pdf file (usually will
+	 *                           be target/downloadedFiles/your-file.pdf), usually
+	 *                           will be in "target/downloadedFiles/" for the
+	 *                           project
+	 * @param milliSecondsToWait the time in milliseconds needed to wait on the
+	 *                           document to be downloaded successfully to the
+	 *                           target folder path
+	 */
+	public PdfFileManager(String url, int milliSecondsToWait) {
```
MohabMohie: Usually timeouts are added to the property files and not passed by the user, and it's unified across the engine that timeouts are in seconds, so we can internally convert the number in the property file from seconds to milliseconds

src/main/java/com/shaft/tools/io/PdfFileManager.java
```diff
@@ -0,0 +1,173 @@
...
+	/**
+	 * @param startPageNumber                 the starting page for the document to
+	 *                                        be validated
+	 * @param endPageNumber                   the ending page for the document to be
+	 *                                        validated
+	 * @param deleteFileAfterValidationStatus the status of deleting the file after
+	 *                                        get the document text
+	 * @return returns the pdf content in string so that can be validated
+	 */
+	public String readPDFContentFromDownloadedPDF(int startPageNumber, int endPageNumber,
```
MohabMohie, line 64: Let's overload this method with one that takes no values from the user and just reads the entire pdf file

src/main/java/com/shaft/tools/io/PdfFileManager.java
```diff
@@ -0,0 +1,173 @@
...
+		parser = parseStreamDocument(stream);
+
+		cosDoc = getParsedDocument(parser);
+		String content = getPdfText(cosDoc, startPageNumber, endPageNumber);
+		closeStreamAndDeleteFile(file, stream, deleteFileAfterValidationStatus);
+
+		return content;
+	}
+
+	private void waitOnDocumentToBeReady(int milliSeconds) {
```
MohabMohie: Better than thread.sleeping for the entire duration, there are methods in the FileActions library that checks if a file exists.
You can have a polling interval, and go check every 5 seconds until the file exists or the timeout is reached.
This is how the selenium fluent wait works, but we'll implement it from scratch.
Better to check and see if there's already a method in SHAFT that waits until a file exists already.

And as agreed, timeouts will be provided from the property file and not as a method parameter

src/main/java/com/shaft/tools/io/PdfFileManager.java
```diff
@@ -0,0 +1,173 @@
...
+				ReportManager.log("Couldn't find the file, File directory may be null or file is not found.");
+				Assert.fail("Couldn't find the file, File directory may be null or file is not found.");
+			}
+		default:
+			break;
+		}
+
+	}
+
+	// Will implement same for preview pdf file
```
MohabMohie, line 208: Add a todo to create a new validation inside the file validations wizard that compares the content of the downloaded file, versus the content of a test data file and attaches both of them in the report.

I can do that one to complement the feature you're working on. Or I can help you do it if you prefer.

src/test/java/testPackage01/Test_PDF.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class Test_PDF {
+
+	@Test
+	public void testPDF_Downloaded() {
+		String url = FileActions.getAbsolutePath("src/test/resources/TestDataFiles/", "sample.pdf");
+
+		System.out.println(url);
+
+		String content = new PdfFileManager(url, 7000).readPDFContentFromDownloadedPDF(1, 1,
```
MohabMohie: We can make this method accept relative path and then do an internal call to convert it to absolute path. That would provide easier user experience

MohabMohie (changes requested):
This looks great! Thanks a lot for your valuable contributions 🙏

Kindly address the feedback points and let me know if you need any support 👍

src/main/java/com/shaft/tools/io/PdfFileManager.java
MamdouhMansour: @MohabMohie 
can we do it like this in FileActions?

if (Boolean.FALSE.equals(doesFileExit)) {
				try {
					Thread.sleep(TimeUnit.SECONDS.toMillis(Long.parseLong(System.getProperty("fileTimeout"))));
				} catch (Exception e1) {
					ReportManagerHelper.log(e1);
				}
			}

src/main/java/com/shaft/tools/io/PdfFileManager.java
MamdouhMansour: @MohabMohie yes please if you can help in that, I'd like to do it.

src/main/java/com/shaft/tools/io/PdfFileManager.java
MohabMohie: this is still one thread.sleep for the entire timeout duration. but we need to have it check every 5 seconds until it finds the file or the timeout is reached.

src/main/java/com/shaft/tools/io/PdfFileManager.java
MohabMohie: @MamdouhMansour great, let's finalize this PR and then we can open a new one for the validation that will use this method (Y)

MamdouhMansour: @MohabMohie I am done and waiting for your comments, thank you for the support as usual

src/main/java/com/shaft/tools/io/PdfFileManager.java
```diff
@@ -0,0 +1,209 @@
...
+
+		cosDoc = getParsedDocument(parser);
+		String content = getPdfText(cosDoc, startPageNumber, endPageNumber);
+		closeStreamAndDeleteFile(file, stream, deleteFileAfterValidationStatus);
+
+		return content;
+	}
+
+	public String readPDFContentFromDownloadedPDF(DeleteFileAfterValidationStatus deleteFileAfterValidationStatus) {
+
```
MohabMohie, line 78: this method should have no parameters, and should call the above method and pass the default values to it. it shouldn't delete the file after validating it by default.

MohabMohie (approved):
thank you for your valued contribution!

enhancement points for future PRs: adding the pdf validation method which will attach both the expected and actual pdfs

PR #418, opened 2021-10-28
Match json schema

samiir95:


samiir95:
I don't know why this error is appeared although I've updated this below
version from 3.0.0 to  4.4.0 pkg:maven/io.rest-assured/json-schema
***@***.***

MohabMohie (approved):
This is an excellent contribution! Thank you :)
for the future kindly consider adding at least one test to show how the new method should look.

samiir95:
> This is an excellent contribution! Thank you :) for the future kindly consider adding at least one test to show how the new method should look.

Okay and here's **the whole test** that I created to test this method. Thank you for your continuous support <3 
image

MohabMohie:
Well since you have it already, kindly create a PR to merge it in to the related src/test/java package.

Those tests run with every new change as user journey tests for the engine to ensure that nothing breaks... So it's very healthy to have tests for every new feature.

samiir95:
> Well since you have it already, kindly create a PR to merge it in to the related src/test/java package.
> 
> Those tests run with every new change as user journey tests for the engine to ensure that nothing breaks... So it's very healthy to have tests for every new feature.

I've uploaded a new pull request for it, Kindly check it. Thank you!

PR #449, opened 2021-11-15
Enhance typeFileLocationForUpload To Auto Detect Relative Path

MamdouhMansour:
- Add a condition to check if the relative path starts with "src"
- Remove deprecated methods from Test_uploadFile test class


MohabMohie (approved):
Thank you for your valuable (and perfect) contribution

MamdouhMansour:
@MohabMohie Always happy to contribute

PR #466, opened 2021-12-06
removed 3 props. from mangeMaxPerformanceMode class

themosaeed:
removed the following 

System.setProperty("forceCheckForElementVisibility", String.valueOf(false));
System.setProperty("forceCheckElementLocatorIsUnique", String.valueOf(false));
System.setProperty("forceCheckTextWasTypedCorrectly", String.valueOf(false));


MohabMohie (approved):
Thank you for your valued contribution!

PR #491, opened 2022-01-10
Feature#490

magdyheiba:
image
image
image


src/main/resources/defaultProperties/ExecutionPlatform.properties
```diff
@@ -1,10 +1,10 @@
...
+###################################################
+##### SHAFT_Engine: ExecutionPlatform.properties
+###################################################
+executionAddress=local
+targetOperatingSystem=Linux-64
+targetBrowserName=GoogleChrome
+headlessExecution=false
+###################################################
+##### END of Properties File
 ###################################################
```
MohabMohie: Looks like something is Wrong in this file line breaks or something... Would you roll it back please.
The rest is perfect! Thank you 🙏🏻

MohabMohie (changes requested):
Thsnk you for your valued contribution 🙏🏻🙏🏻

src/main/resources/defaultProperties/ExecutionPlatform.properties
magdyheiba: Replaced with the one on master and pushed

src/main/java/com/shaft/tools/io/ReportManagerHelper.java
```diff
@@ -61,6 +61,12 @@
     private static ExtentTest extentTest;
     private static String extentReportFileName;
     private static String generateExtentReports;
+	private static final String jdkExtractionLocation = System.getProperty("user.home") + File.separator + ".jdks"
```
MohabMohie: I think we need to remove this variable and stick with the system property like we discussed for both windows and linux/mac... What do you think?

MohabMohie (commented):
Need your input regarding the user.dir variable... I think we should remove it.

src/main/java/com/shaft/tools/io/ReportManagerHelper.java
magdyheiba: At windows intellij downloads the jdks at this path that's why I hardcoded it but at Mac it downloads on the default directory of system JDK which I can get without system variables through this command "$(/usr/libexec/java_home)"

- If the system variable of JAVA on Windows is not the latest it won't run and it will print the system JDK path when I use "java.home" that's why I hardcode the path and use the system variable of "java.runtime.version" and substring it to extract the version number

MohabMohie: What happens in case of an eclipse IDE user?

magdyheiba: Yeah it will cause an issue as the path will not be existed 
At eclipse you will have to download it manually because eclipse isn't downloading the JDK automatically as far as I know

magdyheiba: I don't want to point to the system variable at windows specifically as at our case we don't have an access to "java.home" and most of the time it may be old jdk

MohabMohie: Aywa... Ngebo e7na ezay ba2a fl case de?
We need a property that gets us whatever the java that's currently being used by the IDE and puts it in the sh / bat files...

magdyheiba: the property that gets us whatever the java that's currently being and puts it at the sh/bat is **"java.home"**  if we use it then the feature has no meaning in my opinion as we need to use the latest version 
let me think of something that can work with eclipse as well

MohabMohie: The idea is that we take the java home that the IDE can see, and we statically embed it in the sh/bat... That seems to work for the allure file since it also needs the java home in it... What do you think?? Ngarrab?

magdyheiba: If we run any test class through testng it will be generated correctly and it will not be a problem 
If we run through CMD and the default JDK is not the latest it will be generated with the system variable jdk and it will not run the engine and we have to set the java manually 💯 
I changed it to get the system variable and embedded it in the sh/bat
please let me know if anything else ,Thank you!

MohabMohie (approved):
Excellent!! Thank you for your valuable contribution and this excellent feature 👌

PR #509, opened 2022-02-03
dsl core components v1

MhmdElGazzar:


MhmdElGazzar:
Hello dear mohab 

please review my PR.

Thank you

MhmdElGazzar:
Dear Mohab 

please review

src/main/java/com/shaft/dsl/webElements/Button.java
```diff
@@ -0,0 +1,21 @@
...
+    public Button(WebDriver driver, By locator) {
+        super(driver, locator);
+    }
+    public void click() {
+        elementActions.click(locator);
+    }
+    public String getText() {
+        return elementActions.getText(locator);
+    }
+    public Boolean getStatus(){
```
MohabMohie, line 17: I think this method is commonly knows as isEnabled()
Also I think this returns true if the element is disabled which is actually the opposite of what it should return...
We need some test methods to cover all the new methods here

MohabMohie: I think this method is commonly knows as isEnabled()
Also I think this returns true if the element is disabled which is actually the opposite of what it should return...
We need some test methods to cover all the new methods here

src/main/java/com/shaft/dsl/webElements/Element.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.shaft.gui.element.ElementActions;
+import com.shaft.validation.Validations;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
+
+public class Element {
+    By locator;
+    WebDriver driver;
+    ElementActions elementActions;
```
MohabMohie: Can we privatize the variables?

MohabMohie (changes requested):
This is excellent work Gazzar, thank you!

As a general note we need to have every method covered by at least one test scenario.
I also think we need to rename the sub package to "gui" instead of webElements as the classes in this package will work for any gui element for web and mobile.
Lastly, do you think validations or assertions should be included here? I understand that they are commonly used and will save time... But responsibility-wise shouldn't the validations exist in the test classes? Let's discuss...
If we decide to have validations built-in here, we should make them all assertions or all verifications to have a consistent user experience.

Also, since we are working on user-facing methods, every public method must be documented with a javadocs entry, you can copy the same entry for the wrapped methods that you're calling.

Lastly, we should ensure that we cover all the existing element actions and touch actions for the elements classes...
And figure out how to do the same for the page class

Kindly take any action you think is needed and let's have a walkthrough review session afterwards. We will especially focus on the sample test classes and the end user experience.

src/main/java/com/shaft/dsl/webElements/Element.java
MhmdElGazzar: by default it is private 
because it is the pillar of OOP

src/main/java/com/shaft/dsl/webElements/Button.java
MhmdElGazzar: done

src/main/java/com/shaft/dsl/webElements/Element.java
MohabMohie: I think by default it is protected... Are you sure?

MhmdElGazzar: https://stackoverflow.com/questions/16164902/what-is-the-default-access-modifier-in-java#:~:text=The%20Default%20access%20modifier%20is,only%20from%20the%20same%20package.&text=Your%20constructor's%20access%20modifier%20would,but%20the%20constructor%20will%20not.

MhmdElGazzar:
> As a general note we need to have every method covered by at least one test scenario.
in progress 
> I also think we need to rename the sub package to "gui" instead of webElements as the classes in this package will work for any gui element for web and mobile.
half agreed but okay 
_>Lastly, do you think validations or assertions should be included here? I understand that they are commonly used and will save time... But responsibility-wise shouldn't the validations exist in the test classes? Let's discuss... If we decide to have validations built-in here, we should make them all assertions or all verifications to have a consistent user experience._
**from object oriented programming,** every method should be available only on its exact object
now every object has methods that are valid on it only 
also the only validations valid on it 
and it is surely called in the step definition

src/main/java/com/shaft/dsl/webElements/Element.java
MohabMohie: package-private is closer to protected than it is to private (class private I mean)
Do you think it's better to leave it accessible from any class in the package? Or to make it only accessible within the class?

MohabMohie:
From a page object model pattern standpoint,
Pages should have elements and actions,
Test classes or Step definitions should have assertions 

Accordingly DSL element classes should have actions (including getLocator()) which can be returned to the test classes for assertions.

It's more code but it applies the pattern... Most references (even the creator of the pom pattern and selenium webdriver) agree that validations should not be in the page classes.

Mmm... I get where you're coming from... And it will save lines of code and make things more usable and more readable... It's the same argument @MahmoudElSharkawy has...

I think I really need to see a sample page class and test class to really wrap my head around this...
Also please share any references on people who have applied DSL with POM before... This is a first for me so I'll be asking a lot of questions.

MhmdElGazzar:
>Every public method must be documented with a javadocs entry.
this is not the design perspective I am doing, I am trying to make it so intuitive and easy to understand 
trying to achieve this by being so picky in names and relating components to each other
also using the domain names.
please check all the content if there is something that is not intuitive or might need a documentation 
please let me know, so that I re-work on it until it become easily and correctly digested by any user
> 
> Lastly, we should ensure that we cover all the existing element actions and touch actions for the elements classes... And figure out how to do the same for the page class
in this point, I would like to go with the agile approach, small incremental sprints and additions till we reach the full release.
first I am using the most used elements in our portals 
then, and through code review on me and other mates, any potential component will be extracted and utilized in the engine 
I have plans for pages, but would like to get more familiar with elements design before taking further design decision  
> 
> Kindly take any action you think is needed and let's have a walkthrough review session afterwards. We will especially focus on the sample test classes and the end user experience.
okay sure, any time

src/main/java/com/shaft/dsl/webElements/Element.java
MhmdElGazzar: I need it so 
so that I can work on them in the sub classes 
i have even doughts that I will need to convert them to protected if I needed them in the subclasses

MohabMohie:
Regarding the agile approach -> agreed, usually this makes me move the engine into BETA mode because method names will be subject to change.

Regarding javadocs -> I tried to take the same approach for the validations wizard but several users demanded that it be properly documented... There's actually an issue open for this specific topic right now... Again I get your point because I've done the same, but the customer is always right I guess... And having docs wouldn't really hurt...

Regarding the walkthrough -> let's schedule it after the sample test and page classes.have been pushed... Maybe just seeing them will quench my thirst 🙏🏻

src/main/java/com/shaft/dsl/webElements/Element.java
MohabMohie: Sounds good to me 👌

MhmdElGazzar:
I think I really need to see a sample page class and test class to really wrap my head around this...
> didn't I sent you over teams ?
Also please share any references on people who have applied DSL with POM before... This is a first for me so I'll be asking a lot of questions.
sure 
for a quick idea, this link is good https://airbrake.io/blog/software-design/domain-driven-design
for a good reference, Domain Driven Design for Martin Fowler (he is the boss in sw design and refactoring BTW)
for implementation in testing 
https://www.slideshare.net/alimenkou/dsl-page-object-and-web-driver-the-path-to-reliable-functional-tests

MhmdElGazzar:
Regarding javadocs -> in this point we have the Boss support on his official website 
https://martinfowler.com/bliki/CodeAsDocumentation.html
but to state things clear 
for code to serve as a documentation, it has to be clean code 
applying solid principles and doesn't have code smells 
------------------------------------------------------------------------------
returning to the point 
fowler says -> a code that doesn't reveal intention easily and can't serve as a documentation is most likely to be a bad code 
and investing in the code it self to be more clean is the key 
-----------------------------------------------------------------------------
also to be honest, as a user of shaft
 some methods in shaft cuz confusion and push to miss use
and those require me to have a doc 
that is why I am emphasizing on oop, clean code and solid 
the effort in shaft is extremely excellent, but innovation doesn't stop 
isa there are many and many points to enhance and optimize

MohabMohie:
> 

I mean a sample of the final product in the engine itself.
Great references 👌 I'm convinced...

src/main/java/com/shaft/dsl/webElements/Element.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public Element(WebDriver driver, By locator) {
+        this.driver = driver;
+        this.locator = locator;
+        elementActions = new ElementActions(this.driver);
+    }
+
+    public void shouldExist()
+    { Validations.verifyThat().element(driver,locator).exists().perform();
```
MohabMohie: Can we convert those to assertions? Or at least have two variants one clearly named
assertThatElementExists
And another one
veriftThatElementExists
I think that would be more descriptive and flexible to use...

MohabMohie (changes requested):
This looks perfect! Thank you :)

Let's kindly resolve these code quality comments
https://app.codacy.com/gh/MohabMohie/SHAFT_ENGINE/pullRequest?prid=8826587

And I have only one point to discuss regarding assertions and verifications.

src/main/java/com/shaft/dsl/webElements/Element.java
MhmdElGazzar: Hello Dear 
according to Robert C martin in his book clean code chapter 2 
under Meaningful Names -> Pick one word per concept
using this concept verifyThatElement will break it 
but the point u cleared should be revisited

MhmdElGazzar:
Hello Dear Mohab 
points discussed has been fixed , tests has been added

MhmdElGazzar:
please let me know if u need further actions

src/main/resources/defaultProperties/Reporting.properties
```diff
@@ -8,8 +8,8 @@ alwaysLogDiscreetly=false
 debugMode=false
 cleanAllureResultsDirectoryBeforeExecution=true
 generateAllureReportArchive=false
-openAllureReportAfterExecution=true
-generateExtentReports=true
+openAllureReportAfterExecution=false
+generateExtentReports=false
 cleanExtentReportsDirectoryBeforeExecution=true
 ###################################################
```
MohabMohie: Kindly revert the changes for this file

src/main/java/com/shaft/dsl/webElements/RadioButton.java
```diff
@@ -0,0 +1,17 @@
+package com.shaft.dsl.webElements;
```
MohabMohie: Kindly let's change the package name to "gui" instead of "webElement" because these WebDriver elements are actually usable for web/native mobile/and native desktop elements.
Also gui maps with the current package structure and we can later add other sub packages named api, cli, and db to cover the rest of the dsl objects

MhmdElGazzar: done dear

src/main/resources/defaultProperties/Reporting.properties
MhmdElGazzar: done

MohabMohie (approved):

src/main/java/com/shaft/dsl/webElements/Element.java
MohabMohie: Makes sense

PR #518, opened 2022-02-10
Update Button.java

MhmdElGazzar:


src/main/java/com/shaft/dsl/gui/Button.java
```diff
@@ -11,11 +12,14 @@ public Button(WebDriver driver, By locator) {
...
         elementActions.click(locator);
     }
-    public String getText() {
-        return elementActions.getText(locator);
-    }
     public Boolean isEnabled(){
         return Boolean.parseBoolean(elementActions.getAttribute(locator,"disabled"));
     }
+    public void shouldBeEnabled()
+    {  Validations.assertThat().object(isEnabled()).isTrue().perform();
```
MohabMohie: we need to use assertThat().element().attribute() in this one instead of assert true/false.
the element attribute assertion can capture screenshot evidence and highlight the element in green or yellow if it passed or failed respectively.

MohabMohie (changes requested):

MhmdElGazzar:
Hello dear mohab 
comment has been done

MohabMohie (approved):

PR #530, opened 2022-02-22
Dsl core components v2

MhmdElGazzar:


src/main/java/com/shaft/dsl/gui/SearchField.java
```diff
@@ -0,0 +1,17 @@
+package com.shaft.dsl.gui;
+
+import org.openqa.selenium.By;
+
+public class SearchField extends TextBox {
+   public Button clearBtn;
+
```
MohabMohie: I don't think the search field is a standard DSL expression, can you tell me more about it or remove it?

MohabMohie (changes requested):
Excellent work!

Let's put on the to-do list to make the driver in the element class a threadlocal object, and the set driver / get driver methods "synchronized" to support parallel creation of objects.

Let's close the last comment.

src/main/java/com/shaft/dsl/gui/SearchField.java
MhmdElGazzar: Search field is a textbox with close button that have an id
so it is a domain component not a text field 
also it is a part of another components coming in progress

MohabMohie: I think it's project specific though... didn't see it before

MhmdElGazzar:
@MohabMohie 
please let us meet and know more about synchronizing 
it is a new topic for me , so I would like to understand its mindset

src/main/java/com/shaft/dsl/gui/SearchField.java
MhmdElGazzar: it is very common in the portals with modern 
image

MhmdElGazzar: image
it is W3C element

MhmdElGazzar: I can remove it , but I will need it extensively in coming components; filters , searching and so

MohabMohie: to me, this still looks like a normal text input field (albeit a single line one) the only difference is that you can press enter to search, or click a button to search, or perform .submit...

seems to me that having a regular input field in SHAFT with the ability to type, submit, or perform a key press is sufficient for this need.

can we delay the discussion and remove this component for now so that I can generate the new build? I think you have a valuable point... I'm not getting it yet... and I want to push the new release for you... so let's exclude this file for now but keep the conversation going.

sounds good?

MhmdElGazzar: fair enough

MhmdElGazzar: done

MohabMohie (approved):

PR #532, opened 2022-02-22
Update DatabaseActions.java

TarekGaafar:
Data Base Actions-separating update, insert and delete into separated methods.


MohabMohie (changes requested):
Can we kindly have one private method with the code, and the other methods can call it.

TarekGaafar:
@MohabMohie 
The bolded comment below is done.
Kindly I am ready to do any comments else.
**Can we kindly have one private method with the code, and the other methods can call it.**

thank for supporting

MohabMohie (approved):
Thank you for your valuable contribution! and welcome to our list of contributors!! 💯

PR #551, opened 2022-03-15
Report issues after testcase failure_ TestNG Testing Methods

RihamMostafa:
Code covers:
- Reporting bugs for TestNG test methods failure with the following data:
> Attachments [per test Method].
> Test Method log in the bug description.
> The bug will be named after the test-case method name. 
> Linked to the test case only if TMSLink is provided. 
> assignee if provided in the configuration, the issue will be not assigned if not.
- configuration manager updated with the required enablers.

Notes:
- Reporting issues can be only enabled by enabling the integration and report bugs options from the configuration manager.
- Reporting bugs is not linked to the execution reporting. which means that the user still has the flexibility to report bugs without reporting the execution.
- It is pure JIRA integration doesn't depend on any TMS plugin.

Next: Cucumber scenarios bugs reporting.


src/main/resources/JIRA/createIssueRequestBody.json
```diff
@@ -0,0 +1,15 @@
+{
```
MohabMohie: These files won't exist when users use SHAFT as a dependency, so SHAFT won't be able to read them.

I recommend creating a class that holds both requests as simple strongs with getters so you can get and later replace the values you need.

With the latest JAVA you can actually use three double quotes, and this will create something called multiline string, try it... You'll like it... You can paste the json content inside it and jt will work.

Lastly it's a good practice to type ${VARIABLE_NAME} all in upper snake case and between curly brackets with the dollar sign as the placeholders for the text that you intend to change. It's better for readability so that people who read the file or string can immediately know that you intend to replace these substrings.

src/main/resources/JIRA/linkJIRATicketRequestBody.json
```diff
@@ -0,0 +1,16 @@
+{
```
MohabMohie: Same comment as the other .json file

src/main/java/com/shaft/gui/image/ScreenshotManager.java
```diff
@@ -778,4 +778,7 @@ private static synchronized void appendToAnimatedGif(byte[] screenshot) {
             ReportManagerHelper.log(e);
         }
     }
+    public static String getGifRelativePathWithFileName() {
```
MohabMohie: Can we make the gif method return its oath instead?

MohabMohie (changes requested):
Excellent work as always! Thank you :)
And kindly let's dicuss the couple of minior comments I shared.

hernandezmshawntae:
Please and Thank You.

src/main/resources/JIRA/createIssueRequestBody.json
RihamMostafa: Yeah, I really loved the three double quotes tip :D 
Thank you Mohab for your valuable comments. all changes applied.

src/main/java/com/shaft/gui/image/ScreenshotManager.java
RihamMostafa: Done. I didn't before because _attachAnimatedGif()_ was clearing the path at the end.

hernandezmshawntae:
Privacy Issues Data Loss

hernandezmshawntae:
Please AndThank You

hernandezmshawntae:
please and thank  you

MohabMohie (approved):
Thank You!

PR #586, opened 2022-04-08
openInNewTab

magdyheiba:


src/main/java/com/shaft/gui/element/ElementActions.java
```diff
@@ -533,7 +533,15 @@ public static void switchToIframe(WebDriver driver, By elementLocator) {
...
         WebDriverElementActions.switchToWindow(driver, nameOrHandle);
     }
-
+    /**
+     * Switches focus to another Tap
+     *
+     * @param driver       the current instance of Selenium webdriver
+     * @param URL The name of the URL you want to navigate to
+     */
+    public static void switchToNewTap(WebDriver driver, String URL) {
```
MohabMohie: opening a link in a new tab is not an element action.
kindly move it to browser actions instead.
also please note that every method has two variants, one is static and one is a regular method. the static method takes a driver object as a parameter and the regular method uses the existing driver object.

src/main/java/com/shaft/gui/element/WebDriverElementActions.java
```diff
@@ -1146,6 +1147,26 @@ public static void switchToWindow(WebDriver driver, String nameOrHandle) {
             WebDriverElementActions.failAction(driver, nameOrHandle, null);
         }
     }
+    /**
+     * Switches focus to another Tap
+     *
+     * @param driver       the current instance of Selenium webdriver
+     * @param URL The name of the URL you want to navigate to
+     */
+    public static void switchToNewTap(WebDriver driver, String URL) {
```
MohabMohie: the method should validate that the link has been opened in a new tab successfully.
this required getting window handles before and after opening the new tab, and confirming that the new tab was opened, printing the new tab window handle in the console using logdiscrete

MohabMohie (changes requested):
thank you for your valuable contribution. kindly check the requested comments.

src/main/java/com/shaft/gui/element/WebDriverElementActions.java
magdyheiba: Done(Y)

src/main/java/com/shaft/gui/element/ElementActions.java
magdyheiba: Done (Y)

MohabMohie (approved):
Thanks for your valued contribution!

PR #606, opened 2022-05-13
Tink Integration

MhmdElGazzar:
Hello Mohab 

please find me code for tink integration 

Thank you,


MohabMohie:
Dear Mohammed,
Thank you for this attempt, but I will close this PR since it is not aligned to the agreed requirements based on our previous discussions. This is what we agreed to:
- All test data under the test data directory (already a property) should be encrypted manually by the user using the CMD tool. We should provide the user with a bat file that they can double-click and it will encrypt all their data for them recursively. This is done outside the engine.
- All test data under the test data directory (already a property) should be decrypted automatically and recursively by SHAFT before any test execution.
- To automatically decrypt the data, SHAFT will accept some properties from the user to show the name of the key and any other parameters required by tinkey.
- This should be implemented to support both local keys and remote keys as per tinkeys documentation.
- All decrypted test data should be deleted automatically by SHAFT after any execution to ensure that only the encrypted files exist in SHAFT's test data directory.
- This concludes the initial business requirements, with area for future enhancements.

MohabMohie:
FYI as agreed: https://github.com/MohabMohie/SHAFT_ENGINE/pull/608

I took the time to meticulously document my thought process, and every single step taken to reach this implementation for your reference: https://github.com/MohabMohie/SHAFT_ENGINE/blob/b95e86300d52f49d9810bf5d0d2d184fc0aa5753/src/main/java/com/shaft/tools/security/tinkey.md

This PR (if the tests pass) should satisfy all the needed security requirements, with rough support for AWS and GCP KMS as well as local key files. However I didn't get a chance to test AWS and GCP integration. However, with the current implementation we should get enhanced security and the impact on performance -to go fetch the key only once from the server- will be negligible.

Issue #641, opened 2022-06-22
When used SwipeElementIntoView(Element, TouchActions.SwipeDirection.DOWN), get "element locator is NULL" Error


Mohab-Osama:
When used SwipeElementIntoView(Element, TouchActions.SwipeDirection.DOWN), get "element locator is NULL" Error

**To Reproduce**
1. Add SwipeElementIntoView(Element, TouchActions.SwipeDirection.DOWN) 
ex: ElementActions.performTouchAction(driver).swipeElementIntoView(M2, TouchActions.SwipeDirection.DOWN);

2. Run test

**Expected behavior**
Element swipe into view successfully

**Screenshots**
![image](https://user-images.githubusercontent.com/108004467/175036687-13ced2de-e218-470f-8f0f-6c541587eacb.png)

Issue occurred while testing native android app and could simulate it in chrome desktop browser SHAFT_ENGINE version 6.2.20220609.

**Desktop :
 - OS: Windows
 - Browser Chrome
 - Version 102

**Smartphone :
 - Device: [e.g. Galacy J5]
 - OS: [e.g. Android 9]

MohabMohie:
done and will be pushed with the next release.

PR #643, opened 2022-06-22
1- Fix issue #641 , 2- Add Test_SwipeElementIntoView class to test the swipeElementIntoView function

Mohab-Osama:
1- Fix issue #641 Element locator is NULL by make internalScrollableElementLocator can be null and the condtion become 'OR' not 'AND'
2- Add Test_SwipeElementIntoView class  to test the swipeElementIntoView function


MohabMohie (approved):
Thank you for your valuable contribution

PR #668, opened 2022-08-08
Add advanced JSON path evaluator with tests

MahmoudElSharkawy:
Add advanced JSON path evaluator **jayway.jsonpath** library to be able to use filters and more advanced expressions in the JSON path.


MohabMohie (approved):
Welcome back!

PR #689, opened 2022-08-21
Adding support to read YAML files as test data

Kinasr:
Adding support to read YAML files as test data

### DIFF
--- a/src/main/java/com/shaft/tools/io/YAMLFileManager.java
+++ b/src/main/java/com/shaft/tools/io/YAMLFileManager.java
@@ -1,0 +1,404 @@
+package com.shaft.tools.io;
+
+import com.shaft.tools.support.JavaActions;
+import org.yaml.snakeyaml.Yaml;
+
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.*;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+public class YAMLFileManager {
+    public final static String KEY_CONTAINS_LIST_REGEX = "^[-a-zA-Z\\d_!@#$%^&*()+=|\\\\/?><\"'{}~]*(\\[\\d+])+$";
+    public final static String NUMBER_IN_SQUARE_BRACKETS_REGEX = "\\[\\d+]";
+    public final static String SQUARE_BRACKETS_REGEX = "[\\[\\]]";
+    public final static String KEY_SEPARATOR_REGEX = "\\.";
+
+    private final Map<String, Object> data;
+    private final String filePath;
+
+    // Splitting the key to a list of keys using the dot as the splitter.
+    private final Function<String, List<String>> splitKey =
+            key -> Arrays.stream(key.split(KEY_SEPARATOR_REGEX)).toList();
+    // Checking if the key has an index for a list [index]
+    private final Predicate<String> keyHasList =
+            key -> key.matches(KEY_CONTAINS_LIST_REGEX);
+
+    /**
+     * Creating an instance of {@link YAMLFileManager}
+     *
+     * @param filePath target test data yaml file path
+     */
+    public YAMLFileManager(String filePath) {
+        this.filePath = JavaActions.appendTestDataToRelativePath(filePath);
+        this.data = getData();
+    }
+
+    /**
+     * Fetch all the date from the YAML file
+     *
+     * @return all date existed in the YAML file as {@link Map}
+     */
+    public Map<String, Object> get() {
+        return data;
+    }
+
+    /**
+     * Fetch a single piece of data from the YAML file using a single key or a series of keys
+     * <p>
+     * to support key series use a dot to separate keys "key1.key2"
+     * to support list in the key series "key[index]"
+     *
+     * @param key the path to the wanted data can be a single key or a series of keys
+     * @return the wanted value as {@link Object}
+     */
+    public Object get(String key) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/shaft/tools/io/YAMLFileManager.java
```diff
@@ -0,0 +1,404 @@
...
+    // Checking if the key has an index for a list [index]
+    private final Predicate<String> keyHasList =
+            key -> key.matches(KEY_CONTAINS_LIST_REGEX);
+
+    /**
+     * Creating an instance of {@link YAMLFileManager}
+     *
+     * @param filePath target test data yaml file path
+     */
+    public YAMLFileManager(String filePath) {
```
MohabMohie, line 36: We should add a log message and attach the test data file to the report here in alignment with this method
https://github.com/ShaftHQ/SHAFT_ENGINE/blob/b90e8ff225776bc014f6f57680fc3bf67f646e24/src/main/java/com/shaft/tools/io/JSONFileManager.java#L27-L41

src/main/java/com/shaft/tools/io/YAMLFileManager.java
```diff
@@ -0,0 +1,404 @@
...
+     */
+    private FileInputStream getFile() {
+        FileInputStream in;
+
+        try {
+            in = new FileInputStream(filePath);
+        } catch (FileNotFoundException rootCauseException) {
+            var errorMsg = "Couldn't find the desired file. [" + filePath + "].";
+            sendFailureCompo(errorMsg, rootCauseException);
+            throw new RuntimeException(errorMsg);
```
MohabMohie: Note that in case of exceptions that should fail the execution we should follow this pattern for consistent reporting:
https://github.com/ShaftHQ/SHAFT_ENGINE/blob/b90e8ff225776bc014f6f57680fc3bf67f646e24/src/main/java/com/shaft/tools/io/JSONFileManager.java#L119-L122

src/main/java/com/shaft/tools/io/YAMLFileManager.java
```diff
@@ -0,0 +1,404 @@
...
+                    (k, v) -> nMap.put(k.toString(), parseObjectTo(v, clazz))
+            );
+            return nMap;
+        }
+        var errorMsg = "Can't parse the value of [" + obj + "] to be map";
+        sendFailureCompo(errorMsg);
+        throw new RuntimeException(errorMsg);
+    }
+
+    private void sendFailureCompo(String msg, Exception... rootCauseException) {
```
MohabMohie, line 430: We can use this common method to apply the failure pattern mentioned in the previous review comment, and ensure that all caught exceptions and failed methods call it.

src/test/java/testPackage01/Test_YAMLFileManager.java
```diff
@@ -0,0 +1,254 @@
...
+    static SHAFT.TestData.YAML yaml;
+
+    @BeforeMethod
+    static void setUp() {
+        yaml = new SHAFT.TestData.YAML("src/test/resources/testDataFiles/yaml/yaml_test_data.yaml");
+    }
+
+    @Test
+    static void getObject() {
+        assertThat()
```
MohabMohie, line 23: We should use the built-in assertions to ensure that the end user experience will work as expected, and in case any of the returned datatypes are not handled by the assertion methods of the engine, we should enhance them to ensure consistent behavior.

https://github.com/ShaftHQ/SHAFT_ENGINE/blob/b90e8ff225776bc014f6f57680fc3bf67f646e24/src/test/java/testPackage01/Test_JSONFileManager.java#L10-L14

We should also implement a public method with the same name "getTestData" that has the ability to parse any datatype and return it as a string without the user needing to determine the desired datatype and calling the relevant method.

For that I think the method can be a synonym for getString() in case this one can also handle and parse different data types that are not natively string.

MohabMohie (changes requested):
Okay wow!
Thanks a lot for a very comprehensive PR, it's really very good! I would say you hit 90% completion for this new feature on your first try and without any prior alignment which is EXCELLENT 👌
I have only a few comments to discuss, and they shouldn't be time consuming at all.
Kindly examine the comments and let's discuss as needed here or via slack.

src/main/java/com/shaft/tools/io/YAMLFileManager.java
Kinasr: Thanks 
I will work on these notes.

src/test/java/testPackage01/Test_YAMLFileManager.java
Kinasr: Here I'm using the SHAFT assertion in all tests except tests that throw exceptions.

src/main/java/com/shaft/tools/io/YAMLFileManager.java
Kinasr: I think I solved them.
Thanks

src/test/java/testPackage01/Test_YAMLFileManager.java
MohabMohie: Thank you, this is excellent!
Maybe we should add a new validation method for "throws" in a future PR to cover this case

MohabMohie (approved):
Thanks so much for your amazing effort.
And welcome onboard as one of SHAFT_Engine opensource contributions. 🎉
Feel free to celebrate your new feature on social media and use #shaft_engine

MohabMohie:
PR approved and will be merged automatically after pipelines pass.

PR #775, opened 2022-11-14
Cover bearer token authorization on the XRay integration

RihamMostafa:
- Update the Xray helper class to handle _bearer token_ auth instead of supporting _Basic_ only.  
- Enhance the Xray tab on the configuration manager to cover the changes.


MohabMohie (approved):
Thanks for your valuable contribution!

PR #795, opened 2022-12-13
Feature#794

magdyheiba:
Screenshot 2022-12-14 at 1 12 06 AM
Screenshot 2022-12-14 at 1 11 55 AM
Screenshot 2022-12-14 at 1 11 13 AM

### DIFF
--- a/src/main/java/com/shaft/gui/browser/WebDriverBrowserActions.java
+++ b/src/main/java/com/shaft/gui/browser/WebDriverBrowserActions.java
@@ -231,13 +231,19 @@ public class WebDriverBrowserActions {
      */
     public static void navigateToURL(WebDriver driver, String targetUrl, String targetUrlAfterRedirection) {
         // check if the user wrote the URL ended with "/"
-        String finalUrl = null;
+        String modifiedUrl = null;
         if (System.getProperty("baseURL").endsWith("/")) {
-            finalUrl = System.getProperty("baseURL").substring(0, System.getProperty("baseURL").length() - 1);
+            modifiedUrl = System.getProperty("baseURL").substring(0, System.getProperty("baseURL").length() - 1);
+        }else{
+            modifiedUrl=System.getProperty("baseURL");
         }
         // check if the user sends the URL with the abbreviation of "./"
         if (targetUrl.startsWith(".")) {
-            targetUrl = targetUrl.replaceFirst(".", finalUrl);
+            targetUrl = targetUrl.replaceFirst(".", modifiedUrl);
+        }
+        if(!targetUrl.startsWith("http")){
+            ReportManager.logDiscrete("Your Base Url property is empty");
+            failAction(driver,targetUrl);
         }
         if (targetUrl.equals(targetUrlAfterRedirection)) {
             ReportManager.logDiscrete(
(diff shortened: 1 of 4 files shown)


src/main/resources/defaultProperties/WebCapabilities.properties
```diff
@@ -10,6 +10,7 @@ mobileEmulation.width=
 mobileEmulation.height=
 mobileEmulation.pixelRatio=
 mobileEmulation.userAgent=
+BaseURL=https://demoqa.com/
```
MohabMohie: can we convert to lower camel case please?
we also need to expose this property in the configuration manager :)

MohabMohie (changes requested):
Thanks a lot for the great added value!
Let's consider using lower camel case and add to the configuration manager.
Do you want my support with that?

src/main/java/com/shaft/gui/browser/WebDriverBrowserActions.java
```diff
@@ -230,6 +230,15 @@ private static String getDomainNameFromURL(String url) {
...
      */
     public static void navigateToURL(WebDriver driver, String targetUrl, String targetUrlAfterRedirection) {
+        // check if the user wrote the URL ended with "/"
+        String finalUrl = null;
+        if (System.getProperty("baseURL").endsWith("/")) {
+            finalUrl = System.getProperty("baseURL").substring(0, System.getProperty("baseURL").length() - 1);
+        }
+        // check if the user sends the URL with the abbreviation of "./"
+        if (targetUrl.startsWith(".")) {
+            targetUrl = targetUrl.replaceFirst(".", finalUrl);
```
MohabMohie: should this be "finalUrl =" ??

src/main/java/com/shaft/gui/browser/WebDriverBrowserActions.java
```diff
@@ -830,41 +839,41 @@ public WebDriverBrowserActions fullScreenWindow() {
...
     }
-    
+
     /**
      * Switches focus to another Tab
      *
      * @param driver       the current instance of Selenium webdriver
      * @param URL The name of the URL you want to navigate to
      */
     public static void switchToNewTab(WebDriver driver, String URL) {
```
MohabMohie: formatting issue?

src/main/resources/defaultProperties/WebCapabilities.properties
```diff
@@ -10,6 +10,7 @@ mobileEmulation.width=
 mobileEmulation.height=
 mobileEmulation.pixelRatio=
 mobileEmulation.userAgent=
+baseURL=https://demoqa.com
```
MohabMohie: can we set the default to empty please? we also need to add some validations to ensure that "empty" doesn't throw any exceptions.

src/main/java/com/shaft/gui/browser/WebDriverBrowserActions.java
magdyheiba: I changed it to Modified URL and added cases to handle If user selects to navigate with the new feature but didn't add a base Url

src/main/resources/defaultProperties/WebCapabilities.properties
magdyheiba: Screenshot 2022-12-14 at 3 36 11 PM

When a user selects to navigate with the new feature but didn't add a base Url , SHAFT will throw an error as described at the screenshot

src/main/java/com/shaft/gui/browser/WebDriverBrowserActions.java
magdyheiba: Fixed el7amdoulelah through eclipse :)

src/main/resources/defaultProperties/WebCapabilities.properties
magdyheiba: Done

MohabMohie (approved):
Thanks a lot for your valuable contribution 🙏

PR #837, opened 2023-01-23
handle the null pointer exception.

magdyheiba:
Screenshot 2023-01-23 at 10 09 48 PM
Screenshot 2023-01-23 at 10 09 01 PM
handle the null pointer exception that happen while getting getTargetBrowserName() at Remote execution

### DIFF
--- a/src/main/java/io/github/shafthq/shaft/gui/element/ElementActionsHelper.java
+++ b/src/main/java/io/github/shafthq/shaft/gui/element/ElementActionsHelper.java
@@ -102,13 +102,8 @@ public class ElementActionsHelper {
     }
 
     private static boolean isSafariBrowser() {
-        boolean isSafariBrowser= false;
-        try {
-            isSafariBrowser= DriverFactoryHelper.getTargetBrowserName().toLowerCase().contains("safari");
-        }
-        catch (NullPointerException exception){
-        }
-        return isSafariBrowser ;
+        DriverFactoryHelper.setTargetBrowserName(System.getProperty("targetBrowserName"));
+        return DriverFactoryHelper.getTargetBrowserName().toLowerCase().contains("safari");
     }
 
     public static ArrayList<Class<? extends Exception>> getExpectedExceptions(boolean isValidToCheckForVisibility) {
--- a/src/main/java/io/github/shafthq/shaft/gui/element/ElementActionsHelper.java
+++ b/src/main/java/io/github/shafthq/shaft/gui/element/ElementActionsHelper.java
@@ -102,7 +102,13 @@ public class ElementActionsHelper {
     }
 
     private static boolean isSafariBrowser() {
-        return DriverFactoryHelper.getTargetBrowserName().toLowerCase().contains("safari");
+        boolean isSafariBrowser= false;
+        try {
+            isSafariBrowser= DriverFactoryHelper.getTargetBrowserName().toLowerCase().contains("safari");
+        }
+        catch (NullPointerException exception){
+        }
+        return isSafariBrowser ;
     }
 
     public static ArrayList<Class<? extends Exception>> getExpectedExceptions(boolean isValidToCheckForVisibility) {
--- a/src/main/java/io/github/shafthq/shaft/driver/DriverFactoryHelper.java
+++ b/src/main/java/io/github/shafthq/shaft/driver/DriverFactoryHelper.java
@@ -19,6 +19,7 @@ import io.github.shafthq.shaft.tools.support.JavaHelper;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.NonNull;
+import lombok.Setter;
 import org.openqa.selenium.*;
 import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.chromium.ChromiumOptions;
@@ -57,6 +58,7 @@ public class DriverFactoryHelper {
     @Getter(AccessLevel.PUBLIC)
     private static String targetOperatingSystem;
     @Getter(AccessLevel.PUBLIC)
+    @Setter(AccessLevel.PUBLIC)
     private static String targetBrowserName;
     @Getter(AccessLevel.PUBLIC)
     private static final ThreadLocal<WebDriver> driver = new ThreadLocal<>();
--- a/src/test/java/testPackage01/appium/Test_BasicActionsOnBrowserStack.java
+++ b/src/test/java/testPackage01/appium/Test_BasicActionsOnBrowserStack.java
@@ -1,0 +1,54 @@
+/*
+package testPackage01.appium;
+
+import com.shaft.driver.SHAFT;
+import io.appium.java_client.AppiumBy;
+import org.openqa.selenium.By;
+import org.testng.Assert;
+import org.testng.annotations.*;
+
+import java.time.Duration;
+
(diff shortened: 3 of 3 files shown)


src/main/java/io/github/shafthq/shaft/gui/element/ElementActionsHelper.java
```diff
@@ -102,7 +102,13 @@ private static boolean isValidToCheckForVisibility(By elementLocator, boolean ch
     }
 
     private static boolean isSafariBrowser() {
-        return DriverFactoryHelper.getTargetBrowserName().toLowerCase().contains("safari");
+        boolean isSafariBrowser= false;
```
MohabMohie: Why is it throwing a null? Is the user deleting the target browser property manually or what?

MohabMohie: Let's try to prevent the error from the source rather than try/catching it. I think it may be related to the new properties management system, this issue is a regression as it wasn't there before, so let's investigate the root cause some more to ensure that no other sleeper issues arise.

magdyheiba: Fixed and pushed 
the problem was when calling getExpectedExceptions() and calling isSafariBrowser() when getting getTargetBrowserName() it was null as it wasn't initialized here

MohabMohie: I just added a browserstack thread to the teating pipeline. Is this only happening with browserstack mobile?

magdyheiba:
Screenshot 2023-01-24 at 8 38 53 PM
Yes just tried it with Emulator and working normally and here is the result of the run

MohabMohie:
@magdyheiba this is shaping up to be a compounded issue magdy...
the new properties manager reads properties into variables that you can easily get/set via code without needing to use system.setproperty or system.getproperty

with this setup code block:
```java
        // common attributes
        System.setProperty("targetOperatingSystem", "Android");
        System.setProperty("mobile_automationName", "UIAutomator2");
        System.setProperty("mobile_appWaitActivity","*");
        System.setProperty("mobile_disableWindowAnimation","true");

        // remote browserstack server (existing app version)
        System.setProperty("executionAddress", "browserstack");
        System.setProperty("browserStack.platformVersion", "11.0");
        System.setProperty("browserStack.deviceName", "Google Pixel 4");
        System.setProperty("browserStack.appName", "ApiDemos-debug.apk");
        System.setProperty("browserStack.appRelativeFilePath", "");
        System.setProperty("browserStack.appUrl", "bs://030ae95f0aa6d82ca804e342adde364c2614b419");
        driver = DriverFactory.getDriver();
```
I'm expecting browserstack to open and the test to run, instead i'm getting these logs:
```terminal
2023-01-24 22:12:34.415  INFO [@508 - main] com.shaft.tools.io.ReportManager : Attempting to run locally on: "Windows-64 | Google chrome".
2023-01-24 22:12:34.418  INFO [@508 - main] com.shaft.tools.io.ReportManager : Identifying OS/Driver combination and selecting the correct driver version automatically. Please note that if a new driver executable will be downloaded it may take some time...
2023-01-24 22:12:38.988  INFO [@1238 - main] io.github.bonigarcia.wdm.WebDriverManager : Using chromedriver 109.0.5414.74 (resolved driver for Chrome 109)
2023-01-24 22:12:39.088  INFO [@1291 - main] io.github.bonigarcia.wdm.WebDriverManager : Exporting webdriver.chrome.driver as C:\Users\Mohab\.cache\selenium\chromedriver\win32\109.0.5414.74\chromedriver.exe
Starting ChromeDriver 109.0.5414.74 (e7c5703604daa9cc128ccf5a5d3e993513758913-refs/branch-heads/5414@{#1172}) on port 9146
Only local connections are allowed.
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.
[1674591165.697][WARNING]: Ignoring unrecognized log type: client
[1674591165.697][WARNING]: Ignoring unrecognized log type: profiler
[1674591165.697][WARNING]: Ignoring unrecognized log type: server
Jan 24, 2023 10:12:49 PM org.openqa.selenium.devtools.CdpVersionFinder findNearestMatch
WARNING: Unable to find an exact match for CDP version 109, so returning the closest version found: 108
2023-01-24 22:12:49.382  INFO [@525 - main] com.shaft.tools.io.ReportManager : Successfully Opened: "Windows-64 | Google chrome".
```

This means that the properties set by system.setproperties AFTER the properties loader has been triggered are not being respected.

As you can see from this screenshot, tracing the code at the moment where it tried to get the property and see if it's a browserstack execution or not, what's in-memory doesn't match what was hardcoded and therefore expected.

image

MohabMohie:
locking in on this specific line
image
which calls this method
image
we can finally locate the real culprit!
image

MohabMohie:
going back to the logic... these properties shouldn't exist in testng, this should only be used in case of crossbrowser mode
and here we go... past Mohab with his silent unasked for enhancements introducing sleeper bugs :D
image

This here overrides those variables with what was previously read by the property manager... which resets them to whatever was in the user files and ignores whatever was in the system properties...

I'm starting to think that this is another issue that we just discovered by luck :D but i'll fix, test, and keep track of everything here for later reference.

MohabMohie:
removing the useless else condition, we're now going into the correct logical path... let's proceed
image

MohabMohie:
upgraded the system.setproperty lines to use the new syntax.
setting the properties this way ensures that both the system properties and the SHAFT properties object get updated together, so that the value is correct however you choose to read it at later stages of the process...
image
and we go in for another debugging session to validate...
this time it seems we're going through the right pass and we have the bstack:options and the correct properties...
image
no more null reference exceptions here...
image
everything is looking nice here... preparing to merge bstack:options with the rest of the mobile capabilities...
image
capabilities merged successfully... i'm expecting smooth sailing now :D we're out of the rough waters...
image

magdyheiba:
Mic Dropped :D , great Job! 🥇

MohabMohie:
moment of truth!!!
image
we hit an exception... this is only supposed to happen when dealing with an old appium server... even though the capabilities clearly state using appium 2... alarming... proceeding with caution...
image
second attempt after appending the legacy appium server /wd/hub
image
oops... we hit an exception... but now we will take benefit from another useless feature past mohab implemented... extended stacktrace :D 
image
image

well, this looks like a valid exception to me :D it seems that my testing apk has expired and thus been deleted from the browserstack server... i'll retest with the "new app" configuration instead of the "existing app"

MohabMohie:
new app configuration in place... running (without debugging this time) :D #yolo 
image
hahahahahahahahahaha... as expected... we just solved an entirely different issue... #839 I knew it... that's why I opened it :D 
image
and the journey continues....

MohabMohie:
and now for the real mic drop... I'll just initialize that variable with an empty string... aaaaand
image
image
:D done... I'll open a new PR ;)

MohabMohie:
PR opened https://github.com/ShaftHQ/SHAFT_ENGINE/pull/840
closing this one.

Thanks a lot Magdy, if it weren't for your help I wouldn't have paid attention to the MAJOR underlying issue hiding behind the null. I really appreciate your valued contributions.

Issue #804, opened 2022-12-25
[Feature] API Response assertions - extractedJsonValueAsList


MohabMohie:
make this work:
`driver.assertThatResponse().extractedJsonValueAsList("$[?(@.completed==true)].completed").isEqualTo("true").perform();`

PR #849, opened 2023-01-30
API validation extractedJsonValueAsList

MahmoudElSharkawy:
Solving #804 to be able to assert on a response json list directly.


MohabMohie:
I actually wanted to approach this differently...
Instead of having it's own keyword, I wanted to have the assertion look like this

response().body().jsonPath().asList()

MahmoudElSharkawy:
> I actually wanted to approach this differently...
> 
> Instead of having it's own keyword, I wanted to have the assertion look like this
> 
> 
> 
> response().body().jsonPath().asList()
> 
> 
> 
> Is that even doable though?? 😂


I think this is another enhancement scope that will require to change the response validation structure. We can have another enhancement issue for that; And maybe close this one for now? 😃

MohabMohie (approved):

PR #852, opened 2023-02-02
Handle Null exception scrolling in iOS native

amlosman:


MohabMohie (dismissed):
thank you for your valuable contribution! hopefully the first of many :)

PR #859, opened 2023-02-13
Fix the issue of the report on mac Bug#858

magdyheiba:
Fix the issue of the report on mac

There were actually two things :- 
1- First of all it was CD'ing to generate_allure_report.sh which is No such file or directory so I changed the '' to "" 
Screenshot 2023-02-13 at 10 50 44 PM

After the Fix 

Screenshot 2023-02-13 at 10 51 33 PM

2- The other thing and the one that caused the issue is $parent_path this variable was read as String so it couldn't be appended the variable value well
Screenshot 2023-02-13 at 10 22 16 PM

And here is the generate_allure_report.sh After generation
Screenshot 2023-02-13 at 10 57 49 PM


MohabMohie (approved):

MohabMohie:
I'm surprised no one cried out that it's also broken on linux 😂
Thanks a lot for the clean fix 👌

PR #865, opened 2023-02-15
add swipeElementIntoView using text

Salma2018170:


src/main/java/com/shaft/gui/element/TouchActions.java
```diff
@@ -618,7 +618,19 @@ public TouchActions swipeElementIntoView(By scrollableElementLocator, By targetE
...
         return this;
     }
+      /**
+     * Attempts to scroll element into view using androidUIAutomator
+     *
+     * @param targetText
+     * @param driver
+     * @return a self-reference to be used to chain actions
+     */
+    public TouchActions swipeElementIntoView(String targetText, WebDriver driver) {
```
MohabMohie, line 628: Please let's only take the string value as input.
Inside the method you can use this code to get the current driver instance

`DriverFactoryHelper.getDriver().get()`

MohabMohie (changes requested):
Thanks a lot for your valuable contribution!
Kindly let's fix this small note and then we'll be able to merge directly.

PR #871, opened 2023-02-17
without format

Salma2018170:


MohabMohie (approved):
Thank you for your valuable contribution. This is an excellent feature and it will help a lot of Android users who are using SHAFT.

//TODO: handle fluent wait and implement timeout to locate the element or a synchronization point to handle the case if the page where you want to scroll is not yet visible

//TODO 2: always implement tests to cover the new code you introduce

//TODO 3: styling is still broken, if you're using IntelliJ make sure that you disable automatic code formatting when you create a commit

//TODO 4: learn how to update an existing PR rather than opening a new one

PR #889, opened 2023-02-22
comment publish test results till fixed

MohabMohie:


MohabMohie:
current implementation is failing all external PRs even dependabot
image
Should be fixed by following these steps:
https://github.com/marketplace/actions/publish-test-results#support-fork-repositories-and-dependabot-branches

but they are not straightforward, so let's stabilize the pipelines for now please.

PR #891, opened 2023-02-22
888 bug the appium is giving error no route found for wdhubstatus when using shaft and windows 10 cmd

MohabMohie:


MohabMohie:
pipeline is finally passing. mobile tests on browserstack are timing out because my account doesn't support parallel runs. TODO: create a ticket to modify the pipelines, merge them all together, create a separate pipeline for browserstack tests, force it to run in sequence.

PR #906, opened 2023-02-24
900 feature request scroll element into middle of viewport by default

MohabMohie:


MohabMohie:
In some cases scrolling the element to the top (or the bottom) of the view port will cause it to move under a sticky header or footer.

This change ensures that all scrolling actions are performed the same way (no more trying to do it with the Actions library) and that the target element is actually scrolled to the vertical and horizontal center of the viewport.

MohabMohie:
delete

PR #918, opened 2023-02-27
shadow dom element identification

MohabMohie:
- introduce the ShadowLocatorBuilder
- handle shadow dom unique element identification
- add css locator creation mode
- support xpath generation from webelement object

image
image
image


MohabMohie:
@MahmoudElSharkawy I made a big PR again... breaking them down and holding myself back when there's so many to fix is proving rather difficult...
I'll take ownership of this big PR... if you've reviewed it already then we're all good... if not, then you can focus on the SHADOW DOM handling, and ignore the rest of the minor changes, or just skim them... they are not that important... but they will add value later...

PR #928, opened 2023-03-03
Add tests to increase coverage / migrate built-in cucumber steps to use SHAFT driver / fix broken cucumber scenario

MohabMohie:


MohabMohie:
Codecov up to 45%

PR #929, opened 2023-03-05
update branch name

MohabMohie:


MohabMohie:
@MahmoudElSharkawy this change is to reflect the rename from 'master' to 'main' for our default branch
it will also enable the CI/CD pipeline to run and try to create a release with every push to the main branch.
in case the push contains a new version number for SHAFT, a release will be automatically created and deployed to jFrog. in case the version number already exists on GitHub, the job will fail.

MahmoudElSharkawy (approved):

PR #944, opened 2023-03-11
Fix the issue of GIF on firefox

magdyheiba:
So Basically the issue is on MozillaFirefox specifically , the issue was regarding the height of captured firstImage screenshot 
for some reason - may be update from firefox , I didn't try it on previous versions - after navigation and on pass action method while capturing the first screenshot 
the height appears :- (16) and after scaling (7) then appear at Allure report as described below 
Screenshot 2023-03-11 at 3 19 10 PM
Screenshot 2023-03-11 at 3 19 28 PM
Screenshot 2023-03-11 at 3 20 00 PM
Screenshot 2023-03-11 at 3 20 07 PM

So I first fixed it by setting the height from ScreenShotManager method but I didn't like the fix 
so what I did is I escaped the first screenshot to be taken after the first element action not after the redirection and that fixed the issue as described --> Height at firstImage --> 1560 and after scaling --> 660 then appear at Allure report as described below  
Screenshot 2023-03-11 at 3 25 28 PM
Screenshot 2023-03-11 at 3 25 36 PM
Screenshot 2023-03-11 at 3 27 40 PM
Screenshot 2023-03-11 at 3 27 31 PM 

Also Tested Chrome and working as expected 

Screenshot 2023-03-11 at 3 30 18 PM

** Added Features :- while trying to find the root cause I noticed we didn't have methods to get the height and width for the screen and I needed them so I added them 
If they are not important I can remove them 

** Last Note we should add an enhancement ticket to improve the performance of GIF and adding after each screenshot at it at firefox

### DIFF
--- a/src/main/java/io/github/shafthq/shaft/gui/image/ScreenshotManager.java
+++ b/src/main/java/io/github/shafthq/shaft/gui/image/ScreenshotManager.java
@@ -2,6 +2,7 @@ package io.github.shafthq.shaft.gui.image;
 
 import com.epam.healenium.SelfHealingDriver;
 import com.shaft.cli.FileActions;
+import com.shaft.gui.browser.BrowserActions;
 import com.shaft.tools.io.ReportManager;
 import io.github.shafthq.shaft.driver.DriverFactoryHelper;
 import io.github.shafthq.shaft.enums.Screenshots;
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/shafthq/shaft/gui/image/ScreenshotManager.java
```diff
@@ -562,6 +563,10 @@ private static void startAnimatedGif(byte[] screenshot) {
                         + testCaseName + ".gif";
                 gifRelativePathWithFileName = SCREENSHOT_FOLDER_PATH + SCREENSHOT_FOLDER_NAME + gifFileName;
 
+                // get the width and height of the current window of the browser
+                var height = Integer.valueOf(new BrowserActions().getWindowHeight());
+                var width = Integer.valueOf(new BrowserActions().getWindowWidth());
```
MohabMohie, line 568: this works (Y)
//TODO: optimize, since we already have the browser width and height when we initialize the browser, so there is no need to re-get them at this point. it's costing us two webdriver calls, and in the near future we should focus on reducing the number of webdriver calls as they have significant impact on performance.

MohabMohie (approved):
Thanks a lot for yet another valuable contribution Magdy (Y)

Issue #960, opened 2023-03-15
[Bug]: Getting exception on the logs related to the animated GIF and the images overlays themselves 


MahmoudElSharkawy:
**Describe the bug**

Getting exception on the logs related to the animated GIF and the images overlays themselves and the initial image is small (check the GIF and screenshot attached) with the latest version.

**Link to SHAFT_Engine Console logs**
https://gist.github.com/MahmoudElSharkawy/eb46402828b4d2d05c75411e339a9e5b

**Link to SHAFT_Engine Extent Report HTML**

Create a [GIST](https://gist.github.com) which is a paste of your _full_ extent report .html file, and link it here.

**To Reproduce**

1. Browser Firefox
2. Open google.com
3. Search on anything

**Screenshots**
![1f6d7074-7ca5-4d16-9859-ec27dc1981a7-attachment](https://user-images.githubusercontent.com/46620378/225361567-219d08f7-c444-4820-9b2e-3d94d3dc4205.gif)

![image](https://user-images.githubusercontent.com/46620378/225361530-d7f77bb0-da33-461f-ae16-4238797179c3.png)

![image](https://user-images.githubusercontent.com/46620378/225361044-ca918ffa-adfd-4e73-811a-51c6c165e463.png)

PR #964, opened 2023-03-19
Fix Bug #960 and enhance overlay images

magdyheiba:
here are the screenshots of chrome, firefox and Safari browsers
Screenshot 2023-03-19 at 4 43 32 PM
Screenshot 2023-03-19 at 4 42 02 PM
Screenshot 2023-03-19 at 4 42 49 PM

### DIFF
--- a/src/main/java/io/github/shafthq/shaft/driver/DriverFactoryHelper.java
+++ b/src/main/java/io/github/shafthq/shaft/driver/DriverFactoryHelper.java
@@ -23,10 +23,7 @@ import io.github.shafthq.shaft.tools.io.FailureReporter;
 import io.github.shafthq.shaft.tools.io.ReportManagerHelper;
 import io.github.shafthq.shaft.tools.support.JavaHelper;
 import io.qameta.allure.Step;
-import lombok.AccessLevel;
-import lombok.Getter;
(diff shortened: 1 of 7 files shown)


MohabMohie:
I don't see a watermark for the chrome screenshot. Also, please refrain from using text based browser name comparisons or creating special flows for certain browsers as this introduces inconsistency.

MohabMohie (changes requested):
I don't see a watermark for the chrome screenshot. Also, please refrain from using text based browser name comparisons or creating special flows for certain browsers as this introduces inconsistency.

magdyheiba:
As discussed with @MahmoudElSharkawy in a call we will remove the if/else browser specific and keep it this way to fix the current issue and we will open an enhancement ticket to restructure the GIF builder.

The current issue is in the first image captured by the browser changed from browser to another in terms of dimensions so that we have different in size GIFs across all browsers as we depend on the first screenshot to build GIF.

Currently the GIF of chrome is a bit small but it was originally small as well before the bug but on Safari and firefox more than perfect 

Here is the screenshot of chrome :- 
Screenshot 2023-03-22 at 10 47 13 AM

MohabMohie (approved):

PR #966, opened 2023-03-20
Add attachExtentReportToAllureReport property

MahmoudElSharkawy:
Add attachExtentReportToAllureReport property and it's false by default as it consumes alot of space as discussed in the Slack channel.


magdyheiba:
What do you thing adding it to user guide as well ?

MohabMohie (approved):

PR #980, opened 2023-03-26
Fix Bug#963 and adding Flags interface

magdyheiba:
Fix Bug#963 and adding Flags interface

when the flag is true 
Screenshot 2023-03-26 at 3 49 59 PM

and when the flag is false
Screenshot 2023-03-26 at 3 49 39 PM

### DIFF
--- a/AddNodeModules.bat
+++ b/AddNodeModules.bat
@@ -1,0 +1,9 @@
+@echo OFF
+
+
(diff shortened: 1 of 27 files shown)


MahmoudElSharkawy (approved):
3ash ya Magdy 👍

MohabMohie (approved):

PR #988, opened 2023-03-27
Bug fixes

MohabMohie:


MohabMohie:
@MahmoudElSharkawy please hold until tests look good enough.
Ps. I removed the test reporter plugin as it was causing failures. Let's plan to output the text summary from the execution logs as a commit comment as it'll be simpler.

MohabMohie:
Thought there are failures, they are consistent with what's on the main branch... Meaning that this branch is safe to merge. But failure investigation and tweaking is still need.
I will merge accordingly.

PR #991, opened 2023-03-28
Enhancements

MohabMohie:


MohabMohie:
@MahmoudElSharkawy I moved all the io-prefix packages to unify package names
added internal main package and internal sub-packages where needed to ensure that all categories make sense
exposed the listeners in the main com.shaft.listeners package, and moved any non-user facing packages to the internal main or sub package as needed

In addition to the above, I detected an issue with the safari execution failures, it's hitting an invalid element state exception despite my best attempts, I implemented a fix for that by moving the problematic actions to be performed inside the fluent wait block... thus fixing this issue for good.

Now the type action should be much faster and 100% robust across all systems.

This will be nearly impossible to review, that's why I split it into 3 commits.

- The first commit is for the element action movement inside the fluent wait block. -> this is important for you to understand the new structure and the potential impact moving forward if similar issues arise.
- The second commit is unreviewable since it includes all the classes and packages movements.
- The third commit is a smaller sample, showing the cucumber test runner listener moving into the com.shaft.listeners package so it gives you a feel of the changes that were made in the second commit.

PR #995, opened 2023-03-28
Open Execution Summary Report After Execution

MahmoudElSharkawy:


MahmoudElSharkawy:
> Don't forget to add to the flags class and the configuration manager please.
> 
> I also think we should disable allure by default and keep this one enabled by default, which also requires changes in the default properties file, flags class, and configuration manager

Done :D

MohabMohie (approved):

MohabMohie:
Looking good 👍
Let's check the tests, also I recommend upgrading the e2etests pipeline to upload the allure and summary, since we're disabling extent by default now...

MahmoudElSharkawy:
> Looking good 👍 Let's check the tests, also I recommend upgrading the e2etests pipeline to upload the allure and summary, since we're disabling extent by default now...

Done :D

MohabMohie (approved):
👌

PR #1001, opened 2023-03-28
Add caused by message to element actions error message when fails

MahmoudElSharkawy:
From this:
image
image

To this: image
image


MohabMohie:
I have a suggestion, checkout how it's being handled for browsers now. I discovered a javaclass that allows us to add the rootcause exception message. This will significantly reduce investigation time.

MohabMohie:
I will close this PR since it's been handled globally by the unified failure reporter.

PR #1003, opened 2023-03-29
Feature#994

magdyheiba:
Removed the set methods and sub classes from cucumber, log4j, and TestNG 
Deleted the capabilities class
Added a test class for that 
Fix the issue that caused because of changing the naming convention of Properties.web.mobileEmulation.isCustomeDevice()


MohabMohie (dismissed):
Looks excellent 👌
Thank you 😊

PR #1004, opened 2023-03-30
Trim failure reporter message after line break

MahmoudElSharkawy:


src/main/java/com/shaft/tools/io/internal/FailureReporter.java
```diff
@@ -12,7 +12,7 @@
 public class FailureReporter {
     public static void fail(Class<?> failedFileManager, String message, Throwable throwable) {
         String actionName = "fail";
-        String rootCause = " Root cause: \"" + Throwables.getRootCause(throwable).getLocalizedMessage() + "\"";
+        String rootCause = " Root cause: \"" + Throwables.getRootCause(throwable).getLocalizedMessage().split("\n")[0] + "\"";
```
MohabMohie, line 15: Better to use File.LineSeparator to maintain cross-platform compatibility

I think it's /r/n for linux

PR #1010, opened 2023-04-01
Create configurable Page Source attachment for WebDriver

MahmoudElSharkawy:
* Created new page Source section in the configurations manager
* New whenToTakePageSourceSnapshot property in the VisualValidations.properties and dded them to the visuals class 

image
image

### DIFF
--- a/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
+++ b/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
@@ -978,11 +978,11 @@ public class ElementActionsHelper {
             }
         }
 
-        if (driver != null && !SHAFT.Properties.visuals.whenToTakePageSourceSnapshot().equalsIgnoreCase("Never")) {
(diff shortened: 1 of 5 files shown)


MohabMohie (changes requested):
Can we set the default to never please?
Similar to the gif, this feature will have negative performance impact on the whole test run, so it makes sense to be disabled by default at least at first.

If you agree, please implement it and feel free to merge.

If not let's discuss with the TLC

MahmoudElSharkawy:
@MohabMohie I pushed some code enhancements to make it performance efficient (it wont take the snapshot unless it have to based on the configurations)..
Please check again and tell me if you have any more concerns.

MohabMohie:
I understand, but still an extra WebDriver call on the long run will have negative performance impact.
I'm fine with the implementation itself, I just think it should be disabled by default.

MohabMohie (approved):
Your call Mahmoud :)

PR #1033, opened 2023-04-06
release prep

MohabMohie:
- also migrated all the System.setProperty methods to set the coded SHAFT properties
- partial migration for System.getProperty to the  coded SHAFT properties
- removed property files that are not usually used by the user from the default properties folder.
- TODO: need to [create a user guide reference to all SHAFT properties](https://github.com/ShaftHQ/SHAFT_Engine_Docusaurus/issues/77) and how to use the coded or CLI command to manipulate them.


pom.xml
```diff
@@ -746,8 +746,7 @@
                     </archive>
                     <includes>
                         <include>com/**/*</include>
-                        <include>io/**/*</include>
-                        <include>resources/**/*</include>
+                        <include>resources/properties/default/*</include>
```
MohabMohie, line 749: @MahmoudElSharkawy we use these patterns to define which files should be included in SHAFT's jar. changing it to only the properties directory will remove other sub folders and thus save overall jar file size.

src/main/java/com/shaft/api/RestActions.java
```diff
@@ -76,7 +77,7 @@ public static Response getLastResponse() {
     }
 
     public RestActions(String serviceURI) {
-        initializeSystemProperties(System.getProperty("apiConnectionTimeout") == null);
+        initializeSystemProperties();
```
MohabMohie, line 80: I changed the method removing this parameter because it'll never be null since the properties manager classes have default values for all properties. @MahmoudElSharkawy

src/main/java/com/shaft/cli/FileActions.java
```diff
@@ -681,10 +683,10 @@ private boolean isTargetOSUnixBased() {
             }
         } else {
             // remote execution
-            String targetOS = System.getProperty("targetOperatingSystem");
-            if ("Windows".equals(targetOS)) {
+            String targetOS = SHAFT.Properties.platform.targetPlatform();
+            if (Platform.WINDOWS.name().equals(targetOS)) {
                 return false;
-            } else if ("Linux".equals(targetOS) || "Mac".equals(targetOS)) {
+            } else if (Platform.LINUX.name().equals(targetOS) || Platform.MAC.name().equals(targetOS)) {
```
MohabMohie, line 689: note that we should no longer use the platform name as a hardcoded string @MahmoudElSharkawy @magdyheiba

src/main/java/com/shaft/driver/internal/DriverFactoryHelper.java
```diff
@@ -59,7 +59,7 @@ public class DriverFactoryHelper {
     private static String TARGET_HUB_URL;
     private static Boolean MOBILE_EMULATION;
     private static Boolean MOBILE_EMULATION_CUSTOM_DEVICE;
-    private static Boolean LIGHTHOUSE_EXEUTION;
+    private static Boolean LIGHTHOUSE_EXECUTION;
```
MohabMohie, line 62: @Mohab-Osama @mohabosamaa we need to add the lighthouse execution parameter in the related properties manager class and the configuration manager html.

src/main/java/com/shaft/properties/internal/Web.java
```diff
@@ -38,15 +38,15 @@ private static void setProperty(String key, String value) {
...
     @Key("mobileEmulation.height")
     @DefaultValue("")
-    String mobileEmulationHeight();
+    int mobileEmulationHeight();
 
     @Key("mobileEmulation.pixelRatio")
-    @DefaultValue("")
-    String mobileEmulationPixelRatio();
+    @DefaultValue("1.0")
+    double mobileEmulationPixelRatio();
```
MohabMohie, line 49: @magdyheiba note that we can use int and double datatypes for parameters instead of using string and later casting it.

src/main/java/com/shaft/tools/io/internal/ExecutionSummaryReport.java
```diff
@@ -50,39 +50,52 @@ public static void generateExecutionSummaryReport(int passed, int failed, int sk
...
+                .replace("${TOTAL_ISSUES}", String.valueOf(ReportManagerHelper.getIssueCounter()))
+                .replace("${NO_OPEN_ISSUES_FAILED}", String.valueOf(ReportManagerHelper.getFailedTestsWithoutOpenIssuesCounter()))
+                .replace("${OPEN_ISSUES_PASSED}", String.valueOf(ReportManagerHelper.getOpenIssuesForPassedTestsCounter()))
+                .replace("${OPEN_ISSUES_FAILED}", String.valueOf(ReportManagerHelper.getOpenIssuesForFailedTestsCounters()))
+                .replace("${PASSED_DROPDOWN_OPTION}", StatusIcon.PASSED.getValue() + Status.PASSED.name())
+                .replace("${FAILED_DROPDOWN_OPTION}", StatusIcon.FAILED.getValue() + Status.FAILED.name())
+                .replace("${SKIPPED_DROPDOWN_OPTION}", StatusIcon.SKIPPED.getValue() + Status.SKIPPED.name())
+                .replace("${CASES_DETAILS}", detailsBuilder);
+
+        if (total > 0) {
```
MohabMohie, line 87: @MahmoudElSharkawy note that if you run no checkpoints (a test that only has steps without any verifications) the report will face "division by zero" and will output NaN error instead of the pie chart... it's a stupid issue, but I thought to fix it blmara.

PR #1045, opened 2023-04-11
1044 bug blocking issue whereby shaft overrides the cucumber properties file

MohabMohie:


MohabMohie:
all tests are passing. will force merge a hotfix

Issue #902, opened 2023-02-24
[Feature Request]: enhance switch to iframe usability


MohabMohie:
```java
//TODO: switch to iframe should be a browser action not an element action
        driver.get().browser().navigateToURL("https://github.com/ShaftHQ/SHAFT_ENGINE");
        var text = driver.get().element().switchToIframe(frame).getText(shaft_img);
```

PR #1049, opened 2023-04-12
Enhance Feature #902

magdyheiba:
- switch to iframe should be a browser action not an element action
- Implemented `SwitchToParentFrame()` as it is added as of selenium 4
- Implemented a method to get the current frame name
- Added a test class with the new way by using dummy html so I created an html page that contain parent and 2 children frames to test `switchToParentFrame()` it was great btw enjoyed it :D 
- Refactored this test class as it was outdated and commented `iFrameTest`

### DIFF
--- a/src/main/java/com/shaft/gui/browser/BrowserActions.java
+++ b/src/main/java/com/shaft/gui/browser/BrowserActions.java
@@ -5,6 +5,7 @@ import com.shaft.gui.element.AlertActions;
 import com.shaft.gui.element.SikuliActions;
 import com.shaft.gui.element.TouchActions;
 import com.shaft.gui.element.internal.FluentElementActions;
+import org.openqa.selenium.By;
 import org.openqa.selenium.Cookie;
 import org.openqa.selenium.WebDriver;
 import org.sikuli.script.App;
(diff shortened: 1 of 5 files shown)


src/main/java/com/shaft/gui/browser/BrowserActions.java
```diff
@@ -302,6 +303,43 @@ public static void switchToWindow(WebDriver driver, String nameOrHandle) {
...
     }
 
+    /**
+     * Switches focus to default content, is mainly used in coordination with
+     * {@link #switchToIframe(WebDriver, By)} to exit any iFrame layer and go back
+     * to the main page
+     *
+     * @param driver the current instance of Selenium WebDriver
+     */
+    @Deprecated
```
MohabMohie, line 313: Let's not add any deprecated methods please since we're only a few months away from deleting this class. 🙏

src/test/java/testPackage/iFrameTest.java
```diff
@@ -1,51 +1,54 @@
...
 import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
 
 public class iFrameTest {
-    // Declaring webdriver and excelreader instances
-    WebDriver driver;
+    SHAFT.GUI.WebDriver driver;
+    By iframe_1 = By.name("iframe1");
+    By iframe_2 = By.name("iframe2");
+    String testPage = "data:text/html,<script>var result;</script><button alt='Google' onclick='result=\"Clicked\"'>Go</button>\n" +
```
MohabMohie, line 13: try using multiline string to optimize this block

MohabMohie (approved):

PR #1074, opened 2023-04-26
fixing clickUsingJavascript failure reporting

zanoon2020:
fixing clickUsingJavascript failure reporting and adding related tests


MohabMohie (approved):
thanks for the excellent contribution! and the increased unit test coverage as well :) thank you!

PR #1089, opened 2023-05-01
iframe enhancements

MohabMohie:
- remove iframe methods from browser actions and fluent browser actions
- switch to frame before attempting to locate element until the user uses switch to default content.
- switch to frame before performing an action
- added element text parsing to ElementInformation class
- overrides native "switchTo().frame()" by setting LocatorBuilder.setIFrameLocator(elementInformation.getLocator());
- moved getCurrentFrame to FluentElementActions
- renamed iFrameTest to FrameTests and added more tests to cover more cases
- changed some "new ElementActions()" calls to "ElementActions.getInstance()" to make use of increased singleton performance -


MohabMohie:
PR is clean, will do a version bump and merge to release.

PR #1091, opened 2023-05-04
Junit support

MohabMohie:


MohabMohie:
obviously cli execution is completely broken... I'll test it.

MohabMohie:
@MahmoudElSharkawy 
WARNING: this change may not backwards compatible. Need to test it thoroughly with "Using_" projects.
Will promote to ALPHA release and merge after basic tests pass.

version bump preparing for alpha release "v7.2.20230505_ALPHA"

in theory this alpha release should be backwards compatible, but it may also be completely blocking :D that's why it's an alpha.

if everything goes well, existing users should feel no change, new users should be able to run without configuring their testng listeners, and junit users should also be able to run without changing or configuring anything...

if not, nothing will run for either new or existing users...
well, I guess we'll just have to wait for the tests, merge, release, and then find out.

PR #1097, opened 2023-05-06
Parallelizing BStacktests

MohabMohie:
Browser stack extended and upgraded their sponsored licence to x5 parallel runs.


MohabMohie:
100% pass rate. Ready to merge.

PR #1132, opened 2023-05-24
Fix allure report generation with localhost instead of using ip address

Waleedmohammed:


MohabMohie (approved):
Thank you for your excellent contribution!

PR #1191, opened 2023-07-07
Beta Version Release

zanoon2020:
type method enhancement 
adding attemptToClickBeforeTyping flag


src/main/java/com/shaft/properties/internal/Internal.java
```diff
@@ -10,7 +10,7 @@
 })
 public interface Internal extends EngineProperties {
     @Key("shaftEngineVersion")
-    @DefaultValue("SHAFT v7.3.20230703")
+    @DefaultValue("7.3.20230707_BETA")
```
MohabMohie, line 13: Please add "SHAFT " as a prefix similar to the old value.

MohabMohie (approved):

PR #1238, opened 2023-08-20
properties to set window size globally

georgii-musikhin:
- added two params `browserWindowWidth` and `browserWindowHeight`
- set window size in DriverFactoryHelper
- update WebTests
- create two input params to index.html


MohabMohie (approved):
Amazingly complete PR 👏
Thank you!

MohabMohie:
Please check the test results, I believe the code is just missing x,y which can be set to zero and zero @georgii-musikhin

georgii-musikhin:
@MohabMohie sorry, but I didn't get you
Do you mean that with unit tests the windows size will be 0x0?
I've run then and it was 1920x1080, because they are default values

Issue #1240, opened 2023-08-21
[Feature Request]: Map test dat from Json file to Java Object


MamdouhMansour:
**Is your feature request related to a problem? Please describe.**

The JsonFileManager methods parse the json object from json file in form of String array, which is not proper to map it to a Java Object Class

Same like => `[full_name:SHAFT, phone:123456, birthday:20-20-2022, governorate_id:1, city_id:1, country_id:1, email:x, password:x, passwordConfirm:x, type:x, gender:x, id_number:y]`

Json file => 
`{
  "data": {
    "jsonData": {
      "full_name": "SHAFT",
      "phone": "123456",
      "birthday": "20-20-2022",
      "governorate_id": 1,
      "city_id": 1,
      "country_id": 1,
      "email": "x",
      "password": "x",
      "passwordConfirm": "x",
      "type": "x",
      "gender": "x",
      "id_number": "y"
    }
  }
}`
**Describe the solution you'd like**

Add function getTestDataAsJson 

**Describe alternatives you've considered**

Added function getTestDataAsJson

MamdouhMansour:
@MohabMohie  assign it to me please

PR #1241, opened 2023-08-21
Feature/1240 map test dat from json file to java object

MamdouhMansour:


MamdouhMansour:
Related to #1240

src/main/java/objectclass/BodyObject.java
MohabMohie, line 1: Question...
Is it possible to build this class dynamically in run time after parsing the json file?
@MhmdElGazzar supervised a similar implementation with the CuRL to POJO generator, is it possible to create a dynamic json to POJO parser? better-yet, did someone already do this and can we just use it as a dependency?

To me, it makes little sense having a static mapping with static json property names for test data... we have a similar static implementation for SHAFT properties, but for test data I think we need a dynamic implementation.


aside from that, this looks like an excellent implementation. Thanks Mamdouh.

MohabMohie (changes requested):
one minor change request, can we move the BodyObject.java to the src/test/java directory since it's not a part of the engine (because it has static mapping of the specific test data we're using to test this feature).

src/main/java/objectclass/BodyObject.java
MamdouhMansour: @MhmdElGazzar @MohabMohie would like to know more to implement

MamdouhMansour:
@MohabMohie moved pojo out of the engine

MamdouhMansour:
@MohabMohie what we need to complete this PR?

MohabMohie (approved):

PR #1243, opened 2023-08-22
New feature to disable cache on Firefox, Microsoft Edge and Google chrome browsers

ieZaky:
Added a feature to disable cache using flag (disableCache) on the following browsers:

1. Google chrome 
2. Microsoft Edge
3. Firefox
4. Safari

**Note: I could not find a valid website to make a test case for this feature however, this feature was tested locally, and it works fine on mentioned browsers except Safari due to the macOS availability.**
@MahmoudElSharkawy

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
 
     <properties>
         <!-- Dependencies and Plug-ins Version Numbers Parameters -->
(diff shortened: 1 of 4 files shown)


MahmoudElSharkawy (approved):
Thank you for your valuable contributions. First one from many to come isA!

PR #1258, opened 2023-08-27
Added a feature to get all the rows data for a table

osamaayman1005:
**Purpose** This feature allows users to easily get the rows data for any table that follow the standard html table format: which should have <thead> tag that has all labels in <th> tags and <tbody> which should have all rows in <tr> tags and all the data in <td> tags. for tables. 
**Note** that this feature will only work if the html table follows this standard and does not guarantee results if any tag is missing


MohabMohie (approved):
Thank you for your valuable contribution.
Kindly consider adding this action to the user guide so that more users are aware of it. https://shafthq.github.io/docs/Keywords/GUI/Element_Actions
it's very easy to edit (just .md text)

PR #1265, opened 2023-09-01
Attempting to fix jacoco by reversing argline changes

MohabMohie:
Jacoco was broken when this fix for eclipse was implemented.
1. Try to leave the fix and still define argline where it should be.
2. If this fails, will try to remove the argline from testng and junit (but that will break eclipse again)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1018,9 +1018,7 @@
                             </dependency>
                         </dependencies>
                         <configuration>
-                            <argLine>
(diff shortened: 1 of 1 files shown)


MohabMohie:
@zanoon2020 I found that ever since we made this change to argline it broke code coverage reporting.
(it was made to fix something related to eclipse support back on june 25th)
I reversed the fix and code coverage is now working.
Would you kindly check if everything is working fine on eclipse?

PR #1274, opened 2023-09-01
Added a class to get a unique locator from a generic locator

osamaayman1005:
**Problem:** Making actions (click, select, type,...) on a locator that may have multiple values is not supported
**Use Case:** You have a menu page and you can create an items in that menu. you created the item and you need to click on it to edit it. the newly created item does not have a unique identifier and you will not be able to add the item locator in the page since the items are dynamic.
**Proposed Solution:** Create a class that can handle the generic locators and returns a unique locator by adding a unique class name into the element using JavaScript. and you can get this element by:
1. full text
2. partial text
3. index

**Limitation:** Since it's adding a class name by scripting, refreshing the page or navigating to another page will make the class name get removed so the method will need to be called again.

### DIFF
--- a/src/main/java/com/shaft/gui/internal/locator/MultiLocatorHandler.java
+++ b/src/main/java/com/shaft/gui/internal/locator/MultiLocatorHandler.java
@@ -1,0 +1,95 @@
+package com.shaft.gui.internal.locator;
+
+import com.shaft.driver.internal.DriverFactoryHelper;
+import com.shaft.gui.element.internal.ElementActionsHelper;
+import com.shaft.tools.io.ReportManager;
+import org.openqa.selenium.By;
+import org.openqa.selenium.JavascriptExecutor;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.Wait;
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import java.time.Duration;
+import java.util.List;
+
+public class MultiLocatorHandler {
+    private static int uniqueNumber = 1;
+
+    /**
+     * Get a unique By locator for an element within a list of multiple elements based on exact text match.
+     *
+     * @param multipleElementsLocator The locator for the list of elements.
+     * @param visibleTextOfExactElement The exact text content to match.
+     * @return A unique By locator for the specific element found, or null if not found.
+     * in case multiple options found, returns the first one
+     */
+    public static By getLocatorByText(By multipleElementsLocator, String visibleTextOfExactElement) {
+        waitAndHandleExceptions(multipleElementsLocator);
+
+        List<WebElement> elements = DriverFactoryHelper.getDriver().get().findElements(multipleElementsLocator);
+        for (WebElement element : elements) {
+            if (element.getText().equals(visibleTextOfExactElement)) {
+                String className = injectUniqueClassName(element);
+                return By.className(className);
+            }
+        }
+        ReportManager.logDiscrete("element with text \"" + visibleTextOfExactElement + "\" was not found");
+        return null;
+    }
+    /**
+     * Get a unique By locator for an element within a list of multiple elements based on partial text match.
+     *
+     * @param multipleElementsLocator The locator for the list of elements.
+     * @param partialTextOfExactElement The partial text content to match.
+     * @return A unique By locator for the specific element found, or null if not found.
+     * in case multiple options found, returns the first one
+     */
+    public static By getLocatorContainingText(By multipleElementsLocator, String partialTextOfExactElement) {
+        waitAndHandleExceptions(multipleElementsLocator);
+
+        List<WebElement> elements = DriverFactoryHelper.getDriver().get().findElements(multipleElementsLocator);
+        for (WebElement element : elements) {
+            if (element.getText().contains(partialTextOfExactElement)) {
+                String className = injectUniqueClassName(element);
+                return By.className(className);
+            }
+        }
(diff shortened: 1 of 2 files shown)


MohabMohie:
@osamaayman1005 Can you kindly explain the problem that this class will help resolve further?
What I see from the implementation is that it's looping to find the element using its text, partial text, or index. Why not just use an xpath? Or even simpler, the locator builder?

@MahmoudElSharkawy and @magdyheiba for your kind advice.

MahmoudElSharkawy:
@osamaayman1005 
I believe that we can do this just by simply have a method that returns the locator needed and just pass the dynamic value to it.
It should be something like this:
```
 private final By googleSearchResult_h3 (String index) { 
      return By.xpath("(//h3[contains(@class,'LC20lb')])[" + index + "]");
 }
```
Can you please try this and let us know if this will help you in your case or this is something different that needs your implementation?

osamaayman1005:
@MohabMohie Thank you for your review and feedback.
I will explain to you the use case that i faced.
So i was creating a test case for item creation in a menu page, the text case is:
_Verify that when an item is created it is added correctly to the menu._
the arrangement of the menu is by category and alphabetical. i wanted to create a method in my POM that can click on the newly created item menu. i can do this inline in the method by using xpath as you said like "/div [contains(text(),'" + itemName+"')]", this will only work if the itemName is unique. and i will not be able to add a property to the POM like: private By item = ...; as a user, it was easier for me to have a method that takes a generic locator and a specification and returns the locator for a certain element, so instead of writing the xpath for every time i want a unique element from list of elements, i thought it would be easier to have some sort of generic method that can do that for me. also having the locator as a member variable instead of writing the xpath in every method i need it is much more organized from my point of view.

If i missed an easier solution for this problem please let me know, since i just started using shaft and i am not familiar with LocatorBuilder and i couldn't find a specific solution for me problem.

MahmoudElSharkawy:
> @MohabMohie Thank you for your review and feedback. I will explain to you the use case that i faced. So i was creating a test case for item creation in a menu page, the text case is: _Verify that when an item is created it is added correctly to the menu._ the arrangement of the menu is by category and alphabetical. i wanted to create a method in my POM that can click on the newly created item menu. i can do this inline in the method by using xpath as you said like "/div [contains(text(),'" + itemName+"')]", this will only work if the itemName is unique. and i will not be able to add a property to the POM like: private By item = ...; as a user, it was easier for me to have a method that takes a generic locator and a specification and returns the locator for a certain element, so instead of writing the xpath for every time i want a unique element from list of elements, i thought it would be easier to have some sort of generic method that can do that for me. also having the locator as a member variable instead of writing the xpath in every method i need it is much more organized from my point of view.
> 
> If i missed an easier solution for this problem please let me know, since i just started using shaft and i am not familiar with LocatorBuilder and i couldn't find a specific solution for me problem.

Do you have a search feature for the menu items? If so, you can simply search for the item you just created and make sure it's displaying.
otherwise you can use the method I mentioned as you already know the item's name to use when calling the dynamic locator.
Does this help you in your case?

osamaayman1005:
@MahmoudElSharkawy 
Yes this will Definity works. my solution is more complicated. but it will hide the complexity of finding an element from list of elements from the POM (applying single responsibility principle). so user will not worry about finding a specific element that has other duplicates and will simply call a method that will do this for them.
`        By uniqueLocator = MultiLocatorHandler.getLocatorByIndex(genericElement, 3);
        driver.element().click(uniqueLocator);
`

Correct me if am mistaken

osamaayman1005:
@MahmoudElSharkawy 
Yes this will definitely help my case. i just want to make an easier substitution from a user perspective

MohabMohie:
@osamaayman1005 

You can create a member variable without assigning value to it in your POM.

```java
By listItem_label;
```

And later assign value to it and consume it inside the related method as @MahmoudElSharkawy mentioned.

```java
public void assertMenuItemText(int index, string expectedText){
listItem_label = By.xpath("(//div[@class='menuItem'])["+index+"]"); 
driver.element().assertThat(listItem_label).text().isEqualTo(expectedText).perform();
}
```

This is the traditional approach using Selenium, it will work to solve your use case without any complex code.

The SHAFT / LocatorBuilder approach will be as follows.

For declaration you will declare the common/static part of the locator

```java
var listItem_label = SHAFT.GUI.Locator.byTagName("div").hasClass('menuItem');
```

And then for consumption

```java
public void assertMenuItemText(int index, string expectedText){
driver.element().assertThat(listItem_label.hasIndex(index).build()).text().isEqualTo(expectedText).perform();
}
```

Using the locator builder you'll be able to meet your need and also keep the static part of the locator centralized with the rest of the locator declarations. This will help you identify dynamic locators in your POM class declarations easily.

And in both cases you'll just call the related method and you'll pass the same values.


Now if you want to get the list of items and ensure that they're in alphabetical order, this is logical validation and should be done via the API layer of course... Since the UI just shows the items and doesn't sort them.

So the proper check would be to perform and API request to get the menu, and then assert that the returned list of names is equal to the expected list of names. You can use SHAFT to do that and extract the list of names easily using a JSON path.

Would you kindly try one of the aforementioned solutions and tell us if it works for you?

Please note that it's not recommended to add complex code to test complex code, that's why we never embed loops and conditions inside SHAFT or even in our own test or POM classes.

osamaayman1005:
@MohabMohie 
I have tried 'SHAFT.GUI.Locator' and this serves my use case perfectly. only comparing it with my solution, the locator builder has O(1) while my solution has O(n). 
Thank you so much @MohabMohie and @MahmoudElSharkawy
 i hope i didn't waste too much of your time

PR #1310, opened 2023-10-03
"Exception Fix for SelectMethod"

MustafaAgamy:


MohabMohie:
kindly fix this pr @MustafaAgamy 
it should only show one new commit not 12

PR #1337, opened 2023-10-20
Selenium Webdriver Explicit waits wrapping

MahmoudElSharkawy:
Added some on the most important waits in to the browser and element actions (supporting the fluency wizard) and added a custom method to be used anytime to have the flexibility for creating any custom waits including and & or methods if/when needed by the users.
Usage should be something like this:
image

### DIFF
--- a/src/main/java/com/shaft/driver/SHAFT.java
+++ b/src/main/java/com/shaft/driver/SHAFT.java
@@ -77,7 +77,7 @@ public class SHAFT {
                 return new AlertActions();
             }
 
-            public WaitActions SeleniumExplicitWaits(Function<? super org.openqa.selenium.WebDriver, ?> conditions) {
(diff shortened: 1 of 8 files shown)


MustafaAgamy (dismissed):

MohabMohie:
@MahmoudElSharkawy 
Instead of "CustomExplicitWaits" can we call the method "waitUntil"? I think it'll be more readable and consistent with existing fluent keywords.

Other than that, this is an excellent idea and an excellent implementation. Thank you.

MahmoudElSharkawy:
> @MahmoudElSharkawy Instead of "CustomExplicitWaits" can we call the method "waitUntil"? I think it'll be more readable and consistent with existing fluent keywords.
> 
> Other than that, this is an excellent idea and an excellent implementation. Thank you.

@MohabMohie Done :D .. I also added some java docs for the `waitUltil` method. Please check :D

MohabMohie (approved):
AMAZING.
#flawlessVictory

PR #1350, opened 2023-10-31
"EnhancmenetsAndFixes"

MustafaAgamy:
(SelectMethod - DragAndDrop) Exception fixes and Flags added


src/main/java/com/shaft/gui/element/ElementActions.java
```diff
@@ -701,34 +701,31 @@ public ElementActions keyPress(By elementLocator, Keys key) {
      */
     public ElementActions select(By elementLocator, String valueOrVisibleText) {
         ElementInformation elementInformation = ElementInformation.fromList(ElementActionsHelper.
-                identifyUniqueElement(DriverFactoryHelper.getDriver(), elementLocator));
+                identifyUniqueElement((WebDriver) DriverFactoryHelper.getDriver(), elementLocator));
```
MahmoudElSharkawy: I need to know why you casting the WebDriver while it's not needed previously.

MustafaAgamy: Yeah I forgot about not using threadlocal driver version, so it's not really needed any more but it won't cause any issues, Will Remove it

MahmoudElSharkawy: It's better do it now because we always tend to forged these things :D

MustafaAgamy: Casting removed 😄

MahmoudElSharkawy:
select method is failing in the pipeline and here's the error, please check. 
I notice the you changed some logic; I thought this would the property bug only.
image

MustafaAgamy:
Whats select method added logic is doing
Its checking if we've an Element with the given text below the dropdown element itself or not "for now "

Maybe later we can enhance the logic and make it better but that's whats its actually doing its using Relative locators  to find an Element with the given text below it

MahmoudElSharkawy:
@MustafaAgamy 
So this error is expected and the logic is discussed with @MohabMohie? ... If you, then you can add the expected error param like this `@Test(expectedExceptions = {AssertionError.class})` so it's not failing in the pipeline.

MustafaAgamy:
What is the TC being execute In the pipelines?

MahmoudElSharkawy:
> What is the scenario of the TC being executed In the pipelines?

@MustafaAgamy Check the screenshot I shared in the previous comment..You can also check the pipeline reports.

MustafaAgamy:
What is the TC being execute In the pipelines?

MustafaAgamy:
Done expected exception added
Should work now

src/test/java/testPackage/appium/DragAndDropTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testng.annotations.Test;
+
+public class DragAndDropTest {
+    private SHAFT.GUI.WebDriver driver;
+    private final AppiumBy dragBtn = new AppiumBy.ByAndroidUIAutomator("UiSelector().text(\"Drag\")");
+    private final By toBeDragged =  By.xpath("//android.view.ViewGroup[@content-desc=\"drag-l1\"]/android.widget.ImageView");
+    private final By toBeDropped =By.xpath("//android.view.ViewGroup[@content-desc=\"drop-l1\"]/android.view.ViewGroup");
+    @BeforeMethod
+    public void setupUp(){
+        System.setProperty("mobile_autoGrantPermissions", "true");
```
MohabMohie: Please note that any properties you set in your test must be unset after it. Else it will impact the rest of the execution pipeline.
I suggest using before/after configuration methods.

Also kindly make sure that your new test class is excluded from the desktop/web testing jobs, and included in the android native testing jobs.

You'll find the full pipeline configuration under the .github folder

MohabMohie (changes requested):

src/test/java/testPackage/appium/DragAndDropTest.java
MustafaAgamy: Okay I'll work on it today In-sha-Allah

Thanks.

MohabMohie: @MustafaAgamy @MahmoudElSharkawy are we ready to merge?

MustafaAgamy: I wanted to ask about the Pipeline part
I already mentioned it on slack 😅
If it's not clear yet maybe we can Remove the dragAndDrop test for now

MohabMohie: Kindly comment the class for now

MustafaAgamy: Done I removed the entier class file

PR #1362, opened 2023-11-09
fixing clear method in element actions

zanoon2020:
Dear @MohabMohie , for your recommendations , i commented the part of clearing using selenium native method and only left clearing using back space .

### DIFF
--- a/src/main/java/com/shaft/gui/element/ElementActions.java
+++ b/src/main/java/com/shaft/gui/element/ElementActions.java
@@ -935,7 +935,8 @@ public class ElementActions {
                         ElementActionsHelper.performActionAgainstUniqueElement(DriverFactoryHelper.getDriver(), elementInformation.getLocator(), ElementAction.BACKSPACE);
                     }
                 }
-                if (currentText.isBlank()) {
+                var currentTextAfterClearingUsingBackSpace = getText(elementLocator);
+                if (currentTextAfterClearingUsingBackSpace.isBlank()) {
                     ElementActionsHelper.passAction(DriverFactoryHelper.getDriver(), elementLocator, Thread.currentThread().getStackTrace()[1].getMethodName(), "", null, elementName);
                 } else {
                     ElementActionsHelper.failAction(DriverFactoryHelper.getDriver(), "Expected to clear existing text, but ended up with: \"" + currentText + "\"",
--- a/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
+++ b/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
@@ -654,11 +654,11 @@ public class ElementActionsHelper {
 
     private static void clearBeforeTyping(ElementInformation elementInformation, TextDetectionStrategy successfulTextLocationStrategy) {
         // attempt clear using the webElementReference, else try to find the element again
-        try {
-            (elementInformation.getFirstElement()).clear();
-        } catch (WebDriverException webDriverException) {
-            performActionAgainstUniqueElement(DriverFactoryHelper.getDriver(), elementInformation.getLocator(), ElementAction.CLEAR);
-        }
+//        try {
+//            (elementInformation.getFirstElement()).clear();
+//        } catch (WebDriverException webDriverException) {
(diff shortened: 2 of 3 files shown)


MahmoudElSharkawy:
@zanoon2020 we have an issue related to the change.
I want to ask please; Why are we doing this change?, what's wrong with the Selenium clear method?
image

zanoon2020:
> @zanoon2020 we have an issue related to the change.
> I want to ask please; Why are we doing this change?, what's wrong with the Selenium clear method?
> image
> 

@MahmoudElSharkawy with some angular input elements this method cause type method to type only 1 string , we investigated and found the issue in selenium clear method , in shaft we are clearing the field by default before type although we have flag for clearing before typing , why we have the flag if we are trying to clear first anyway , so i commented the part that clearing any way and left the related part to the flag as suggestion not final solution .

MahmoudElSharkawy:
> > @zanoon2020 we have an issue related to the change.
> > I want to ask please; Why are we doing this change?, what's wrong with the Selenium clear method?
> > image
> 
> @MahmoudElSharkawy with some angular input elements this method cause type method to type only 1 string in firefox , we investigated and found the issue in selenium clear method , in shaft we are clearing the field by default before type although we have flag for clearing before typing , why we have the flag if we are trying to clear first anyway , so i commented the part that clearing any way and left the related part to the flag as suggestion not final solution .

@zanoon2020  Can't we just move the flag?

zanoon2020:
@MahmoudElSharkawy the method first clearing using default clear method the trying to clear letter by letter using back space , we will need to update the logic if clearing using clear method pass then not to try clearing letter by letter , but the question is why we should have two methods to clear the element of one of them is working 100%

zanoon2020:
@MahmoudElSharkawy the method first clearing using default clear method the trying to clear letter by letter using back space , we will need to update the logic if clearing using clear method pass then not to try clearing letter by letter , but the question is why we should have two methods to clear the element of one of them is working 100%

zanoon2020:
@MahmoudElSharkawy i have another solution, to call shaft clear method from element actions , but didn't know how to 😁.anf if the flag is enabled. Then clear logic will be applied

zanoon2020:
@MahmoudElSharkawy i see all E2E tests pass , how can i get to the failed one , is it from my pull request ?

MahmoudElSharkawy:
> @MahmoudElSharkawy i see all E2E tests pass , how can i get to the failed one , is it from my pull request ?

yes, check the triggered cases from your PR. I just created a new PR and it's passing.

MahmoudElSharkawy:
@zanoon2020 
Can't we just have it like this
```
If(clearBeforeTyping) {
    if(backSpaceMethod) {
      // backspace login
    } else {
          // use clear method
           }
}
```

zanoon2020:
@MahmoudElSharkawy the issue because type method was clearing first before typing By default and the flag was only for clearing using backspace , there is 2 solutions for this , one to set the clearing using backspace flag default value to true , second to fix the method clearing before typing itself ,

zanoon2020:
> @zanoon2020 Can't we just have it like this
> 
> ```
> If(clearBeforeTyping) {
>     if(backSpaceMethod) {
>       // use backspace method
>     } else {
>           // use clear method
>      }
> }
> ```
we have one flag attemptClearBeforeTypingUsingBackspace , and the typing method always to clear first

zanoon2020:
@MahmoudElSharkawy let us take step back , is correct behavior for type method is trying to clear first by default ? or it's optional

MahmoudElSharkawy:
> @MahmoudElSharkawy let us take step back , is correct behavior for type method is trying to clear first by default ? or it's optional

It should be optional and enabled by default in the properties.

zanoon2020:
> > @MahmoudElSharkawy let us take step back , is correct behavior for type method is trying to clear first by default ? or it's optional
> 
> It should be optional and enabled by default in the properties.

exactly , tests failed because it's done hard coded and the default value of  flag is false 
image

MahmoudElSharkawy:
@zanoon2020 
let's do the flag then :D

PR #1369, opened 2023-11-19
Type revamp

zanoon2020:
type method has been revamped

### DIFF
--- a/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
+++ b/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
@@ -7,6 +7,7 @@ import com.shaft.driver.internal.DriverFactoryHelper;
 import com.shaft.enums.internal.ClipboardAction;
 import com.shaft.enums.internal.ElementAction;
 import com.shaft.gui.browser.internal.BrowserActionsHelper;
+import com.shaft.gui.element.ElementActions;
 import com.shaft.gui.element.SikuliActions;
 import com.shaft.gui.internal.exceptions.MultipleElementsFoundException;
 import com.shaft.gui.internal.image.ImageProcessingActions;
@@ -673,6 +674,8 @@ public class ElementActionsHelper {
         }
     }
 
+
+
     private static String confirmTypingWasSuccessful(ElementInformation elementInformation, TextDetectionStrategy successfulTextLocationStrategy) {
         //get a fresh instance of the element
         var updatedElementInformation = ElementInformation.fromList(identifyUniqueElementIgnoringVisibility(DriverFactoryHelper.getDriver(), elementInformation.getLocator()));
@@ -682,6 +685,8 @@ public class ElementActionsHelper {
         }
         return readTextBasedOnSuccessfulLocationStrategy(updatedElementInformation, updatedSuccessfulTextLocationStrategy);
     }
+
+
 
     private static TextDetectionStrategy determineSuccessfulTextLocationStrategy(ElementInformation elementInformation) {
         if (DriverFactoryHelper.isMobileNativeExecution()) {
@@ -737,6 +742,9 @@ public class ElementActionsHelper {
         }
         return "";
     }
+
+
+
 
     public static boolean performClipboardActions(WebDriver driver, ClipboardAction action) {
         try {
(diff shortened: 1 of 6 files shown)


zanoon2020:
Dears , @MohabMohie , @MahmoudElSharkawy , @magdyheiba , @MhmdElGazzar this message for reference , it's the sudo code
https://shaft-engine.slack.com/archives/C04SAA8BWCT/p1700142762303159?thread_ts=1700127523.414399&cid=C04SAA8BWCT

src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
```diff
@@ -819,6 +827,115 @@ public static String typeWrapper(ElementInformation elementInformation, String t
...
+        var elementInformation = ElementInformation.fromList(ElementActionsHelper.performActionAgainstUniqueElement(DriverFactoryHelper.getDriver(), elementLocator));
+        var attemptClearBeforeTyping = SHAFT.Properties.flags.attemptClearBeforeTyping();
+        var attemptClearBeforeTypingUsingBackspace = SHAFT.Properties.flags.attemptClearBeforeTypingUsingBackspace();
+        if (attemptClearBeforeTyping && attemptClearBeforeTypingUsingBackspace) {
+            clearBeforeTypingUsingBackSpace(elementInformation);
+        }
+        else if (attemptClearBeforeTyping) {
+            clearBeforeTypingUsingNativeClear(elementInformation);
+        }
+        else if(attemptClearBeforeTypingUsingBackspace){
```
MahmoudElSharkawy: I think this part is redundant.
My suggestion this so it wont go to the condition if the `attemptClearBeforeTyping ` property is disabled from the beginning:

```
if (attemptClearBeforeTyping) {
        if (attemptClearBeforeTypingUsingBackspace) {
             clearBeforeTypingUsingBackSpace(elementInformation);
        }
        else {
              clearBeforeTypingUsingNativeClear(elementInformation);
        }
}
```

src/main/java/com/shaft/properties/internal/Flags.java
```diff
@@ -34,13 +34,17 @@ private static void setProperty(String key, String value) {
     boolean forceCheckElementLocatorIsUnique();
 
     @Key("forceCheckTextWasTypedCorrectly")
-    @DefaultValue("false")
```
MahmoudElSharkawy: Why did we change the value of this flag?

MahmoudElSharkawy (changes requested):
Great PR ya @zanoon2020. I added some small comments.
Last this I would suggest is to remove the working 'new' to the methods name as this will be the current type methods, just to minimize any confusion in the future.

src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
AhmdZanoon: > I think this part is redundant. My suggestion this so it wont go to the condition if the `attemptClearBeforeTyping ` property is disabled from the beginning:
> 
> ```
> if (attemptClearBeforeTyping) {
>         if (attemptClearBeforeTypingUsingBackspace) {
>              clearBeforeTypingUsingBackSpace(elementInformation);
>         }
>         else {
>               clearBeforeTypingUsingNativeClear(elementInformation);
>         }
> }
> ```

@MahmoudElSharkawy if the attemptToclearBeforeTyping is false and using back space is true then no clearing will happen , i think we should seprate flags from each other , what do u think

zanoon2020:
> Great PR ya @zanoon2020. I added some small comments. Last this I would suggest is to remove the wording 'new' to the methods name as this will be the current type methods, just to minimize any confusion in the future.

dear @MahmoudElSharkawy the word new is temp as i intend to remove it once i launch the beta version , as i suggest to go with beta version first after this PR

src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
AhmdZanoon: done

MohabMohie:
@MahmoudElSharkawy are we ready to merge this PR yet? Feel free to raise your findings please.

MahmoudElSharkawy (approved):

MahmoudElSharkawy:
> @MahmoudElSharkawy are we ready to merge this PR yet? Feel free to raise your findings please.

I'm good :D

MahmoudElSharkawy:
> @MohabMohie @MahmoudElSharkawy is it ok to merge 😊do i need to do any kind of update for the branch before merging 

@zanoon2020 you can check the pipeline reports, if they are passing then we can merge and have a release 😃🙏

MahmoudElSharkawy (approved):

PR #1395, opened 2023-12-07
1393 feature request enable single file allure report generation

MohabMohie:
I really don't want to merge this. It's not just the full execution log that's broken... It's all the setup/teardown attachments...

@zanoon2020 can you help me by collaborating on this branch?
I think we should create a blank project using allure, testng, and selenium. see if the same issue exists when everything is native.
If it does, we need to work around it, or fix it.

In the meantime, I think we should roll-back the main branch to use the older allure binaries.


zanoon2020:
I already did it , all native , anything under @before suit not reported , i opened new issue on allure2 repo yesterday.

zanoon2020:
for main branch all we need is to keep version in internal property file to 2.21.0

MohabMohie:
@zanoon2020 can you merge your logic to have the logo in the single file to this branch please?

zanoon2020:
@MohabMohie in 30 min will be done.

zanoon2020:
@MohabMohie done from my side

MohabMohie:
@zanoon2020 I'm merging main branch into this one and resolving conflicts.

Now we only need to figure out what to do for the missing attachments.

Let's discuss via slack.

MohabMohie:
@zanoon2020 we agreed that the following action would be to pinpoint the drawbacks of this upgrade
- before/after suite
- before/after test
- before/after class
- before/after method
- validation method (pass/fail) attachment
- element action failed attachment
- browser action failed attachment

zanoon2020:
@MohabMohie all working fine except before suit after suit 

Attachments in nested steps working fine 
Before class and after class logged correctly 
Before method and after method working correctly 
Validation attachment fine

PR #1397, opened 2023-12-07
Fix Lambdatest issues in iOS

magdyheiba:
- Simply they deleted the old app url so the app was not found 
- added the flag to disable the clearing before typing in Mobile element till it's fixed


MohabMohie (approved):
Thanks Magdy 🤍

PR #1429, opened 2023-12-19
Complete core restructure for multiple driver management

MohabMohie:


MohabMohie:
the only user facing change here is that .getInstance will no longer be an option that the user can choose.

This code will no longer work:
```java
BrowserActions.getInstance().closeCurrentWindow();
```

This code will be the replacement:
```java
new BrowserActions(driver).closeCurrentWindow();
```

This is a reuired change that I could not work around.

PR #1431, opened 2023-12-24
restructuring validation classes

MohabMohie:
The objective is to be able to chain validations after calling . perform();

This will require a complete restructuring of the validation helpers/builders/executors.

A recommended approach is to separate web browser and web element executors first... Then create a new webnative class that extends the native validations builder class.

Lastly have each executor return an object of the relevant validation builder...

It's an experiment and the pr will be merged with the issue closed as "won't do" if it fails.


MohabMohie:
experiment failed.

PR #1462, opened 2024-01-20
update HTML helper

zanoon2020:
update search and filteration script

now you can filter by status and search for specific test name within filtered status only and vice versa


MohabMohie:
@zanoon2020 & @MahmoudElSharkawy shall we merge this?

MahmoudElSharkawy (approved):

PR #1471, opened 2024-01-25
update type method

zanoon2020:
adding try and catch to catch exception and add screenshot to fail message


MohabMohie (commented):
this is exactly the correct fix.

PR #1490, opened 2024-02-13
Allure report hotfix

MustafaAgamy:


src/main/java/com/shaft/listeners/TestNGListener.java
```diff
@@ -203,14 +203,14 @@ public void onExecutionFinish() {
...
         Thread.ofVirtual().start(GoogleTink::encrypt);
-        Thread.ofVirtual().start(AllureManager::openAllureReportAfterExecution);
         Thread.ofVirtual().start(ReportManagerHelper::logEngineClosure);
         try {
             summaryReportGeneration.join();
             allureArchiveGeneration.join();
         } catch (InterruptedException e) {
             throw new RuntimeException(e);
         }
+        AllureManager.openAllureReportAfterExecution();
```
MohabMohie: Why are we calling it outside the virtual thread?

src/main/java/com/shaft/tools/io/internal/AllureManager.java
```diff
@@ -153,11 +160,12 @@ private static void createAllureListenersMetaFiles() {
                 "io.qameta.allure.listener.StepLifecycleListener", "io.qameta.allure.listener.TestLifecycleListener").forEach(fileName -> FileActions.getInstance(true).writeToFile(Properties.paths.services(), fileName, "com.shaft.listeners.AllureListener"));
     }
 
-    private static void writeAllureReport() {
+    public static void writeAllureReport() {
```
MohabMohie: Why is it public now?

src/main/java/com/shaft/tools/io/internal/AllureManager.java
```diff
@@ -153,11 +160,12 @@ private static void createAllureListenersMetaFiles() {
...
         String commandToCreateAllureReport;
         allureBinaryPath = allureExtractionLocation + "allure-" + SHAFT.Properties.internal.allureVersion()
                 + "/bin/allure";
         String outputDirectory = System.getProperty("user.dir") + File.separator + "target" + File.separator + allureReportPath;
+        allureOutPutDirectory = outputDirectory;
```
MohabMohie: We don't need the local variable, and we don't need to assign value inside this method for the global variable, we can assign on the same declaration line since the path is static.

MohabMohie (changes requested):

MohabMohie (approved):

PR #1518, opened 2024-03-07
Added new separate class for handling attachments

MustafaAgamy:
- AttachementReporter Class Added containing the newly added logic for handling attachments
- Some Enhancements are made to the logic :
1- LinkedHashMap is now used instead of HashMap
2- Switch case now switches over the LinkedHashMap key instead of the AttachementType for better consistency
3- Get Attachment Case now returns  empty string ("") for better readability


MohabMohie (approved):
looks great, I validated an API and GUI sample and everything looks awesome.

PR #1601, opened 2024-04-22
Fix custom options

MustafaAgamy:
~~- Enable BiDi property default value is set to false as it was creating conflicts with session related properties like incognitoMode or setting profiles~~
- incognitoMode property is added for the following browsers:
- Google Chrome
- Microsoft Edge
- Mozilla Firefox

### DIFF
--- a/src/main/java/com/shaft/driver/internal/DriverFactory/OptionsManager.java
+++ b/src/main/java/com/shaft/driver/internal/DriverFactory/OptionsManager.java
@@ -76,6 +76,10 @@ public class OptionsManager {
                 if (SHAFT.Properties.web.headlessExecution()) {
                     ffOptions.addArguments("-headless");
                 }
+                //Incognito mode for Firefox
+                if(SHAFT.Properties.web.incognitoMode()) {
+                    ffOptions.addArguments("-private");
+                }
                 ffOptions.setLogLevel(FirefoxDriverLogLevel.WARN);
                 ffOptions.setPageLoadStrategy(PageLoadStrategy.NORMAL);
                 ffOptions.setPageLoadTimeout(Duration.ofSeconds(SHAFT.Properties.timeouts.pageLoadTimeout()));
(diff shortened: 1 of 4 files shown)


src/main/java/com/shaft/properties/internal/Platform.java
```diff
@@ -44,7 +44,7 @@ private static void setProperty(String key, String value) {
     boolean jvmProxySettings();
 
     @Key("enableBiDi")
-    @DefaultValue("true")
+    @DefaultValue("false")
```
MohabMohie: this change is not acceptable. the default must remain "true" for BiDi.

src/test/java/testPackage/properties/WebTests.java
```diff
@@ -33,6 +34,7 @@ public void beforeClass() {
         baseURL = SHAFT.Properties.web.baseURL();
         browserWindowWidth = 1920;
         browserWindowHeight = 1080;
+        incognitoMode = SHAFT.Properties.web.incognitoMode();
```
MohabMohie, line 36: this test will break the rest of the pipeline, any changes in configuration/properties must be done before method, and must be reversed after method.

src/main/java/com/shaft/driver/internal/DriverFactory/OptionsManager.java
```diff
@@ -293,7 +297,14 @@ private ChromiumOptions<?> setupChromiumOptions(ChromiumOptions<?> options, Muta
             options.addArguments("--window-position=0,0", "--window-size=" + DriverFactoryHelper.getTARGET_WINDOW_SIZE().getWidth() + "," + DriverFactoryHelper.getTARGET_WINDOW_SIZE().getHeight());
         }
         if (!SHAFT.Properties.flags.autoCloseDriverInstance()) options.setExperimentalOption("detach", true);
-
+        //Incognito mode for Chrome and Edge
+        if(SHAFT.Properties.web.incognitoMode()) {
+            if(options.getBrowserName().equalsIgnoreCase("chrome")) {
```
MohabMohie: kindly use the browser name enums instead of using string values for browser names.

MohabMohie: this change should be implemented inside the "chromiumOptions" method to avoid increasing complexity. kindly move it there.

MohabMohie (changes requested):

src/test/java/testPackage/properties/WebTests.java
MustafaAgamy: I'm using the default value which is still false , can you please elaborate more because no values were set

src/main/java/com/shaft/driver/internal/DriverFactory/OptionsManager.java
MustafaAgamy: The code is added in the setupChromiumOptions method for both Chrome and Edge browsers
As for firefox it's not a chromium-based browser so the logic was added directly in the setDriverOptions method

As for the Enums approach it's fixed now, can you please check? 

Thanks.

MustafaAgamy:
@MohabMohie FYI

MahmoudElSharkawy (approved):

MohabMohie (approved):

PR #1613, opened 2024-05-14
ProgressBar_Enhancement

ahmednasr95:
* Fixed the repetitive logging during failed assertion
* Added a new logger to the log4j2.properties to be able to print the progress bar on the console
* Added the progress bar to indicate the time elapsed since start of assertion

### DIFF
--- a/src/main/java/com/shaft/tools/internal/support/JavaHelper.java
+++ b/src/main/java/com/shaft/tools/internal/support/JavaHelper.java
@@ -97,7 +97,7 @@ public class JavaHelper {
      */
     public static int compareTwoObjects(Object expectedValue, Object actualValue, Object comparisonType,
                                         Boolean validationType) {
-        ReportManager.logDiscrete("Expected \"" + expectedValue + "\", and actual \"" + actualValue + "\"");
+        //ReportManager.logDiscrete("Expected \"" + expectedValue + "\", and actual \"" + actualValue + "\"");
         if ("null".equals(expectedValue)) {
             expectedValue = null;
         }
--- a/src/main/java/com/shaft/tools/io/internal/ReportManagerHelper.java
+++ b/src/main/java/com/shaft/tools/io/internal/ReportManagerHelper.java
@@ -51,6 +51,7 @@ public class ReportManagerHelper {
     private static List<List<String>> listOfNewIssuesForFailedTests = new ArrayList<>();
     private static String featureName = "";
     private static Logger logger;
+    private static Logger progressLogger;
 
     private ReportManagerHelper() {
         throw new IllegalStateException("Utility class");
@@ -168,6 +169,7 @@ public class ReportManagerHelper {
         // initialize
         Configurator.initialize(null, PropertyFileManager.getCUSTOM_PROPERTIES_FOLDER_PATH() + "/log4j2.properties");
         logger = LogManager.getLogger(ReportManager.class.getName());
+        progressLogger = LogManager.getLogger("progressConsole");
     }
 
     public static void logEngineVersion() {
@@ -788,5 +790,14 @@ public class ReportManagerHelper {
         }
         return duration;
     }
-
+    public static void printProgressBarAfterUnitsElapsed(int maximumValue, int units){
+        // to keep the bar short in case of long waiting times
+        if(maximumValue > 60){
+            maximumValue %= 60;
+        }
+        progressLogger.info("\rWaiting.. |"+"\u001B[107m \u001B[0m".repeat(units)+" ".repeat(maximumValue-units)+"| "+units+" seconds");
+    }
+    public static void printNewlineAfterProgressBar(){
+        logger.info("\r\n");
+    }
 }
--- a/src/main/java/com/shaft/validation/internal/ValidationsHelper2.java
+++ b/src/main/java/com/shaft/validation/internal/ValidationsHelper2.java
@@ -1,5 +1,6 @@
 package com.shaft.validation.internal;
 
+import com.google.common.util.concurrent.AtomicDouble;
 import com.shaft.driver.SHAFT;
 import com.shaft.driver.internal.DriverFactory.SynchronizationManager;
 import com.shaft.gui.browser.BrowserActions;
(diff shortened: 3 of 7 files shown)


MhmdElGazzar:
@MohabMohie  @MahmoudElSharkawy 
could you please help us with this PR?

src/main/java/com/shaft/tools/internal/support/JavaHelper.java
```diff
@@ -97,7 +97,7 @@ public static String convertBase64(String text) {
      */
     public static int compareTwoObjects(Object expectedValue, Object actualValue, Object comparisonType,
                                         Boolean validationType) {
-        ReportManager.logDiscrete("Expected \"" + expectedValue + "\", and actual \"" + actualValue + "\"");
+        //ReportManager.logDiscrete("Expected \"" + expectedValue + "\", and actual \"" + actualValue + "\"");
```
MohabMohie: instead of commenting this line, kindly log it using DEBUG level.

src/main/java/com/shaft/validation/internal/ValidationsHelper2.java
```diff
@@ -75,6 +86,11 @@ protected void validateBrowserAttribute(WebDriver driver, String attribute,
 
         try {
             new SynchronizationManager(driver).fluentWait(false).until(f -> {
+                // Increase the bar after 6 (100) millisecond runs to take into account code execution time
+                if(millisCount.incrementAndGet() == 6){
```
MohabMohie: this is the root cause for the race condition issue (sometimes the counter ends at 80 or 90% not reaching 100%)
instead of creating the logic inside the fluent wait block, it should be created outside it and run in a separate virtual thread.

it is recommended to implement this virtual thread outside this class to avoid code duplication, specifically in the ValidationsExecutor class (which calls the validation helper) before the switch case statement that determines how the assertion will be performed.

src/main/java/com/shaft/tools/io/internal/ReportManagerHelper.java
```diff
@@ -51,6 +51,7 @@ public class ReportManagerHelper {
     private static List<List<String>> listOfNewIssuesForFailedTests = new ArrayList<>();
     private static String featureName = "";
     private static Logger logger;
+    private static Logger progressLogger;
```
MohabMohie: Kindly use the same logger object rather than creating a different one.
creating a different logger config file will add a lot of complexity as we need to bundle it to SHAFT users else it will not work as expected.

src/test/java/poms/GoogleSearch.java
```diff
@@ -60,7 +60,7 @@ public void unSelectQuery() {
     public void assertPageIsOpen() {
             Validations.assertThat()
                     .element(driver, googleLogo_image)
-                    .exists()
+                    .doesNotExist()
```
MohabMohie: kindly revert the changes to these classes as they are used in other tests.

src/test/java/poms/GoogleSearchResults.java
```diff
@@ -18,7 +18,7 @@ public void assertResultsStatsExistsAndIsNotEmpty() {
             Validations.assertThat()
                     .element(driver, resultsStats_label)
                     .text()
-                    .doesNotEqual("")
+                    .isEqualTo("")
```
MohabMohie: kindly revert the changes to these classes as they are used in other tests.

src/test/resources/TestSuites/validationTests.xml
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
```
MohabMohie: kindly delete this file as it's not part of the project itself.

MohabMohie (changes requested):
Thank you for an excellent implementation, and kindly find my comments inline.

regarding the virtual threads, kindly tag @MustafaAgamy via Slack and he'll be happy to help.

src/main/java/com/shaft/tools/io/internal/ReportManagerHelper.java
ahmednasr95: Regarding this part, I had to make a new logger because the existing one has a newline character at the end of its pattern which will print the progress bar in a newline after each update. Thus, I had to create a new logger with the same pattern as the existing one but only has the newline character removed. 
If there is a way, which I guess there is not, to prevent the existing logger from printing a newline during the progress bar printing process only, then I will be more than glad to implement it.
Moreover, please note that the new logger I created is made specifically for the progress bar only and is not intended to be used by any other logging processes.

src/main/java/com/shaft/tools/internal/support/JavaHelper.java
ahmednasr95: I tried to logging using the `DEBUG` level but to no avail. As shown in the screenshot below, The message is still logged to the console as an `INFO`.

Screenshot

src/main/java/com/shaft/validation/internal/ValidationsHelper2.java
ahmednasr95: * Thanks for the virtual thread suggestion, it is working perfectly 😄. 
* Regarding calling the thread before the switch case in the `ValidationsExecutor` class, I tried this approach, but the problem is that the progress bar is called multiple times during which it should not be called as shown in the screenshot below. Thus causes the console log to appear distorted. 
* The progress bar is made specifically to be called only during a fluent wait so that it shows the user that the script is actually running and waiting for the assertion to pass thus preventing the "false stuck script" issue.

Screenshot

src/test/resources/TestSuites/validationTests.xml
ahmednasr95: I apologize for this. I was using it to check that the functionalities are intact. 😅

src/main/java/com/shaft/validation/internal/ValidationsHelper2.java
ahmednasr95: Thanks to Mostafa Agamy's notice, I tested it again and it worked from my side so the thread call is now before the switch statement and the shutdown call is after the statement.

MohabMohie (approved):

PR #1698, opened 2024-08-13
Update AllureManager.java

zanoon2020:


src/main/java/com/shaft/tools/io/internal/AllureManager.java
```diff
@@ -53,9 +53,8 @@ private static void copyAndOpenAllure() {
     }
 
     private static String renameAllureReport() {
```
zanoon2020, line 55: @MohabMohie i need your opinion here , i did change to make any allure report named with generation date whether accumulate flag is enabled or not .

Issue #1697, opened 2024-08-13
[Bug]: (API) SOAP Request Fails Due to Improper Encoding of XML Content Type


KyrillosNageh:
**Describe the bug**

When attempting to send a SOAP request with an XML body, the request fails due to an improper encoding of the content type text/xml. The system throws a java.lang.IllegalArgumentException indicating that it doesn't know how to encode the SOAP envelope as a byte stream

**Environment**

 1. SHAFT_Engine version : Latest version (8.3.20240803)

**To Reproduce**

1. Initialize a RequestBuilder object to make a SOAP request.
2. Set the Content-Type header to text/xml.
3. Attempt to send the request with the provided XML body.
4. Observe the failure in the logs.

**Example code:**

  api = new SHAFT.API("http://www.dneonline.com/calculator.asmx");
String requestXml="<soapenv:Envelope xmlns:soapenv=\"http://schemas.xmlsoap.org/soap/envelope/\" xmlns:tem=\"http://tempuri.org/\">\n" +
        "   <soapenv:Header/>\n" +
        "   <soapenv:Body>\n" +
        "      <tem:Multiply>\n" +
        "         <tem:intA>10</tem:intA>\n" +
        "         <tem:intB>2</tem:intB>\n" +
        "      </tem:Multiply>\n" +
        "   </soapenv:Body>\n" +
        "</soapenv:Envelope>";
        
api.addHeader("Content-Type", "text/xml");
api.addHeader("SOAPAction", "http://tempuri.org/Multiply");
Response response = api.post("")
    .setRequestBody(requestXml)
    .performRequest();

**Expected behavior**

The SOAP request should be properly encoded and sent to the server, with the XML body being handled as text/xml content.

**Actual  behavior**

The system throws a java.lang.IllegalArgumentException, indicating that it does not know how to encode the SOAP envelope as a byte stream, causing the request to fail.

**Link to SHAFT_Engine Console logs**

https://gist.github.com/KyrillosNageh/086244efc9d26761bfc34e03c58a78e7

MohabMohie:
can you kindly share a native rest assured code snippet that works?

KyrillosNageh:
String requestXml="<soapenv:Envelope xmlns:soapenv=\"http://schemas.xmlsoap.org/soap/envelope/\" xmlns:tem=\"http://tempuri.org/\">\n" +
                "   <soapenv:Header/>\n" +
                "   <soapenv:Body>\n" +
[...]

PR #1703, opened 2024-08-17
Fix: Ensure Content-Type is set based on header value if not explicit…

KyrillosNageh:
Fixes [#1697](https://github.com/ShaftHQ/SHAFT_ENGINE/issues/1697)

This PR ensures that the `Content-Type` is set based on the header value if it is not explicitly set by the user.
- Modified the `prepareRequestSpecs` method to check if `contentType` is still `ContentType.ANY`.
- If `ContentType` is not explicitly set by the user, it now retrieves the `Content-Type` from `sessionHeaders` and applies it directly to the `RequestSpecBuilder`.
- This change ensures that the request uses the exact `Content-Type` provided in the headers, preventing encoding and serialization issues.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,7 @@
 
     <dependencyManagement>
         <dependencies>
+
             <!-- JUNIT5 BOM -->
             <!-- https://mvnrepository.com/artifact/org.junit/junit-bom -->
             <dependency>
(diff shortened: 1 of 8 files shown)


KyrillosNageh:
.

PR #1704, opened 2024-08-17
Fix: Ensure Content-Type is set based on header value if not explicitly set

KyrillosNageh:
Fixes [#1697](https://github.com/ShaftHQ/SHAFT_ENGINE/issues/1697)

This PR ensures that the `Content-Type` is set based on the header value if it is not explicitly set by the user.
- Modified the `prepareRequestSpecs` method to check if `contentType` is still `ContentType.ANY`.
- If `ContentType` is not explicitly set by the user, it now retrieves the `Content-Type` from `sessionHeaders` and applies it directly to the `RequestSpecBuilder`.
- This change ensures that the request uses the exact `Content-Type` provided in the headers, preventing encoding and serialization issues.


MohabMohie (approved):
Thank you for your valuable contribution!
Note that in the future it would be great if you can add a test class that can cover this case to ensure our code coverage doesn't decrease.

PR #1800, opened 2024-11-08
fix allure history and trends generation

MohabMohie:
-allure.accumulateHistory=true
will now behave as expected, saving execution history and showing relevant trends to help identify flaky tests.

-allure.accumulateReports=true
will now be more relevant, as the report will hold details of the current test run only; previous run details will no longer show in the retries tab.

### DIFF
--- a/src/main/java/com/shaft/tools/io/internal/AllureManager.java
+++ b/src/main/java/com/shaft/tools/io/internal/AllureManager.java
@@ -138,12 +138,14 @@ public class AllureManager {
     }
 
     private static void cleanAllureResultsDirectory() {
-        // clean allure-results directory before execution
-        var allureResultsPath = allureResultsFolderPath.substring(0, allureResultsFolderPath.length() - 1);
-        try {
-            internalFileSession.deleteFolder(allureResultsPath);
-        } catch (Exception t) {
-            ReportManager.log("Failed to delete '" + allureResultsPath + "' as it is currently open. Kindly restart your device to unlock the directory.");
+        if (SHAFT.Properties.allure.cleanResultsDirectory()) {
+            // clean allure-results directory before execution
+            var allureResultsPath = allureResultsFolderPath.substring(0, allureResultsFolderPath.length() - 1);
+            try {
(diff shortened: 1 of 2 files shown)


src/main/java/com/shaft/tools/io/internal/AllureManager.java
```diff
@@ -177,22 +176,55 @@ private static void overrideAllurePluginConfiguration() {
...
+        if (SHAFT.Properties.allure.accumulateHistory()) {
+            // move existing history to the correct folder
+            if (internalFileSession.doesFileExist(pathToLastHistoryDirectory)) {//if not the first test run and history already exists
+                internalFileSession.copyFolder(pathToLastHistoryDirectory, allureResultsFolderPath + File.separator + "history");
+            } else {
+                internalFileSession.createFolder(pathToLastHistoryDirectory);
+            }
+
+            internalTerminalSession.performTerminalCommand(getCommandToCreateAllureReport(customReportName, true));
+            internalTerminalSession.performTerminalCommand(getCommandToCreateAllureReport(customReportName, false));
```
AhmdZanoon, line 197: Dear @MohabMohie , I think this line will cancel the previous one , so the report will always be without history .

AhmdZanoon (changes requested):

src/main/java/com/shaft/tools/io/internal/AllureManager.java
MohabMohie: History data is now being moved to a temporary folder in the target directory.

This line ensures that there's no out of sync duplication between the history and retries tab.

By design, retries tab should only be for the current test run. And this allure-results folder should be cleared every run except for the history data, which is what we now map to.

MohabMohie: After discussion via slack your points made lots of sense and we agreed to use the following setup

allure.accumulateHistory=true
allure.accumulateReports=true
allure.cleanResultsDirectory=true

AhmdZanoon (approved):
nice and clean implementation as usual.
thanks @MohabMohie for your time and effort 😊

PR #1806, opened 2024-11-15
Fixing Access SQL Custom connection

Ayanasrallah:


MohabMohie (approved):
Thank you for your valuable contribution Aya!
With these changes custom database connection strings will work as they should have, without impacting existing functionality.

PR #1860, opened 2025-01-28
Allure results dynamic path

MustafaAgamy:
Added Logic to make Allure-Results path Dynamic via Properties file 

You can find more info in the following link
https://allurereport.org/docs/testng/#specifying-allure-results-location

What's done in that PR:
- Create `"AllureResults"` Interface that contains the property value that will be in the `"allure.properties"` file
- Create `"allure.properties"` file with the `"allure.results.directory"` property in the same directory as the `default properties`
- Create the method that will copy `"allure.propeties"` in the `"initializeAllureProperties"` method at `"PropertiesHelper"` class following the same approach of  `"initializeDefaultProperties`" method
- Fix the paths of allureresults in `"AllureManager"` to instead of `"SHAFT.Properties.paths.allureResults()"` to `"SHAFT.Properties.allureResults.allureResultsDirectory()"`
- Added the logic to `"ReportHelper"` class in order to attach the `"allure.properties"` at `"src/test/resources"` directory to be attached to the allure report
- Created `"AllureResultsTests"` to cover the newly added Property

**The logic will do the following, when running the project for the first time and allure.properties doesn't exist it will create it under test/resources after that when the user can change the allure-results path**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1720,4 +1720,3 @@ OpenLHReport.js
 /src/main/resources/docker-compose/data/storage/prj-users/thumbnail-default
 /src/main/resources/docker-compose/data/storage/prj-users/thumbnail-superadmin
 /src/test/resources/testDataFiles/allure-results/environment.xml
-/src/test/resources/allure.properties
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,7 @@ debug.log
 /generateJdk.bat
 /generateJdk.sh
 /src/main/resources/docker-compose/android-emulator/*
-/src/main/resources/properties/*
+#/src/main/resources/properties/*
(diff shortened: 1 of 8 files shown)


MohabMohie:
Can we discuss this more via slack?
I don't get the ROI of separating allure properties in a separate file if we still only support the path variable (which we already do under the path properties).

Is this an initial PR, and you have a plan to support all allure properties? If so, should we also move the report logo property?

MohabMohie:
I'll close this PR for now as per my last comment. We already have the allure path variable in our path properties file.
@AhmdZanoon to advise.

MustafaAgamy:
@MohabMohie let's discuss more on slack

PR #1887, opened 2025-02-14
Updating CSVFileManager

mohamedtawfek05354:
Adding getFirstColumn,MaxCell , MinCell and Column count and their comments


MohabMohie (approved):
thank you for the excellent contribution!

PR #1956, opened 2025-03-26
Enable WebDriverListener Attachment via getDriver(WebDriverListener) for Non-Appium Drivers

Kinasr:
This pull request introduces a new feature to the `getDriver(WebDriverListener)` method, allowing users to attach a `WebDriverListener` to Selenium WebDriver instances. This enhances the driver's functionality by enabling custom event handling and logging.

**Changes Made:**

- Implemented the `getDriver(WebDriverListener)` method to wrap non-Appium drivers (excluding `AndroidDriver` and `IOSDriver`) with `EventFiringDecorator`.
- Added logic to bypass `EventFiringDecorator` for Appium drivers due to compatibility issues.
- Included a unit test to verify the correct application of the listener for non-Appium drivers.

**Rationale:**

- Provides users with the ability to attach custom listeners for advanced WebDriver interactions.
- Improves flexibility and enables better debugging and logging.
- Maintains compatibility with Appium drivers by avoiding `EventFiringDecorator`.
- Adheres to the recommendation of using SHAFT's `WebDriverListener` for optimal usage.

**Testing:**

- Added a unit test to ensure that the listener is correctly applied to non-Appium drivers.

This feature adds significant value by allowing users to extend the WebDriver's behavior through custom listeners, while ensuring compatibility with various driver types.

### DIFF
--- a/src/main/java/com/shaft/driver/SHAFT.java
+++ b/src/main/java/com/shaft/driver/SHAFT.java
@@ -19,8 +19,11 @@ import com.shaft.validation.internal.RestValidationsBuilder;
 import io.appium.java_client.android.AndroidDriver;
 import io.appium.java_client.ios.IOSDriver;
 import io.restassured.response.Response;
+import javassist.expr.Instanceof;
 import org.openqa.selenium.MutableCapabilities;
+import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.support.events.EventFiringDecorator;
+import org.testng.internal.thread.Async;
 
 import java.io.InputStream;
 import java.sql.ResultSet;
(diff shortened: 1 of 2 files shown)


MohabMohie:
@Kinasr I don't understand the business proposition dear.

So you're returning the native driver and you're attaching a custom listener to it?

That's outside SHAFT's scope of work. Ideally any enhancements you need (logging, reporting, ...etc) should be bundled within the engine itself. But if you're returning a native driver then you're outside SHAFT's scope of work and you can attach a listener to it on your own.

This is highly unrecommended dear. A project that's using both SHAFT and Native drivers with conflicting listeners will lead to unexpected events.

Kindly explain the business proposition to me, so that I can understand it better.

Kinasr:
Hi @MohabMohie, regarding your inquiry about the business proposition for PR #1956:

The primary objective is to enable the smooth integration of custom functionalities before and after element and browser actions. 
My initial approach involved using WebDriverListener, but I encountered limitations in attaching it directly to the SHAFT driver.

Consequently, I resorted to a workaround: extracting the native WebDriver instance from SHAFT and attaching my custom listener directly. 
However, this approach presents a significant issue: SHAFT also attaches its own internal listeners to the native driver. 
This potential overlap between my custom listener and SHAFT's internal listeners poses a considerable risk of instability.

Therefore, the proposed feature in PR #1956 seeks to offer a native, integrated solution for attaching custom WebDriverListeners within the SHAFT framework, thereby preventing the multiple assignment of listeners and the associated conflicts.

MohabMohie:
you can disable the attachment of SHAFT's native listener by closing this flag: respectBuiltInWaitsInNativeMode=false

this will return a clean / native webdriver for you to play around with. but again this is completely outside SHAFT's scope of work. And therefore I have no intention of providing SHAFT users with more capabilities to break the setup of their projects.

If you're doing anything with your listeners that will add value to logging/reporting then the correct action would be to contribute your code to the listener that's inside SHAFT. I do not recommend overriding it.

PR #2357, opened 2026-03-29
fix(mobile): Appium 2 session URL, APK badging caps, and skip Maven CD on forks

Mochxd:
## Summary
- **Appium 2:** After a failed W3C `POST /session`, skip the legacy `/wd/hub` retry when the target platform is Android or iOS. On Appium 2 that path returns 404 and hides the real `SessionNotCreatedException`. Users on Appium 1 can still set `/wd/hub` in `executionAddress` if needed.
- **APK-only config:** If `mobile_app` points to an APK and `mobile_appPackage` / `mobile_appActivity` are missing, infer them from `aapt` / `aapt2 dump badging` when `ANDROID_HOME` or `ANDROID_SDK_ROOT` is set. Opt out: `-Dshaft.skipApkPackageActivityInference=true`.
- **Launch:** If `appium:app` is set and `appium:appWaitActivity` is not, set `appium:appWaitActivity` to `*` to avoid hanging on splash vs main activity.
- **CI:** Do not run `mavenCentral_cd` on **forked** repositories (no GPG/OSSRH secrets → signing fails).

## Implementation
- New helper: `AndroidApkBadgingReader` (+ unit tests for badging parsing).
- Wired from `PropertyFileManager.getAppiumDesiredCapabilities()` after resolving the APK path.
- `OptionsManager` sets default `appium:appWaitActivity` for native Appium caps where applicable.
- `DriverFactoryHelper.attemptRemoteServerConnection`: conditional legacy hub retry.

## Manual flow:
- Appium 2.x, executionAddress without /wd/hub, deliberate caps failure → server log should not show POST /wd/hub/session after the failed /session. With only mobile_app + ANDROID_HOME, check logs for inferred package/activity.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
     <properties>
         <surefireArgLine></surefireArgLine>
         <!-- Dependency Versions -->
-        <log4j2.version>2.25.3</log4j2.version>
+        <log4j2.version>2.25.4</log4j2.version>
         <!-- Build Commands-->
         <installToLocalRepository>mvn clean install "-DskipTests" "-Dgpg.skip"</installToLocalRepository>
         <publishJavaDocs>mvn resources:resources javadoc:javadoc scm-publish:publish-scm</publishJavaDocs>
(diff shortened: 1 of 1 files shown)


Mochxd:
Hey @MohabMohie 👋 
The pipeline’s throwing a lot of failures I don’t fully understand yet, no rush when you get a chance, could i have ur support

PR #2566, opened 2026-05-07
fix(reporting): close attachment streams and ReportPortal temp cleanup

Mochxd:
### Summary
- Close InputStream in ReportManagerHelper#createAttachment after copying bytes (reduces FD leaks under parallel runs).
- AttachmentReporter: delete rp-test* temp files after ReportPortal.emitLog; on failure, log and continue instead of throwing.
- Test-data and media attach paths use in-memory bytes where the report path previously opened an extra FileInputStream.

### Tests
- ReportManagerHelperAttachmentIoUnitTest --> stream closure + no extra rp-test* files when ReportPortal is mocked.
- Verify it through this-->  mvn -Dtest=ReportManagerHelperAttachmentIoUnitTest test

### DIFF
--- a/src/main/java/com/shaft/gui/internal/image/AnimatedGifManager.java
+++ b/src/main/java/com/shaft/gui/internal/image/AnimatedGifManager.java
@@ -84,7 +84,8 @@ public class AnimatedGifManager {
         // stop and attach
         if (SHAFT.Properties.visuals.createAnimatedGif() && !gifRelativePathWithFileName.get().isEmpty()) {
             try {
-                ReportManagerHelper.attach("Animated Gif", String.valueOf(System.currentTimeMillis()), new FileInputStream(gifRelativePathWithFileName.get()));
+                byte[] gifBytes = java.nio.file.Files.readAllBytes(java.nio.file.Paths.get(gifRelativePathWithFileName.get()));
+                ReportManagerHelper.attach("Animated Gif", String.valueOf(System.currentTimeMillis()), new ByteArrayInputStream(gifBytes));
                 if (gifWriter.get() != null) {
                     gifManager.get().close();
                 }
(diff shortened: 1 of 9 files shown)


Mochxd:
as usual, I have a problem from the pipeline so appreciate ur spport @MohabMohie

MohabMohie:
@Mochxd don't worry about the pipeline failures dear, I'll check and handle them until the pipelines are stabelized. There are around 60 tickets for coverage and testing and it's planned to be fixed within this month.
For now, let's just focus on your scope, if you can attach screenshots showing that you've fixed the issue locally, and you can address copilot review comments and human review comments from myself or any other maintainer, then we should be good to merge.
Once more, thank you for your contribution!

Mochxd:
@MohabMohie 
Hi Mohab, thanks a lot that really helps knowing you’ll tackle the pipeline side while coverage/testing work is scheduled
I stayed within the reporting/attachment scope. Copilot’s points are addressed (including the follow-up commit), and locally I ran ReportManagerHelperAttachmentIoUnitTest with everything passing I’ll attach a screenshot of that summary on the PR so you can see the local check.
[2026-05-08_14-56-04-538_AllureReport.html](https://github.com/user-attachments/files/27519261/2026-05-08_14-56-04-538_AllureReport.html)
 if you or another maintainer wants changes on review, tell me, thanks again for the guidance as usual

Mochxd:
@MohabMohie I’ve pushed another small update for the latest Copilot feedback on this PR:
- GIF + visual compare attachments now use streaming (Files.newInputStream) instead of loading full files into byte[] first, so we’re not doubling memory before createAttachment.
- The ReportPortal temp-dir test runs with @Test(singleThreaded = true) so we’re not fighting other tests over java.io.tmpdir.
I ran ReportManagerHelperAttachmentIoUnitTest locally and it’s green

PR #2646, opened 2026-05-08
Monorepo Phase 1: multi-module Maven structure + ValidationsBuilder boundary fix

MustafaAgamy:
## Summary

Converts SHAFT_ENGINE from a monolithic artifact into a proper 6-module Maven monorepo, eliminates a class-shadowing anti-pattern in the validation layer, and consolidates all dependency versions into the BOM.

### Module structure after migration

| Module | Contents |
|--------|----------|
| `shaft-bom` | Bill of Materials — all version pins |
| `shaft-core` | Properties, tools, listeners, validation core, enums, CLI |
| `shaft-api` | REST API (RestActions, RequestBuilder, validations) |
| `shaft-db` | DatabaseActions |
| `shaft-web` | GUI/driver/cucumber + all test suites + all resources |
| `shaft-engine` | Thin aggregator (depends on shaft-web) — backwards-compatible umbrella |

### ValidationsBuilder boundary fix

`shaft-web` carried a full duplicate of `com.shaft.validation.internal.ValidationsBuilder` at the same FQCN as `shaft-core`'s — a class-shadowing anti-pattern that would silently diverge over time.

**Fix:**
- Deleted shaft-web's duplicate `ValidationsBuilder.java`
- Created `WebValidations` typed entry point returning `WebValidationsBuilder` from `assertThat()` / `verifyThat()`
- Updated `WizardHelpers.java` 4 delegation lines to route through `WebValidations`
- Updated 23 call sites across test/main source: `element()`/`browser()` chains → `WebValidations`, `response()` chains → `ApiValidations`

### BOM version consolidation

All module `pom.xml` files had retained explicit `<version>` tags for BOM-managed dependencies from the original monorepo extraction — an incorrect pattern that caused `RequireUpperBoundDeps` enforcer failures. Fixed by:

- `shaft-bom` is now the single source of truth for all dependency versions
- Removed 44+ explicit `<version>` tags from `shaft-web/pom.xml`
- Removed duplicate log4j version property and explicit version tags from `shaft-core/pom.xml`
- Bumped `browserstack-java-sdk` to `1.58.2` in the BOM (was missing the bump applied to shaft-web)
- Added missing `log4j-api` entry to BOM `<dependencyManagement>`
- Module-local `<version>` retained only where legitimately required: platform native binaries not in BOM, and plugin `<dependencies>` which are not governed by BOM managed versions

### Bug fixes discovered during test run

[...]

PR #2688, opened 2026-05-10
fix(#2687): fix parallel Cucumber IndexOutOfBoundsException and webdriver.http.factory regression

### LINKED ISSUE #2687: [Bug]: IndexOutOfBoundsException: Range [0, 1) out of bounds for length 0 on Chrome driver init in parallel execution — LTS regression
Description:

When running Cucumber tests in parallel using AbstractTestNGCucumberTests with @DataProvider(parallel = true), SHAFT Engine throws an IndexOutOfBoundsException during driver initialization. All tests fail instantly (~5–20ms) and no browser is ever opened. Running the same tests sequentially works fine.

This is a regression introduced in the LTS version — the issue does not occur on older non-LTS versions.

SHAFT Version: 10.2.20260506 (LTS)
Java: 25
Parallel config: setParallel=CLASSES, setThreadCount=3, setDataProviderThreadCount=3

Steps to Reproduce:

Use AbstractTestNGCucumberTests with @DataProvider(parallel = true)
Set setParallel=CLASSES and setThreadCount > 1
Run any Cucumber tag with multiple scenarios
Error:

java.lang.AssertionError: Failed to start Chrome driver: Range [0, 1) out of bounds for length 0
    at org.testng.Assert.fail(Assert.java:111)
    at ui.connector.Hooks.startDriver(Hooks.java:36)
Expected: All threads initialize Chrome drivers successfully in parallel.

Actual: All but one thread fail immediately — one thread wins the initialization race; the rest receive an empty browser path list and subList(0, 1) throws IndexOutOfBoundsException.

Workaround:
Adding a @BeforeSuite to pre-warm SHAFT's static initialization before parallel threads start resolves the issue:

@BeforeSuite(alwaysRun = true)
public void preWarmDriverFactory() {
    Properties.platform.executionAddress();
}
Suspected Root Cause:
DriverFactoryHelper performs lazy, unsynchronized static initialization of the Chrome binary path. Under concurrent access, threads 2–N call getBrowserPath() before thread 1's resolution completes, receive an empty list, and subList(0, 1) throws. This was not an issue in older SHAFT versions, suggesting the initialization strategy changed in LTS.

Image

MustafaAgamy:
## Summary

Fixes `IndexOutOfBoundsException: Range [0, 1) out of bounds for length 0` in parallel Cucumber+TestNG runs (`AbstractTestNGCucumberTests`, `setParallel=CLASSES`, `setThreadCount=3`).

### Root cause

PR #2488 built `ThreadLocalPropertiesManager` with the right intent — move SHAFT-internal property writes off `System.setProperty()` to prevent cross-thread contamination. The flaw was applying `setGlobalProperty()` to a different category of properties: ones consumed by **external libraries** (Selenium, Log4j, Allure, `java.net.*`) that call `System.getProperty()` directly and have no knowledge of SHAFT's internal map.

There are two completely separate runtime property stores:
- **SHAFT `globalOverrides`** — only SHAFT's OWNER proxies see this (3-tier fallback chain)
- **JVM `System.getProperties()`** — every Java library reads this; SHAFT also reads it (as fallback tier 3)

Writing to one does **not** write to the other.

### Five root causes fixed

**1. `webdriver.http.factory` invisible to Selenium**
`setKeySystemProperties()` used `setGlobalProperty()` for all five key properties. Selenium's `HttpClientFactory` calls `System.getProperty("webdriver.http.factory")` to choose between JDK and Netty HTTP clients. When null, Selenium fell back to Netty — which is not thread-safe — causing the `subList(0,1)` race in TestNG 7.12.0 under concurrent ChromeDriver initialization. Now uses `System.setProperty()`.

**2. `alter()` Cucumber misdetection**
`identifyRunType()` checked the call stack for `io.cucumber.core.runner.Runner`, which is never present when `alter()` fires. Cucumber runs were always misidentified as TestNG, calling `configureTestNGProperties()` instead of `configureCucumberProperties()`. Fixed via `isCucumberRun()` which inspects `XmlSuite` class names using `isAssignableFrom(AbstractTestNGCucumberTests.class)`.

**3. Null `Properties.testNG` in `alter()`**
`Properties.testNG` is `null` at `alter()` time because `loadProperties()` runs later in `onExecutionStart()`. Silent NPE left the user's parallel settings unchanged, feeding the TestNG `TestRunner.privateRun()` race. `configureTestNGProperties()` now lazily initializes via OWNER `ConfigFactory.create()` when `null`.

**4. `configureJVMProxy()` — 6 proxy keys invisible to `java.net.*`**
`TestNGListenerHelper.configureJVMProxy()` used `setGlobalProperty()` for `http.proxyHost`, `http.proxyPort`, `https.proxyHost`, `https.proxyPort`, `ftp.proxyHost`, `ftp.proxyPort`. Java's networking stack reads these exclusively from `System.getProperty()`. Changed to `System.setProperty()`.

**5. `SE_DRIVER_MIRROR_URL` — Edge-arm-only with post-construction clear**
`DriverFactoryHelper` set this in the Edge switch arm via `setGlobalProperty()`. Selenium Manager reads it via `System.getProperty()` to choose the Edge CDN. Two scope corrections:
- It must stay in the Edge switch arm — *not* moved to `setKeySystemProperties()`. Selenium Manager's `runCommand()` passes every `SE_*` JVM system property to the `selenium-manager` binary as an environment variable, so a global value would redirect Chrome and Firefox driver downloads to the Edge CDN (which doesn't serve those binaries).
- `System.setProperty()` is JVM-global, so the Edge arm now also calls `System.clearProperty("SE_DRIVER_MIRROR_URL")` immediately after `new EdgeDriver()` returns. Selenium Manager runs synchronously inside the constructor, so clearing afterwards is safe and prevents the property from leaking to subsequent non-Edge driver inits in the same JVM.

### Systemic guard

Added two `public static final Set<String>` catalogue constants in `PropertiesHelper`:
- `JVM_SYSTEM_PROPERTIES_ALWAYS` — 5 keys always set at engine startup (`webdriver.http.factory`, `rp.properties.path`, `log4j.configurationFile`, `allure.testng.hide.configuration.failures`, `allure.testng.hide.disabled.tests`)
- `JVM_SYSTEM_PROPERTIES_PROXY` — 6 proxy keys set when proxy is configured

`SE_DRIVER_MIRROR_URL` is intentionally **not** in either catalogue — it's browser-specific and short-lived, not a startup constant.

Two unit tests assert every catalogued key is present in `System.getProperty()` after the relevant setup call. CI fails immediately if any key is accidentally routed back to `setGlobalProperty()`. A separate `DriverFactoryHelperUnitTest` covers the Edge clear-after-construction guarantee.

## Test plan

- [x] `PropertiesHelperCoverageUnitTest#initializeAndSetKeySystemPropertiesShouldLoadPropertyStateAndSystemKeys` — asserts `webdriver.http.factory == jdk-http-client`
- [x] `PropertiesHelperCoverageUnitTest#allAlwaysSetExternalLibPropertiesMustBeInSystemPropertiesAfterSetup` — new: clears all catalogue keys, calls `setKeySystemProperties()`, asserts all 5 are non-null in `System.getProperty()`
- [x] `PropertiesHelperCoverageUnitTest#proxyExternalLibPropertiesMustBeInSystemPropertiesWhenProxyIsEnabled` — new: enables fake proxy, calls `configureJVMProxy()`, asserts all 6 proxy keys are non-null in `System.getProperty()`
- [x] `TestNGListenerHelperCoverageUnitTest#configureJVMProxyShouldSetSystemProxyPropertiesWhenEnabled` — updated: asserts all 6 proxy keys in `System.getProperty()` after `configureJVMProxy()`
- [x] `TestNGListenerCoverageUnitTest#alterShouldRouteCucumberSuiteToCucumberHelperWithoutThrowing` — Cucumber detection via class inspection
- [x] `TestNGListenerCoverageUnitTest#alterShouldRouteTestNGSuiteToConfigureTestNGPropertiesWithoutThrowing` — TestNG path + lazy init
- [x] `TestNGListenerCoverageUnitTest#configureTestNGPropertiesShouldLazilyInitPropertiesTestNGWhenNull` — forces `Properties.testNG = null`, verifies no NPE
- [x] `DriverFactoryHelperUnitTest#edgeDriverInitClearsMirrorUrlAfterConstruction` — new: uses `Mockito.mockConstruction(OptionsManager.class)` and `mockConstruction(EdgeDriver.class)` to invoke the real `createNewLocalDriverInstance(EDGE, 0)` without launching a browser; asserts `SE_DRIVER_MIRROR_URL` is `null` in `System.getProperty()` after the Edge arm completes

## Known failing checks

- **`codecov/project`** and **`codecov/patch`** — expected, not fixed in this PR. Same pattern as PR #2646:
  - Base `main` coverage (~78%) is measured by 11 full E2E jobs in `e2eTests.yml` (real browsers, BrowserStack devices, Postgres). This branch only uploads coverage from `unit-tests.yml`, which is unit-test-only by design.
  - Roughly half the new production lines (proxy configuration, Edge-driver init paths) require real browser/network infrastructure to exercise and are not coverable at unit-test level.
  - No production code was deleted or made unreachable. Coverage will be restored when the merged branch runs against the full E2E suite on `main`.

Closes #2687

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -565,5 +565,20 @@ Add entries to the **Session Learnings Log** section below. Keep each entry comp
 
 - Date: 2026-05-10
 - Area: Pattern / JVM system property catalogue
-- Lesson: `PropertiesHelper` now owns two public `Set<String>` catalogue constants — `JVM_SYSTEM_PROPERTIES_ALWAYS` (6 keys always set at engine startup) and `JVM_SYSTEM_PROPERTIES_PROXY` (6 proxy keys set when jvmProxySettings is enabled). Any property consumed by a third-party library via `System.getProperty()` must be added to the appropriate catalogue set; a CI-enforced unit test will fail if any key is missing after engine setup. This replaces the previous "deferred" entry.
-- Evidence: `PropertiesHelper.java`, `PropertiesHelperCoverageUnitTest`, PR #2688 commits f3774bfd70 / 563ee322b6 / 5053324423
+- Lesson: `PropertiesHelper` now owns two public `Set<String>` catalogue constants — `JVM_SYSTEM_PROPERTIES_ALWAYS` (5 keys always set at engine startup) and `JVM_SYSTEM_PROPERTIES_PROXY` (6 proxy keys set when jvmProxySettings is enabled). Any property consumed by a third-party library via `System.getProperty()` must be added to the appropriate catalogue set; a CI-enforced unit test will fail if any key is missing after engine setup.
+- Evidence: `PropertiesHelper.java`, `PropertiesHelperCoverageUnitTest`, PR #2688
+
+- Date: 2026-05-10
+- Area: Anti-pattern / Properties
+- Lesson: `SE_DRIVER_MIRROR_URL` must stay in the Edge switch arm (not global startup) and be cleared after `new EdgeDriver()`. Selenium Manager's `runCommand()` passes ALL `SE_*` JVM system properties to the selenium-manager binary process — setting it globally would redirect Chrome/Firefox driver downloads to the Edge CDN. Selenium Manager runs synchronously during driver instantiation so clearing immediately after the constructor is safe.
+- Evidence: `DriverFactoryHelper.java` Edge arm, `SeleniumManager.runCommand()` source, `DriverFactoryHelperUnitTest.edgeDriverInitClearsMirrorUrlAfterConstruction`, PR #2688
+
+- Date: 2026-05-10
+- Area: Pattern / Testing
+- Lesson: `Mockito.mockConstruction(X.class)` (Mockito 5.x core, no extra dep) intercepts all `new X()` calls inside the try-with-resources block without running the real constructor. Default return for object-typed mock methods is `null` — usually sufficient for dependencies you don't need to exercise. Prefer this over test-induced method extractions (extracting a `static void helper(Runnable)` only for testability is a code smell).
+- Evidence: `DriverFactoryHelperUnitTest.edgeDriverInitClearsMirrorUrlAfterConstruction`, PR #2688
+
+- Date: 2026-05-10
+- Area: CI / Coverage
+- Lesson: `codecov/patch` and `codecov/project` failures are expected on bug-fix branches whose CI only runs unit tests — the base `main` coverage (~78%) is measured by 11 full E2E jobs (browsers, BrowserStack, Grid). New production lines requiring real infrastructure (browsers, proxies) cannot be covered at unit-test level; document in PR description with **Known failing checks** section rather than forcing artificial coverage.
+- Evidence: PR #2688, PR #2646 (same pattern), codecov comparison against commit `3d4e395`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -547,3 +547,23 @@ Add entries to the **Session Learnings Log** section below. Keep each entry comp
 - Area: Pattern / grep / multiline
 - Lesson: Single-line `grep` patterns miss Java fluent chains split across lines (e.g., `Validations.assertThat()\n.response(res)`). When doing impact analysis for method-level refactors, also grep with multiline mode or read files directly after the first compile pass.
 - Evidence: Session 2026-05-08, 3 rounds of compile-error discovery during ValidationsBuilder fix
+
+- Date: 2026-05-10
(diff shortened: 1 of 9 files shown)


MohabMohie:
I'm going with the more targeted and more accurate fix https://github.com/ShaftHQ/SHAFT_ENGINE/pull/2693

PR #2712, opened 2026-05-13
codex: Align CI failure logs with Allure results

MohabMohie:
### Motivation
- Prevent Surefire-only failures from expanding the job summary below the Allure report so the Allure report remains the single source of truth for test failures.

### Description
- Update `scripts/ci/extract_allure_failures.py` to limit CLI Markdown output and exit-code semantics to Allure `failed`/`broken` results while keeping Surefire parsing available for JSON diagnostics and `--format json` use.
- Adjust the `--surefire-reports` argument help text and example usage to emphasize Surefire is for diagnostics only and not included in the default Markdown output.
- Change `.github/workflows/coverage-readiness.yml` so the Extract Allure failures step calls the extractor with only `allure-results` (removing the default Surefire cross-check in the Markdown path).
- Add a regression unit test `test_cli_markdown_uses_allure_as_single_source_of_truth` in `tests/scripts/test_extract_allure_failures.py` to verify Surefire-only failures are not printed in the CLI Markdown output and the script exit code follows Allure results.

### Testing
- Ran `python -m unittest tests.scripts.test_extract_allure_failures -v` and the test suite passed (all tests OK).
- Ran the full script test discovery `python -m unittest discover -s tests/scripts -p 'test_*.py'` and all tests passed (OK).
- Verified the modified Coverage Readiness workflow step now consumes the extractor output produced from `allure-results` without appending Surefire-only failures to the job summary.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a03fc588a688320b9984e1ef58c7161)

### DIFF
--- a/.github/workflows/coverage-readiness.yml
+++ b/.github/workflows/coverage-readiness.yml
@@ -78,7 +78,7 @@ jobs:
         continue-on-error: true
         run: |
           if [ -d allure-results ]; then
-            scripts/ci/extract_allure_failures.py allure-results --surefire-reports target/surefire-reports | tee allure-failures.md
+            scripts/ci/extract_allure_failures.py allure-results | tee allure-failures.md
           else
             echo 'No allure-results directory was generated.' | tee allure-failures.md
           fi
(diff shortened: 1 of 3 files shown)


MohabMohie:
@codex init
recreate this branch and work again.

The issue is still here
https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/25795438583

The objective is that each job summary should show only the same failures that are already listed in the allure report. If the report summary shows 2 failed and 1 broken then the details should show and expand on the same.

Issue #2695, opened 2026-05-11
Feature: Reusable SSH session client (JSch), multi-command sessions, SFTP, tunnels


Mochxd:
### Summary

Today `TerminalActions` opens a new JSch `Session` for each `performTerminalCommand`, runs `ChannelExec`, then disconnects. There is no public API to reuse one SSH login across multiple commands, no structured type for stdout/stderr/exit code, and JSch features like SFTP, port forwarding, or richer auth are not exposed through SHAFT.

### Value

- Cut connection overhead for multi-step remote runs.
- Stay aligned with maintained [mwiede/jsch](https://github.com/mwiede/jsch) (already on the classpath) and document how that differs from classic [JCraft JSch](http://www.jcraft.com/jsch/) capabilities.
- Make host keys and credentials something we can extend safely instead of one-off JSch code in tests.

### Plan (high level)

**PR1 --> Foundation:** `JschSessionFactory` + `SshConnectionOptions`, `RemoteSshClient` (`AutoCloseable`, one `Session`, multiple `ChannelExec`), `SshCommandResult`, extracted docker/command bundling; `TerminalActions` delegates to the factory with backward-compatible behavior; `SHAFT.CLI.remoteSsh(...)`; properties, docs, unit + env-gated integration test.

**PR2 --> Extended:** SFTP facade, local/remote port forwarding with cleanup on close, password/keyboard-interactive + `known_hosts`, exec env + optional streaming, optional experimental `ChannelShell`/PTY, reporting attachments + redaction. (Can split into sub-issues if the review gets heavy.)

### Acceptance criteria

- [ ] Existing `TerminalActions` remote behavior unchanged for callers who do not use the new API.
- [ ] New API can run multiple commands over **one** connected `Session`.
- [ ] Docs cover mwiede/OpenSSH algorithm gotchas (e.g. legacy `ssh-rsa`, `jsch.*` JVM properties) and keeping a **single** JSch artifact on the classpath.

@MohabMohie As usual I wanna hear your opinion and how we can enhance it, then I’ll hit the implementation.

MohabMohie:
@Mochxd well this looks great.
I think the CLI module is one of the oldest modules in shaft, definitely needs some love.
I believe the main issue I had with keeping the same session for multiple commands was the session timing out in case I'm executing something that takes time...
Like for example installing a CLI application and waiting for the installation to finish.

[...]

PR #2720, opened 2026-05-13
feat(cli): Reusable SSH client (RemoteSshClient), SHAFT.Properties.ssh, JSch session factory

Mochxd:
## Summary
Adds a **reusable SSH** API on top of existing JSch usage: connect once (by default), run multiple remote `exec` commands, optional **fresh session per command** for timeout/isolation scenarios discussed in #2695. Refactors `TerminalActions` to share the same connection helpers **without changing** per-call remote behavior (still a new session each `performTerminalCommand`).

Closes nothing yet (PR2 can follow for SFTP, tunnels, richer auth).

## What changed
- **`RemoteSshClient`** (`AutoCloseable`), **`SshCommandResult`**, **`SshSessionPolicy`** (`REUSE_SESSION` | `NEW_SESSION_PER_COMMAND`)
- **`SHAFT.CLI.remoteSsh(...)`** overloads (plain SSH and SSH + Docker wrap, same semantics as `TerminalActions`)
- **`SHAFT.Properties.ssh`**: `sshStrictHostKeyChecking`, `sshConnectTimeout`, `sshServerAliveInterval`
- **Internal**: `JschSessionFactory`, `SshConnectionOptions`, `RemoteCommandBundler`, `ShellCommandNormalizer`
- **`TerminalActions`**: uses factory + bundler; legacy remote flow preserved
- **Tests**: `RemoteCommandBundlerTest`, `RemoteSshClientTest` (no live SSH)
- **Docs**: `ARCHITECTURE.md` CLI bullet
- **`.mvn/jvm.config`**: remove `--sun-misc-unsafe-memory-access=allow` (rejected on JDK 25; Maven couldn’t start)

## How to try
```java
try (var ssh = SHAFT.CLI.remoteSsh("host", 22, "user", "keys/", "id_rsa")) {
    ssh.connect();
    SshCommandResult r = ssh.performCommand("uname -a");
} ```

### DIFF
--- a/.github/actions/post-test-report/action.yml
+++ b/.github/actions/post-test-report/action.yml
@@ -139,6 +139,21 @@ runs:
             echo "::warning::No tests were executed. Check test configuration."
           fi
           if [ "$FAILED" -gt 0 ] || [ "$BROKEN" -gt 0 ]; then
+            if command -v python3 >/dev/null 2>&1; then
+              ALLURE_FAILURE_SOURCE=""
+              if [ -d "allure-report" ]; then
+                ALLURE_FAILURE_SOURCE="allure-report"
+              elif [ -d "allure-results" ]; then
+                ALLURE_FAILURE_SOURCE="allure-results"
+              fi
+              if [ -n "$ALLURE_FAILURE_SOURCE" ]; then
+                {
+                  echo ""
+                  echo "### ❌ Failed and broken test methods"
+                  python3 scripts/ci/extract_allure_failures.py "$ALLURE_FAILURE_SOURCE" || true
+                } >> "$GITHUB_STEP_SUMMARY"
+              fi
+            fi
             echo "::error::Allure report shows $FAILED failed and $BROKEN broken test(s) out of $TOTAL total. Check the Allure report artifact for details."
             exit 1
           fi
--- a/.github/codex/prompts/refresh-agent-instructions.md
+++ b/.github/codex/prompts/refresh-agent-instructions.md
@@ -1,0 +1,61 @@
+# Refresh Agent Instructions
+
+You are running in GitHub Actions as a repository-maintenance agent for SHAFT_ENGINE. Your task is to refresh durable Codex/agent instructions, primarily the root `AGENTS.md`, so future agents can work effectively and safely.
+
+## Scope and safety
+- Update only durable agent guidance and instruction documentation.
+- Do not change application/product source code, tests, package manifests, lockfiles, CI workflows, generated files, build outputs, vendored dependencies, secrets, `.env` files, credentials, or unrelated docs.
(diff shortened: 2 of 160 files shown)


Mochxd:
[2026-05-13_22-36-24-877_AllureReport.html](https://github.com/user-attachments/files/27730281/2026-05-13_22-36-24-877_AllureReport.html)
[2026-05-13_22-30-02-906_AllureReport.html](https://github.com/user-attachments/files/27730290/2026-05-13_22-30-02-906_AllureReport.html)
Attached: 
(1) Allure run --> TerminalActions + RemoteCommandBundler (18 passed).
(2) Allure run --> RemoteSshClient + SshCommandResult (13 passed). 
 both local, no SSH integration in these suites
@MohabMohie Letsss go 🚀

Mochxd:
I will make the second PR now as it will help u to review them separate @MohabMohie

MohabMohie:
@Mochxd I'm not sure about the final user experience though...

My intention with SHAFT is always to have a unified entry point. No try/catch blocks. And no *if you know you know* methods that you need to know where they are in order to call them.

This PR offers a working solution, but from a consistent user experience standpoint  and acting as a product owner I don't think this is the best implementation we can deliver.

An example of what a reusable session may look like following existing shaft architecture.

- It should be similar to a webdriver session, you create it once, it's stored inside the engine, you can reuse the same session, and you terminate it when you call the quit method.
- It should have multiple constructors similar to the webdriver, or can use the builder pattern similar to the API driver. Whichever makes more sense to the implementation you're looking for.
- As a user I want to create a new CLI driver, I do not mind or care how the engine handles it from the background, maybe rather than just calling the default CLI constructor I'll call one which has a boolean reusable flag set to true.
- As a user I don't want to be exposed to any custom return types, I want to call a method and get a string (or other simple data type) to assert against.
- As an advanced user, I want to access the complex data types when/if needed by calling my shaft CLI driver and using a method inside it that returns the native object where applicable (similar to API.getResponseObject and GUI.getDriver).


The above is a human authored sample... I did not use AI to generate the response or validate it. Just pure traditional product owner work...

I get the sense that the code in this PR is AI generated and not Human created (with ai assistance). If my assumption is correct, kindly let's start from scratch in a more human-led PR on a different branch.

Mochxd:
hi @MohabMohie mohab, thanks for spelling out how you want CLI to feel in SHAFT the WebDriver lifecycle, string-first results, and a clear quit(), with native JSch only when someone opts in, all makes sense so I’m not married to the first shape of the PR; I want it to match that mental model.
for the “AI generated” bit no worries on my side, I’m not taking it personally, additional for your info i don’t use an assistant to write the code...I use tools sometimes to some check or review, but the implementation and decisions are mine. Easiest path forward for everyone is a follow up PR after we align on the approach, and we can walk through the implementation properly there

PR #2724, opened 2026-05-15
feat(cli): reusable JSch SSH session, Terminal lifecycle, SFTP and tunnels

Mochxd:
implements reusable SSH / JSch support for CLI as outlined in #2695 : `SHAFT.CLI.Terminal` with explicit `quit()` (same lifecycle idea as WebDriver), string-first commands there, and `RemoteSshClient` / `remoteSshBuilder` when tests need SFTP, port forwarding, or richer `Builder` options

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -97,7 +97,7 @@ graph TB
 ### Core Testing Modules
 - **🌐 GUI Module**: Selenium and Appium-based web and mobile automation with fluent API
 - **🔌 API Module**: REST API testing powered by REST Assured
-- **💻 CLI Module**: Command-line execution and file system operations
+- **💻 CLI Module**: Shell automation (`TerminalActions`, `SHAFT.CLI.Terminal`), optional reusable remote SSH closed with `quit()`, `RemoteSshClient` for SFTP and port forwards, file operations, and `SHAFT.Properties.ssh`
 - **🗄️ DB Module**: Database connectivity and SQL operations
 
 ### Supporting Modules
--- a/src/main/java/com/shaft/cli/RemoteSshClient.java
+++ b/src/main/java/com/shaft/cli/RemoteSshClient.java
@@ -1,0 +1,369 @@
+package com.shaft.cli;
+
+import com.jcraft.jsch.ChannelExec;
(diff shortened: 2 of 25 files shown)


MohabMohie:
@Mochxd looking at the sheer amount of changes, and at how the test files look. I believe this is still a bulk AI-generated solution that is far from the concise, focused, and user-friendly coding standard that we hold ourselves to.
I apologize, but I have no choice but to close this PR as well with the same justification as the one before it
https://github.com/ShaftHQ/SHAFT_ENGINE/pull/2720#issuecomment-4462957057

Moving forward, kindly note that this feature remains a valid feature to be added to the engine. But it needs to be far more simplified, and provide a far better user experience. Following the same comments I mentioned in the earlier PR.

Mochxd:
@MohabMohie this is my latest redesign for reusable remote ssh, if this scope is still not what you want (Ai generated), i will stop working on this feature and leave it for the maintainers

**files I will touch (3 only):**
- [ ] src/main/java/com/shaft/cli/TerminalActions.java --> Add optional reusable SSH session behind existing JSch code
- [ ] src/main/java/com/shaft/driver/SHAFT.java --> Add nested SHAFT.CLI.Terminal with quit() and string commands
- [ ] src/test/java/testPackage/unitTests/TerminalActionsUnitTest.java -->  Add a few tests in the existing class (no new test files)

**behavior in TerminalActions:**
- default (unchanged): existing constructors and SHAFT.CLI.terminal() keep the current behavior --> each remote performTerminalCommand still opens ssh, runs ChannelExec, and disconnects, no breaking change for current tests
- opt in reuse: new overload with boolean reusableRemoteSshSession, when true the first remote command connects later commands reuse the same Session until quit()
- quit(): disconnects the cached session and clears it, Safe no op on local-only terminals.
- getJschSession(): only when remote + reusable flag is true, throws a clear IllegalStateException otherwise so people do not grab a session on the wrong constructor
- implementation detail: I am not introducing RemoteSshClient or new public types. Same createSSHsession(), buildLongCommand(), and ChannelExec path as main; only obtainRemoteSession() decides connect-once vs connect-per-call, and exec channels are disconnected after each command while the session stays open when reuse is on

PR #2743, opened 2026-05-17
fix: stable property file loading and stream cleanup in PropertyFileManager

Mochxd:
this small fix in `PropertyFileManager` for how SHAFT reads the `.properties` files in `src/main/resources/properties/` at startup

### Describtion

when tests start, SHAFT loads those files into `System.getProperties()` so the rest of the engine (OWNER / `SHAFT.Properties`) can pick them up, that part is fine but the loader had a few annoying bugs:

**streams not closed**  
we were doing `new FileInputStream(...)` and never closing it, on Windows that can lock the file for a while u will notice it when tests try to delete or overwrite stuff in the properties folder, or when you rerun things in the same JVM

**file order was random**  
`FileUtils.listFiles()` doesn't guarantee order so if u ever had the same key in two files (not the normal SHAFT setup, but it happens), one machine might pick value A and another value B, it will Looks like a flaky test when it's really just load order

**System got updated file by file**  
after each file we pushed into `System.getProperties()` if a later file failed to read, u could still end up half-loaded

### what i changed
nothing public-facing, same flow as before:
- try with resources when reading each file
- sort files by path before loading so order is always the same
- load everything first, then merge overrides and write to `System` once at the end

- for normal SHAFT projects (`custom.properties` + TestNG + log4j etc. with different keys) behavior should be the same, just consistent
- If someone puts the same key in two files, it's now predictable: later file in sorted order wins, then `-D` / `SHAFT.Properties.set()` still win like before

### tests
extended `PropertyFileManagerUnitTest`:
- bad file read doesn't touch System
- new test for sorted load order (a, b, z --> z wins)

### DIFF
--- a/src/main/java/com/shaft/properties/internal/PropertyFileManager.java
+++ b/src/main/java/com/shaft/properties/internal/PropertyFileManager.java
@@ -190,17 +190,16 @@ public final class PropertyFileManager {
                     FileActions.getInstance(true).unpackArchive(url, "target/");
                     propertiesFolderPath = "target/resources/properties/default/";
                 }
-                // reading regular files
-                Collection<File> propertiesFilesList;
                 if (FileActions.getInstance(true).doesFileExist(propertiesFolderPath)) {
-                    propertiesFilesList = FileUtils.listFiles(new File(propertiesFolderPath), new String[]{"properties"},
-                            false);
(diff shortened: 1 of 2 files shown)


MohabMohie (approved):
Thank you for your valuable contribution

Issue #2780, opened 2026-05-29
[Bug]: Starting from version 10.2.20260501, the Remote WebDriver can no longer be instantiated


roaamagdy:
Till version `10.2.20260424`. the session was instantiated normally against remote server but starting from `10.2.20260501`different exception is coming up:

- Version **10.2.20260501**:
`Driver Factory Action "setRemoteDriverInstance" failed. With the following test data "Failed to create Remote WebDriver instance.". Root cause: "org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Response code 500. Message: internal error "
`
- Version **10.2.20260505** gives:
```
Caused by: java.lang.NoClassDefFoundError: com/epam/reportportal/testng/ITestNGService
	at com.shaft.driver.DriverFactory.reloadProperties(DriverFactory.java:79)
	at com.shaft.driver.DriverFactory.readLastMinuteUpdatedProperties(DriverFactory.java:57)
	at com.shaft.driver.DriverFactory.getHelper(DriverFactory.java:175)
	at com.shaft.driver.SHAFT$GUI$WebDriver.<init>(SHAFT.java:120)

```
- Version **10.2.20260523** : `java.lang.RuntimeException: Driver Factory Action "setRemoteDriverInstance" failed. With the following test data "Failed to create Remote WebDriver instance.". Root cause: "org.openqa.selenium.SessionNotCreatedException: Could not start a new session. Response code 500. Message: internal error "
	at com.shaft.tools.io.internal.FailureReporter.fail(FailureReporter.java:34)
	at com.shaft.driver.internal.DriverFactory.DriverFactoryHelper.failAction(DriverFactoryHelper.java:133)
	at com.shaft.driver.internal.DriverFactory.DriverFactoryHelper.setRemoteDriverInstance(DriverFactoryHelper.java:682)
	at com.shaft.driver.internal.DriverFactory.DriverFactoryHelper.configureRemoteDriverInstance(DriverFactoryHelper.java:690)
	at com.shaft.driver.internal.DriverFactory.DriverFactoryHelper.createNewRemoteDriverInstance(DriverFactoryHelper.java:614)
	at com.shaft.driver.internal.DriverFactory.DriverFactoryHelper.initializeDriver(DriverFactoryHelper.java:860)
	at com.shaft.driver.DriverFactory.getHelper(DriverFactory.java:182)
[...]

PR #2781, opened 2026-05-29
fix remote webdriver bootstrap and restore jvm system properties

Mochxd:
Fixes for #2780
### What was wrong

1. After #2488, keys like `webdriver.http.factory` were only written to SHAFT's global property map, selenium still reads `System.getProperty`, so remote sessions often failed with a 500 from the grid

2. `DriverFactory.reloadProperties()` self-heal called `TestNGListener`, which loads ReportPortal types at class init, hat blows up if listeners are not set up yet or ReportPortal is not on the classpath

### What i did

- dual write JVM keys (`webdriver.http.factory`, log4j, rp path, Allure flags, proxy settings) to both `System` and SHAFT global properties
- move `identifyRunType()` to `ProjectStructureManager` and engine startup to `PropertiesHelper.bootstrapEngine()`
- `DriverFactory` uses those paths instead of loading `TestNGListener` during self-heal. `TestNGListener.engineSetup()` still delegates for normal listener runs

### DIFF
--- a/src/main/java/com/shaft/driver/DriverFactory.java
+++ b/src/main/java/com/shaft/driver/DriverFactory.java
@@ -7,8 +7,8 @@ import com.shaft.db.DatabaseActions.DatabaseType;
 import com.shaft.driver.internal.DriverFactory.BrowserStackHelper;
 import com.shaft.driver.internal.DriverFactory.DriverFactoryHelper;
 import com.shaft.driver.internal.DriverFactory.LambdaTestHelper;
-import com.shaft.listeners.TestNGListener;
 import com.shaft.listeners.internal.TestNGListenerHelper;
+import com.shaft.properties.internal.PropertiesHelper;
 import com.shaft.properties.internal.ThreadLocalPropertiesManager;
 import com.shaft.tools.io.internal.ProjectStructureManager;
 import lombok.Getter;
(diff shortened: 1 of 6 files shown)


MohabMohie (approved):
The changes look okay, THANK YOU for your valued contribution!

Notes for the future:
- kindly keep the changes surgical and minimalistic, for instance moving the bootstrapping code and renaming the method both make sense, but are not directly related to the issue we're fixing. Having these changes along with the main issue fix makes reviewing the PR more challenging. I would highly appreciate it if we have two separate PRs, one for the surgical issue fix (adding things to system.setproperties) and another one for infrastructure enhancement (moving and renaming the bootstrapping method, and adding a default branch + any other local review comments)
- kindly attach an execution report or screenshots for any local testing performed. Since we disabled the e2e testing pipeline from running against PRs we need to have such evidences before merging to the main branch. For now there is no load and we can afford to merge and then test on the main branch, but for future I believe running the testing job on your fork before merging and attaching a link to the PR is a better way.

MohabMohie:
Link to relevant test run https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/26676937329

MohabMohie:
CI failure traceability: follow-up issue #2782 tracks the remaining native Selenium Grid E2E failures from run 26676937329 after this PR fixed the #2780 remote WebDriver/bootstrap regression.

PR #2899, opened 2026-06-14
add command environment variables to remote terminal

Mochxd:
Part of #2695
- Adds optional environment variables to terminal command execution on the existing TerminalActions API.

**changes**
- performTerminalCommand(String command, Map<String, String> environmentVariables)
- performTerminalCommands(List<String> commands, Map<String, String> environmentVariables)
- both still return String like the existing methods

**Behavior**
- Local: env vars are added to the spawned process environment
- Remote: env vars are sent through SSH setEnv on the exec channel (needs server AcceptEnv when configured)
- Dockerized remote: env vars are injected into the container via docker exec -e
- Existing no-env methods are unchanged
- Env is per command, not stored on the terminal instance

**Tests**
TerminalActionsUnitTest  --> 44 passed
[Env-Tests.html](https://github.com/user-attachments/files/28926793/Env-Tests.html)

**Usage**
TerminalActions terminal = SHAFT.CLI.remoteTerminal("host", 22, "user", "keys/", "id_rsa");
String output = terminal.performTerminalCommand(
        "echo $APP_ENV",
        Map.of("APP_ENV", "staging"));
terminal.quit();

### DIFF
--- a/shaft-engine/src/main/java/com/shaft/cli/TerminalActions.java
+++ b/shaft-engine/src/main/java/com/shaft/cli/TerminalActions.java
@@ -233,10 +233,28 @@ public class TerminalActions {
     }
 
     public String performTerminalCommands(List<String> commands) {
+        return performTerminalCommands(commands, Collections.emptyMap());
+    }
+
+    /**
+     * Executes one or more terminal commands with the supplied environment variables.
+     *
+     * <p>For local terminals the variables are added to the spawned process environment.
+     * For dockerized remote terminals they are injected into the container through
+     * {@code docker exec -e}. For non-dockerized remote terminals they are sent as SSH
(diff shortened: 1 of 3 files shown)


Mochxd:
@MohabMohie not sure this is the best way to handle env on terminal commands so i need ur opinion if u want some enhancement, thank u as usual for ur supoort

MohabMohie:
@Mochxd it looks good to me. Also, I think the dockerized execution and all related properties need to be deprecated for removal. Perhapse we can deprecated it now and avoid making changes to it.
What do you think?

Mochxd:
@MohabMohie yes i agree with you, we can deprecate dockerized terminal execution and the related timeout property now and stop extending that path.
I'll drop the docker exec -e part from this PR and keep env for local + plain remote only

PR #2926, opened 2026-06-17
send SSH keep-alive packets during long remote terminal sessions

Mochxd:
Part of #2695

Addresses the long-running remote command scenario, reusable SSH sessions can be dropped by firewalls or idle timeouts while a command like a long install is still running

## What changed
- configure JSch `ServerAliveInterval` (60 seconds) before remote SSH `connect()`
- applies to all remote sessions created by `TerminalActions` (reusable and one-shot)
-uses `Session.setServerAliveInterval()` (mwiede/jsch recommended path; config Properties alone is unreliable)

## Behavior
- sends lightweight keep-alive traffic on the SSH session during long, quiet commands
- does **not** change `shellSessionTimeout` or the existing reusable-session `quit()` lifecycle
- return values and public API unchanged

@MohabMohie I used a fixed 60s keep-alive interval for a small first slice what do u think for this?

- a `Properties.timeouts` setting instead of a constant, or
- keep-alive only on reusable `remoteTerminal()` sessions

### DIFF
--- a/shaft-engine/src/main/java/com/shaft/cli/TerminalActions.java
+++ b/shaft-engine/src/main/java/com/shaft/cli/TerminalActions.java
@@ -534,17 +534,13 @@ public class TerminalActions {
             }
             session = jsch.getSession(sshUsername, sshHostName, sshPortNumber);
             session.setConfig(config);
-            applySshKeepAliveSettings(session);
+            session.setServerAliveInterval(SSH_SERVER_ALIVE_INTERVAL_MILLISECONDS);
             session.connect();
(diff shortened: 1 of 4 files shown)


MohabMohie:
@Mochxd let's expose it as a property with the rest of the timeouts if none of the existing ones make sense to reuse please.

Mochxd:
@MohabMohie done check the tests too
[longRemoteTerminalSessions.html](https://github.com/user-attachments/files/29103909/longRemoteTerminalSessions.html)

Mochxd:
@MohabMohie done check the tests too
[longRemoteTerminalSessions.html](https://github.com/user-attachments/files/29103909/longRemoteTerminalSessions.html)

Issue #3322, opened 2026-07-06
Assistant local-agent runtime fidelity: live streaming, real tokens, real models, auto-compact


MohabMohie:
## Summary

Bug 19 is a self-contained event-semantics defect in the recorder's keyboard capture: editing keys pressed while typing into an input are emitted as separate keyboard actions instead of being folded into one final `type` action. The root cause (raw keydown emission vs. final-value coalescing) is independent of the persistence/lifecycle cluster and of the assertion UX cluster, so it ships alone.

The recorder should suppress standalone action emission for editing keys (Backspace, Delete, arrow keys) when focus is inside a text-entry element, and instead coalesce the whole editing burst into a single `type` action carrying the element's final value — emitted on blur/change/submit or before the next non-typing action. Non-input contexts (e.g. Backspace as page navigation or on a focused button) must keep their current behavior.

**Ship unit:** `coalesce-input-editing-keys`

Primary files:
- `shaft-capture/src/main/resources/browser/shaft-capture-recorder.js`
- `shaft-capture/src/main/java/com/shaft/capture/runtime/CaptureEventPipeline.java` (coalescing here too if raw key events reach the pipeline)
- `shaft-capture/src/test/java/com/shaft/capture/runtime/CaptureEventPipelineTest.java`

## Bugs covered

- 19: Pressing Backspace while typing is recorded as its own keyboard action; the complete final text should be recorded as one type action

## Acceptance criteria

- [ ] A recorder test types `abcd`, presses Backspace twice, types `xy`, blurs the field, and asserts exactly one recorded `type` action with value `abxy` and zero standalone Backspace/keyboard actions in the session JSON
- [ ] The recorder UI step list for that sequence shows a single `type` step (playwright screenshot)
- [ ] Backspace outside a text-entry element still records (or navigates) exactly as before — covered by a regression assertion in the diff's tests
- [ ] Regression coverage added for type-then-backspace-then-type sequences

**Evidence:** playwright-screenshot

Issue #3320, opened 2026-07-06
Remove trailing .perform() from generated assertions and the official user guide (Bug 18)


MohabMohie:
## Summary

Bugs 10-13 all live in the same subsystem: the IntelliJ plugin's local-agent runner integration (`AssistantLocalAgentRunner` plus the panel's stream-token plumbing in `ShaftAssistantPanel` around lines 813-1240). Every one of them is about surfacing **real** agent-process data — live output stream, actual token usage, actual supported models — and shaping the request pipeline (auto-compact before each request), rather than Swing layout. This is a distinct root cause from the layout cluster even though it is the same screen, so it ships as one focused unit.

The goal is to upgrade `AssistantLocalAgentRunner` and `ShaftAssistantPanel`'s agent path so that:

- **Verbose mode streams real output.** When Verbose is checked, pipe the agent process stdout/stderr line-by-line through the existing `appendLocalAgentOutput(streamToken, ...)` callback into the streaming bubble in real time (EDT-safe via `invokeLater`), replacing the static `Running Local Assistant` placeholder.
- **Token counts are real.** Parse the agent CLI's structured result metadata (e.g. `claude` CLI `--output-format json` usage fields) and display actual input/output token counts, falling back to the estimate only when the agent reports none (and labeling it `estimated`).
- **Model picker is live.** Query the connected agent CLI for its supported model list (e.g. models listing command/flag) and populate the model selector from that live list instead of a hardcoded one.
- **Context is compacted per request.** Before dispatching each new user request, send the agent's compact/compress command (e.g. `/compact`) as a preamble step, with a settings toggle and graceful skip when the CLI does not support it.

## Bugs covered

- 10: Verbose checkbox shows only a static 'Running Local Assistant' message instead of a full real-time stream of agent output
- 11: Token count shown is an estimate, not the real tokens consumed by the underlying agent
- 12: Model picker does not show the real list of models the connected agent supports (e.g. Sonnet 5, Opus 4.8)
- 13: No context-compact/compress command is sent ahead of each new request to keep context compressed

## Acceptance criteria

- [ ] With Verbose checked, each line the fake/stubbed agent process emits appears in the transcript bubble before process exit (test drives a stub process emitting 3 delayed lines and asserts 3 incremental `appendLocalAgentOutput` deliveries, all marshalled to the EDT)
[...]

PR #3339, opened 2026-07-06
Fix cross-origin recorder panel state loss

### LINKED ISSUE #3319: Web recorder session lifecycle: cross-domain continuity, real-time JSON persistence, and browser shutdown
## Summary

_This issue's body previously showed content for an unrelated cluster (Assistant local-agent runtime fidelity, now shipped separately as #3322) due to a title/body pairing bug in the batch issue-creation tooling used to file this run's clusters. Correcting it here to match the actual title and the work in progress._

The web recorder's session lifecycle has three concerns:

- **Real-time JSON persistence**: `CaptureSessionStore` must write through to disk atomically on every `start()`/`append()`/`updateEvents()` call, so no in-flight recording is lost if the process or browser dies mid-session.
- **Cross-domain continuity**: the recorder's in-page step list must survive navigating to a different origin mid-recording (e.g. a login page handing off to a separate payment-gateway domain). Browser-side page-scoped storage (`sessionStorage`) is wiped on cross-origin navigation, so the recorder UI needs to rehydrate its visible step list from the server-side session rather than relying on local storage.
- **Browser shutdown**: stopping a capture (`capture_stop` / `ManagedCaptureRecorder.stop()`) must reliably quit the recording browser/driver process and leave the on-disk session file `COMPLETE` with every performed action present.

## Acceptance criteria

- [ ] `CaptureSessionStore` writes each event through to disk atomically at capture time (already correct going in; regression-covered)
- [ ] The recorder UI step list, exposed via a server-side steps accessor (`ManagedCaptureRecorder.steps()` / a loopback `/steps` endpoint), keeps steps recorded before a cross-origin navigation and appends the post-navigation domain's actions to the same session rather than starting a new one -- verified by a real two-domain (different hostnames) Selenium test, not just a same-origin simulation
- [ ] `capture_stop` quits the recording browser/driver (the WebDriver session is unusable afterward) and leaves the on-disk session `COMPLETE` with every action from both domains present
- [ ] Loopback CORS endpoints used by the recorder (`/event`, `/steps`) correctly answer preflight `OPTIONS` requests and advertise `Access-Control-Allow-Private-Network` where required, so Chromium's Private Network Access policy doesn't silently block cross-origin-adjacent loopback calls

### Affected files

- `shaft-capture/src/main/java/com/shaft/capture/collector/BidiBrowserEventCollector.java`
- `shaft-capture/src/main/java/com/shaft/capture/collector/BrowserEventScript.java`
- `shaft-capture/src/main/java/com/shaft/capture/collector/PollingBrowserEventCollector.java`
- `shaft-capture/src/main/java/com/shaft/capture/control/CaptureControlClient.java`
- `shaft-capture/src/main/java/com/shaft/capture/control/CaptureControlServer.java`
- `shaft-capture/src/main/java/com/shaft/capture/model/CaptureStep.java`
- `shaft-capture/src/main/java/com/shaft/capture/runtime/BrowserEventSink.java`
- `shaft-capture/src/main/java/com/shaft/capture/runtime/CaptureEventPipeline.java`
- `shaft-capture/src/main/java/com/shaft/capture/runtime/CaptureManager.java`
- `shaft-capture/src/main/java/com/shaft/capture/runtime/ManagedCaptureRecorder.java`
- `shaft-capture/src/main/java/com/shaft/capture/storage/CaptureSessionStore.java`
- `shaft-capture/src/main/resources/browser/shaft-capture-recorder.js`

**Evidence expected:** screenshots of the recorder panel step list surviving a real cross-origin navigation, captured from a live two-domain browser test.

MohabMohie:
## Summary

Closes #3319. That issue's own body was corrupted by a bug in the batch issue-creation tooling used earlier today — its body is a verbatim duplicate of unrelated issue #3322's content, also duplicated onto #3320/#3321. I worked from the title alone plus direct investigation of the actual "SHAFT Capture" recorder session-lifecycle code.

Investigation findings against the three concerns named in the title:
- **Real-time JSON persistence** — already correct. `CaptureSessionStore.update()` does an atomic read-modify-write on every event; no buffering risk.
- **Browser shutdown** — already correct. `ManagedCaptureRecorder`'s 1-second health check detects browser death and marks the session `INCOMPLETE` with a timestamp; nothing is lost.
- **Cross-domain continuity** — real, reproducible bug, fixed here: the in-page recorder overlay persisted its panel state (action list, next id, readiness, instanceId) to `sessionStorage`, which browsers clear on cross-origin navigation. A multi-domain recording (e.g. login -> payment gateway) made the panel look like it had silently restarted, even though the server-side session JSON never lost anything.

## Fix

- `BrowserEventSink` now answers a token-authenticated `GET` on its existing loopback `/event` endpoint with the live session snapshot (`instanceId`, `eventCount`, `readinessState`), sourced from `ManagedCaptureRecorder.sessionSnapshot()`.
- `BidiBrowserEventCollector`'s preload script — which normally wins the reinstall race over the polling fallback collector on a fresh document, and previously never received the event sink at all — now carries the same `{url, token}` sink object (embedded as a literal in the preload source, no new Selenium BiDi API surface used), so the in-page script can always reach the sink regardless of which collector installs first.
- `shaft-capture-recorder.js`: on a fresh top-level document with no locally-persisted state, the recorder asks the sink whether a session is already in progress. If so, it keeps the stable `instanceId`, continues action numbering from the server's event count, adopts the current readiness state, and announces "Continuing recording after navigating to a new domain (N action(s) already captured)" instead of silently looking like a new recording.

## Test plan

- [x] `BrowserEventScriptTest`-style additions in `CaptureCollectorUtilityTest`: new `preloadFunction(attrs, endpoint, token)` overload embeds the sink correctly and falls back to the plain preload when blank/null.
- [x] New `BrowserEventSinkTest`: authorized GET returns the supplied snapshot JSON with CORS header; missing/wrong token gets 403; existing POST behavior unchanged.
- [x] `ManagedCaptureRecorderControlTest`: new test asserts `sessionSnapshot()` reflects event count/readiness and keeps a stable `instanceId` across signals, without launching a browser.
- [x] `mvn -pl shaft-capture test-compile` and targeted `mvn -pl shaft-capture test -Dtest=BrowserEventSinkTest,CaptureCollectorUtilityTest,ManagedCaptureRecorderControlTest` — 10/10 passing.

No live-browser cross-origin E2E was run (requires the `external-e2e`-tagged infra per repo convention); the fix is covered at the unit level on both the Java state/auth logic and the JS embedding logic.

### DIFF
--- a/shaft-capture/src/main/java/com/shaft/capture/collector/BidiBrowserEventCollector.java
+++ b/shaft-capture/src/main/java/com/shaft/capture/collector/BidiBrowserEventCollector.java
@@ -19,6 +19,8 @@ public final class BidiBrowserEventCollector implements BrowserEventCollector {
 
     private final WebDriver driver;
     private final List<String> testIdAttributes;
+    private final String eventEndpoint;
+    private final String eventToken;
     private final Map<String, String> promptTypes = new ConcurrentHashMap<>();
     private Script script;
     private BrowsingContextInspector contexts;
@@ -40,11 +42,31 @@ public final class BidiBrowserEventCollector implements BrowserEventCollector {
      * @param testIdAttributes locator test-id attributes
      */
     public BidiBrowserEventCollector(WebDriver driver, List<String> testIdAttributes) {
+        this(driver, testIdAttributes, "", "");
+    }
+
+    /**
+     * Creates a BiDi collector whose preload script also carries the loopback event sink,
+     * so the in-page recorder can rehydrate its panel state after a cross-origin navigation
+     * re-installs it on a fresh, storage-isolated document.
+     *
+     * @param driver BiDi-capable driver
+     * @param testIdAttributes locator test-id attributes
+     * @param eventEndpoint loopback event endpoint
+     * @param eventToken loopback event token
+     */
+    public BidiBrowserEventCollector(
+            WebDriver driver,
+            List<String> testIdAttributes,
+            String eventEndpoint,
+            String eventToken) {
         if (driver == null) {
             throw new IllegalArgumentException("Capture WebDriver is required.");
         }
         this.driver = driver;
         this.testIdAttributes = testIdAttributes == null ? List.of() : List.copyOf(testIdAttributes);
+        this.eventEndpoint = eventEndpoint == null ? "" : eventEndpoint;
+        this.eventToken = eventToken == null ? "" : eventToken;
     }
 
     @Override
@@ -66,7 +88,7 @@ public final class BidiBrowserEventCollector implements BrowserEventCollector {
             }
         });
         preloadId = script.addPreloadScript(
-                BrowserEventScript.preloadFunction(testIdAttributes),
+                BrowserEventScript.preloadFunction(testIdAttributes, eventEndpoint, eventToken),
                 List.of(new ChannelValue(CHANNEL)));
 
         contexts = new BrowsingContextInspector(driver);
--- a/shaft-capture/src/main/java/com/shaft/capture/collector/BrowserEventScript.java
+++ b/shaft-capture/src/main/java/com/shaft/capture/collector/BrowserEventScript.java
@@ -38,6 +38,28 @@ public final class BrowserEventScript {
     }
 
     /**
+     * Returns the BiDi preload function declaration with a loopback event sink so the
+     * in-page recorder can rehydrate its panel state after a cross-origin navigation,
+     * where BiDi re-installs this script on a fresh, storage-isolated document.
+     *
+     * @param testIdAttributes locator test-id attributes
+     * @param eventEndpoint loopback event endpoint
+     * @param eventToken loopback event token
+     * @return JavaScript function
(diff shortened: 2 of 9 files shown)


MohabMohie:
Superseding with a more complete fix for #3319: this PR's continuation signal (message-only) doesn't restore the actual step list in the panel, treats real-time persistence/shutdown as already-correct without evidence, and CodeQL (java) is currently failing here. A separate in-flight branch (`claude/issue-3319-write-through-session-store-and-stop-lifecycle`) adds a `/steps` endpoint that rehydrates the real step descriptions into the visible panel across cross-origin navigation, fixes a CORS/Private-Network-Access preflight gap, and includes a live two-domain Selenium E2E test with screenshot evidence. Closing this one to avoid conflicting changes to the same files; will link the replacement PR here once opened.

MohabMohie:
Replacement PR: #3342

PR #3381, opened 2026-07-08
Fix mobile capture CA directory permissions breaking Linux CI

MohabMohie:
## Summary
- Fixes PR #3378's `release-candidate` CI failure (ubuntu-22.04): `MobileApiCaptureControllerTest` and `MobileServiceConfigurationTest` failed with `Capture CA could not be loaded or generated.`, even after #3379 wired in `@TempDir`-backed CA isolation for those tests.
- Root cause: `CaptureCertificateAuthority`'s constructor restricts its own storage directory to owner **read+write only** via `OwnerOnlyFilePermissions.restrictToOwner(directory)`. On POSIX systems, a directory needs its own **execute (search)** bit for anyone — including its owner — to traverse into it and access files by name. Every subsequent write of the key/cert files into that now-locked-down directory fails with a permission error, which the constructor wraps as `CaptureProxyException("Capture CA could not be loaded or generated.")`.
- This only reproduces on POSIX (the Linux CI runner): Windows' ACL fallback path is masked by the OS's default "bypass traverse checking" privilege, so the same code silently "worked" locally.
- Fix: add `OwnerOnlyFilePermissions.restrictDirectoryToOwner(Path)`, a directory-aware variant that keeps the owner's execute/traverse permission (POSIX `rwx------`; ACL fallback adds `EXECUTE`), and use it for the CA directory instead of the file-oriented `restrictToOwner(Path)`. `restrictToOwner` itself is unchanged for its existing file call sites (the CA key/cert files, and `SecretHeaderReplacer`'s secrets file).

## Test plan
- [x] `mvn -pl shaft-capture test -Dtest=OwnerOnlyFilePermissionsTest` — includes a new regression test (`restrictedDirectoryRemainsTraversableAndWritableByItsOwner`) that reproduces the exact failure mode: writing a file into a directory immediately after restricting it.
- [x] `mvn -pl shaft-capture-proxy test -Dtest=CaptureCertificateAuthorityTest`
- [x] `mvn -pl shaft-mcp test -Dtest=MobileApiCaptureControllerTest,MobileServiceConfigurationTest` — the originally failing tests, now passing.

Note: all verification above ran on Windows, where this bug doesn't reproduce (see root cause above) — the fix targets POSIX (Linux CI) semantics specifically. The regression test asserts the actual POSIX permission bits and directory-writability, so it will exercise the real behavior on the Linux CI runner for this PR.

### DIFF
--- a/shaft-capture-proxy/src/main/java/com/shaft/capture/proxy/CaptureCertificateAuthority.java
+++ b/shaft-capture-proxy/src/main/java/com/shaft/capture/proxy/CaptureCertificateAuthority.java
@@ -83,7 +83,7 @@ public final class CaptureCertificateAuthority {
         this.directory = directory;
         try {
             Files.createDirectories(directory);
-            OwnerOnlyFilePermissions.restrictToOwner(directory);
+            OwnerOnlyFilePermissions.restrictDirectoryToOwner(directory);
             Path keyPath = directory.resolve(KEY_FILE);
             Path certPath = directory.resolve(CERT_FILE);
             if (Files.exists(keyPath) && Files.exists(certPath)) {
--- a/shaft-capture/src/main/java/com/shaft/capture/storage/OwnerOnlyFilePermissions.java
+++ b/shaft-capture/src/main/java/com/shaft/capture/storage/OwnerOnlyFilePermissions.java
@@ -14,7 +14,7 @@ import java.util.List;
 import java.util.Set;
 
 /**
- * Restricts a file to owner-only access on any filesystem. Extracted from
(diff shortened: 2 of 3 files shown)


MohabMohie:
Fixed the Codacy finding: `doesNotThrowForAMissingDirectory()` (shaft-capture/src/test/java/com/shaft/capture/storage/OwnerOnlyFilePermissionsTest.java) had no assertion (PMD `JUnitTestsShouldIncludeAssert`). Wrapped the call in `assertDoesNotThrow`, matching the pattern already used elsewhere in this file — pushed in 04898f3.

Issue #3385, opened 2026-07-08
SHAFT Capture: browser E2E coverage gap + redundant collector polling


MohabMohie:
## Context

Follow-ups identified while diagnosing the flaky `ManagedCaptureRecorderBrowserTest.overlayGuidedAssertionFlowPersistsVerificationEvents` failure on [PR #3378](https://github.com/ShaftHQ/SHAFT_ENGINE/pull/3378)'s `release-candidate` job, fixed in [PR #3384](https://github.com/ShaftHQ/SHAFT_ENGINE/pull/3384). Neither item is a correctness bug on its own — both are process/efficiency gaps surfaced by that investigation — so they're deliberately not bundled into the fix PR.

## 1. Browser E2E coverage only runs at release time, not on ordinary PRs

`.github/workflows/shaft-pilot-release.yml` (the `SHAFT Pilot Release Candidate` workflow) is the **only** CI job that runs `ManagedCaptureRecorderBrowserTest` / `CaptureGeneratedReplayBrowserTest` against real headless Chrome/Edge (`-DincludeCaptureBrowserE2E`). It triggers only on:

```yaml
on:
  pull_request:
    paths:
      - "pom.xml"
      - ".github/workflows/shaft-pilot-release.yml"
  workflow_dispatch:
```

This means a PR that adds or changes `shaft-capture` overlay/recorder behavior — including new browser E2E test methods — never actually exercises them in CI until a release PR bumps `pom.xml`. That's exactly what happened here: `overlayGuidedAssertionFlowPersistsVerificationEvents` was added by #3374 (which didn't touch `pom.xml`), so its first-ever CI run was on the #3378 release PR, weeks later, where a real layout bug in the overlay finally surfaced as a flaky click.

**Suggested fix:** add `shaft-capture/**` (or at least `shaft-capture/src/**`) to the `paths` trigger for this workflow (or split the browser-E2E step into its own lighter job that runs on `shaft-capture` PRs), so overlay regressions are caught on the PR that introduces them instead of at release time.

## 2. Dual BiDi + Polling event collectors run concurrently regardless of BiDi health

`ManagedCaptureRecorder.startCollector()` (`shaft-capture/src/main/java/com/shaft/capture/runtime/ManagedCaptureRecorder.java`) always builds a `CompositeBrowserEventCollector` running **both** `BidiBrowserEventCollector` and `PollingBrowserEventCollector` together, not one-or-the-other despite the "fallback" naming:

[...]

PR #3387, opened 2026-07-08
Back off SHAFT Capture polling collector once BiDi is proven healthy

MohabMohie:
## Summary
- `ManagedCaptureRecorder.startCollector()` always ran a `BidiBrowserEventCollector` and a `PollingBrowserEventCollector` concurrently. The polling collector re-ran its `executeScript` install/drain round-trips, plus WebDriver-based window/navigation polling, on a fixed 100ms tick **regardless of whether BiDi was already delivering the same signals** — redundant renderer main-thread churn on every capture session.
- Worse: BiDi and polling identify windows via different ID schemes (BiDi browsing-context IDs vs. WebDriver window handles), so `CaptureEventPipeline.logicalWindow(...)` would map the *same* real tab open/close to two different logical windows if both collectors fired for it — a plausible duplicate-event bug, not just overhead.
- Adds `BidiActivityGate`: a small shared health tracker. A thin wrapper around the BiDi collector's signal consumer (in `ManagedCaptureRecorder`, `BidiBrowserEventCollector` itself is untouched) calls `recordActivity()` on every BiDi-delivered signal. `PollingBrowserEventCollector` checks the gate each tick:
  - **BiDi healthy**: skips emitting window/navigation signals (BiDi already covers them) and skips the `executeScript` round-trips entirely — but still walks the cheap WebDriver-only window/URL cursors so they never go stale.
  - **BiDi quiet** (no signal within the grace period, default 2s): resumes full-rate polling exactly as before, with cursors already current, so there's no catch-up burst of stale duplicate signals on resume.
- All existing `PollingBrowserEventCollector` constructors keep working unchanged (they delegate to the new gate-aware constructor with a default `() -> false` gate, i.e. today's always-poll behavior).

Closes the second half of #3385 (the CI coverage-gap half was #3386).

## Test plan
- [x] New unit test `PollingBrowserEventCollectorBidiBackoffTest` (no browser needed — a hand-rolled `WebDriver`/`JavascriptExecutor` stub): proves `executeScript` stops firing while the gate reports healthy, proves no `window_open` signal is emitted for a window that opens during that window, and proves resuming after the grace period lapses does **not** replay that window as new (cursor was kept current throughout).
- [x] Verified the test actually catches the regression: temporarily hard-coded `bidiHealthy = false` in `poll()`, watched the test fail deterministically (`executeScript` count `2` → `6`), then restored the real gate check.
- [x] `mvn -pl shaft-capture test` (full non-E2E suite): 213/213 passed.
- [x] `mvn -pl shaft-capture test -DincludeCaptureBrowserE2E -Dtest=ManagedCaptureRecorderBrowserTest,CaptureGeneratedReplayBrowserTest` (real headless Chrome + Edge, including the popup/window-open journey): 9/9 passed — confirms the real BiDi wiring in `ManagedCaptureRecorder` still works end-to-end after the rewrite.

### DIFF
--- a/shaft-capture/src/main/java/com/shaft/capture/collector/BidiActivityGate.java
+++ b/shaft-capture/src/main/java/com/shaft/capture/collector/BidiActivityGate.java
@@ -1,0 +1,53 @@
+package com.shaft.capture.collector;
+
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BooleanSupplier;
+
+/**
+ * Tracks whether a {@link BidiBrowserEventCollector} running alongside a
+ * {@link PollingBrowserEventCollector} has delivered a signal recently, so the polling collector
+ * can skip its own redundant script round-trips while BiDi is proven healthy, and resume full-rate
+ * polling the moment BiDi goes quiet.
+ */
+public final class BidiActivityGate implements BooleanSupplier {
+    private static final Duration DEFAULT_GRACE_PERIOD = Duration.ofSeconds(2);
+    private static final long NEVER = Long.MIN_VALUE;
+
+    private final long gracePeriodNanos;
+    private final AtomicLong lastActivityNanos = new AtomicLong(NEVER);
+
+    /**
+     * Creates a gate using the default two-second grace period.
+     */
+    public BidiActivityGate() {
+        this(DEFAULT_GRACE_PERIOD);
+    }
+
+    /**
+     * Creates a gate with a custom grace period.
+     *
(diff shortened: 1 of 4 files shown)


MohabMohie:
Addressed the 3 Codacy PMD \`UncommentedEmptyMethodBody\` findings (\`PollingBrowserEventCollectorBidiBackoffTest.java\` lines 107, 136, 140 — the \`RecordingDriver\` stub's \`get()\`/\`close()\`/\`quit()\` no-ops) by adding a one-line comment to each explaining why it's intentionally empty, in f444aa6.

Issue #3390, opened 2026-07-08
Agentic config enhancements: skill trigger overlap, unpinned MCP image, dead budget key


MohabMohie:
## Context

Found while auditing and optimizing all agentic AI configuration (AGENTS.md/CLAUDE.md, skills, MCP config, Claude Code settings, and the `.memory/` store) against industry-standard practices, per [[bug-handling-policy]] this session's non-blocking findings are consolidated into this one ticket. (Blocking/contradictory items found in the same pass were already fixed inline in the companion PR — stale/contradictory `.memory` objects marked stale or corrected, a stale doc comment in `guard.py` fixed, and a routing gap in AGENTS.md closed.)

## Findings

1. **Skill trigger overlap risk** — `.agents/skills/ci-failure-investigator/SKILL.md`'s description ("failed GitHub Actions, SHAFT Allure reports, or provider E2E runs") overlaps `.agents/skills/allure-extent-report-operator/SKILL.md`'s ("SHAFT Allure/Extent reports"), so a plain "look at this Allure report" request could match either skill ambiguously. Suggest narrowing `ci-failure-investigator`'s trigger to failed-run/CI triage specifically.
2. **Ambiguous "workflows" wording** — `.agents/skills/mcp-transport-contract-auditor/SKILL.md`'s description uses the bare word "workflows", which can attract GitHub Actions/CI-shaped requests that actually belong to `ci-failure-investigator`. Suggest rewording to something like "shaft-mcp transport/tool-contract workflows" to disambiguate from CI workflows.
3. **Stale/unreachable tool references** — `.github/skills/shaft-marketing-ad-producer/SKILL.md`'s Stack section routes to `creative-production:*` skills, `imagegen`, and `browse`, none of which exist in any configured marketplace or the enabled plugin set. On any host following this guidance, the model will hunt for nonexistent tools. Suggest updating to whatever image/creative tooling is actually available, or removing the dead routing.
4. **Unpinned MCP server image** — `.mcp.json`'s `maven-tools-mcp` server runs `arvindand/maven-tools-mcp:latest` via `docker run`. An unpinned `:latest` tag means behavior can silently drift between docker pulls with no changelog signal. Suggest pinning to a specific released tag/digest.
[...]

PR #3391, opened 2026-07-08
Audit and optimize agentic config: fix stale memory, add bounded delegation exception

MohabMohie:
## Summary

Full scan of this repo's agentic AI configuration (AGENTS.md/CLAUDE.md, skill bridges, `.mcp.json`, `.claude/settings.json`, `.claude/hooks/guard.py`, and the `.memory/` relational store) plus the machine-wide `~/.claude` config, using a read-only Fable-model research pass (per AGENTS.md Agent Hierarchy's read-only-subagent allowance) followed by main-thread verification and fixes. Findings were triaged per [[bug-handling-policy]]: blockers fixed inline here, non-blocking enhancements consolidated into #3390.

- **Agent Hierarchy policy update** (user-directed): main thread still owns all substantive implementation/integration/final verification, with two new bounded exceptions - a Fable subagent (fallback Opus) for read-only architecture/planning research, and a Haiku subagent for clearly minor, low-risk mechanical edits under main-thread review. This replaces the prior all-or-nothing "read-only research is the sole exception" rule after reconciling a user request that both wanted this pipeline and wanted everything kept on the main thread. Also closed a routing-table gap (`agentic-pdca-loop` wasn't listed).
- **`.memory/` corrections**: two objects were marked stale because they actively contradicted current policy (broad unsupervised delegation; discarding uncommitted work), one fact was marked stale for citing a deleted method, `architecture.current`'s stale "five skills" claim was corrected (repo now has 14), and `start-new-tasks-from-origin-main` got the missing session-reuse caveat. A pre-existing `content_hash` drift on the intellij gradle9/jdk21 gotcha was also repaired.
- **`guard.py`**: fixed a doc comment describing wiring (`matcher`, `command`) that no longer matches `.claude/settings.json`'s actual configuration.
- **`.claude/settings.json`**: added PowerShell allowlist parity for already-approved read-only Bash commands, plus `memory check`/`memory audit`, to reduce unnecessary permission prompts.

Global `~/.claude` settings and this project's auto-memory files (outside this repo, not git-tracked) were reviewed and corrected in the same pass - not part of this diff.

## Test plan

- [x] `py -3 scripts/ci/validate_agent_setup.py` passes (76.82% reduction, above the 76% floor)
- [x] `py -3 .claude/hooks/guard.py --self-test` passes (43/43)
- [x] `memory check` passes with no warnings
- [x] `py -3 -c "import json; json.load(open('.claude/settings.json'))"` confirms valid JSON

Closes nothing directly; non-blocking findings tracked in #3390.


MohabMohie:
Pushed a follow-up commit addressing further requests from the same session:

- **Portability**: `.claude/settings.local.json` is now excluded via the repo's own `.gitignore` (was only excluded via this developer's personal global git config before), and all of its permission entries — including the destructive git wildcards, by explicit user choice — are promoted into the tracked `.claude/settings.json` with PowerShell mirrors.
- **Memory consolidation**: 9 new portable `.memory/` objects migrate durable policy/gotcha content that previously existed only in this developer's machine-local Claude Code memory (not git-tracked, wouldn't survive cloning this repo on another machine): scoped-test-execution (the JVM-crash gotcha), bug-handling policy, worktree-housekeeping policy, single-branch/worktree/PR policy, the memory+graphify mandatory-entry-point rule, the agent-hierarchy decision history, an installer-script staleness gotcha, a `shlex.split()` Windows-path gotcha, and a live-Gemini-E2E test runbook. The 12 corresponding files in this developer's global auto-memory (outside the repo) were reduced to thin pointers to remove the duplicated-policy risk.
- **mvn test enforcement verified, not newly added**: confirmed `.claude/hooks/guard.py`'s R1/R2 already mechanically block (not just document) any qualifying `mvn test`/`verify`/`install`/`deploy`/`package`/`surefire`/`failsafe` invocation missing `-DheadlessExecution=true` or lacking test scoping, and any `allure serve`/`open`/`allure:serve` call — no gap found. Updated one doc-comment cross-reference to cite the newly-portable gotcha instead of the non-portable memory it previously pointed at.

All test-plan checks from the PR description re-verified green after this commit.

Issue #3409, opened 2026-07-09
SHAFT Assistant: followups from the local-agent timeout / recorder fixes


MohabMohie:
Followups noticed while fixing the local-agent-CLI stdin hang (PR to be linked), consolidated into one issue per repo policy.

## 1. `capture_code_blocks` can return a bare "Done" with no generated code

Sending `/codegen recordings/<file>.json` against a real recording can produce a transcript bubble whose entire body is the JSON string `"Done"` (no `codeBlocks`, no fenced Java) even though the tool call reports success.

`shaft-mcp`'s `CaptureService.codeBlocks()` always returns a `McpCaptureReplayResult` record, which always serializes as a JSON object with a `codeBlocks` array (empty at worst) -- never a bare string. So a literal `"Done"` payload means something upstream of that record isn't the path that actually executed, or the MCP transport/tool-call layer is returning something else entirely. Root cause needs a live repro with a real recording session (static reading of `shaft-mcp`/`shaft-capture` didn't turn up a `"Done"` literal anywhere in that path).

As a stopgap, `shaft-intellij`'s `ShaftAssistantPanel.showFinalToolResult` now appends an explicit "no code was generated" note whenever a `capture_code_blocks`/`playwright_recording_code_blocks` call succeeds but the rendered markdown has no code fence, so this no longer looks like a silent/confusing dead end. That doesn't fix the underlying tool-side defect.

## 2. Confirm which trigger orphans the browser after `capture_stop`

`CaptureManager.stop()` was silently returning a stale/unchanged status when the current process had no active recorder, which is now fixed to surface an explicit warning (`"No active recording in this SHAFT MCP session..."`). That makes the failure mode diagnosable, but doesn't explain *why* the browser-owning process and the process handling `capture_stop` diverge in the first place. Three plausible triggers, not yet confirmed against a live repro:

- The recording was started via a natural-language agent-CLI prompt rather than `/record-web`, so it's owned by the agent CLI's own separate `shaft-mcp` subprocess instead of the panel's shared one.
[...]

PR #3420, opened 2026-07-10
SHAFT Assistant: pin engine warm-up version, fix codegen-after-recording dead end, retire '+' button for a filtered slash menu

MohabMohie:
## Problems

1. **Setup screen warm-up command was broken**: `mvn -B dependency:get -Dartifact=io.github.shafthq:SHAFT_ENGINE:LATEST` uses the Maven `LATEST` meta-version (deprecated, snapshot-inclusive, unsupported by `dependency:get` on Maven 4) instead of a real version.
2. **Codegen after a saved recording did nothing**: approving the reviewed Capture code launches a Claude Code run with `--permission-mode acceptEdits` — but `acceptEdits` only auto-approves *file edits*. Every Bash and MCP tool call was auto-DENIED in `--print` mode because the granted path wired no approval bridge. Reproduced live against `claude` v2.1.203: **16 permission denials in a single capture-integration run** (10 Bash, 3 `shaft_coding_partner_plan`, guide search, guardrails — all of which the integration prompt explicitly ordered the agent to call), ~28 turns mostly burned fighting denials. A less persistent run ends as a bare `Done` bubble with no code, no details, and no verbose output — exactly the reported symptom and item 1 of #3409.
3. The `'+'` context button duplicated the typed `/ @ #` triggers, and the slash dropdown showed an unfiltered flat list that inserted doc examples (placeholder paths users then ran verbatim).

## Fixes

- **SetupPrerequisites**: warm-up command now pins the real latest release, resolved asynchronously (never on the EDT) from Maven Central `maven-metadata.xml`, falling back to the plugin's own release version, and only then to Maven's `RELEASE` meta-version for offline dev builds.
- **AssistantLocalAgentRunner**: granted (acceptEdits) Claude AGENT runs now launch the same `--permission-prompt-tool` MCP-over-HTTP bridge as ungranted runs, so shell/MCP calls surface as interactive approval prompts instead of being silently denied. Verified live: with the bridge wired, the identical capture-integration run routes `shaft_guide_search`/Bash through the approval round-trip and completes with files created.
- **Activity footer**: the final Agent answer now always carries a factual `Local agent activity` footer (files created/edited from the CLI's own Write/Edit events; permission denials with per-tool counts from the terminal event) even with Verbose off — a run can never again end as an unexplained bare confirmation.
- **Capture-integration prompt**: helpers are now "try once, continue without" instead of hard requirements; file writes must use file-editing tools; the final answer must list created files and include the generated classes in fenced java blocks, never a bare "Done".
- **Panel**: when the integration run produced no files, the reviewed Capture code blocks stay available for re-approval and the transcript says so explicitly.
- **`/codegen` regenerates**: `overwrite=true` for the deterministic `RecordedFlowTest`, so re-running codegen never fails on an existing class.
- **Composer UX**: removed the `'+'` context button; the slash dropdown live-filters as you type (`/co` → `/codegen`), renders each command's summary as secondary text, and inserts the clean command + trailing space instead of a placeholder example. Placeholder text now advertises `/ @ #`.

## Verification

- Live CLI probes (before/after) against real `claude` v2.1.203 with a stub approval bridge implementing the `LocalAgentApprovalBridge` HTTP contract.
- Live `capture_code_blocks` + `mobile_recording_code_blocks` calls against a running shaft-mcp produce real SHAFT-standard code blocks (the duplicated-locator/`windows.put` Page Object draft bugs seen on the installed 10.3.20260709 build are already fixed at HEAD by #3418 with regression tests, confirmed green).
- Gradle (shaft-intellij): SetupPrerequisitesTest, AssistantLocalAgentRunnerCommandTest, AssistantLocalAgentRunnerVerboseStreamTest, AssistantCommandTest, ShaftPanelSetupTest — 245 tests, 0 failures.
- Scoped Maven (shaft-mcp, headless): McpCaptureCodeBlockServiceTest, MobileRecordingServiceTest — green.

Docs PR follows in shafthq.github.io.

Closes the remaining open item 1 of #3409.


MohabMohie:
Follow-up commit 561028ce56 (user-requested): SHAFT's own MCP tools are now auto-allowed in local Agent runs — `--allowedTools mcp__shaft-mcp` on Claude AGENT commands (validated live: zero denials, no prompt round trip), plus bridge-side auto-approval of `mcp__shaft-mcp__*` as defense in depth. Shell and third-party MCP calls keep the interactive approval bubble. Tests: AssistantLocalAgentRunnerCommandTest + ShaftPanelSetupTest green (162 tests).

PR #3433, opened 2026-07-12
Stream-patch the generated Allure single-file index to fix teardown OOM (#3407)

### LINKED ISSUE #3407: AllureManager.patchGeneratedAllureReportIndex OOMs on large single-file reports
## Summary

`AllureManager.patchGeneratedAllureReportIndex` (`shaft-engine/src/main/java/com/shaft/tools/io/internal/AllureManager.java:733-771`) reads the entire generated Allure single-file `index.html` into a `String` via `Files.readString`, then builds additional full copies of it via `String.substring` concatenation to inject theme/preview-fix markup. Allure's single-file report inlines every attachment (screenshots, etc.) as base64 into that one HTML file, so for suites with many screenshots the file can be large enough that this read-plus-multiple-full-copies pattern exhausts the default surefire-forked-JVM heap.

## Repro (confirmed via CI logs)

Found incidentally while investigating scheduled-run failures on 2026-07-08 (runs [28916954272](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/28916954272) / [28917051358](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/28917051358)). Both `MacOSX_Safari_Local` and `MacOSX_Chrome_Local` jobs in run 28917051358 crashed with an identical stack trace during teardown, **after all tests had already passed** — the job still reported green because the crash happens in the `Post-Test Report and Check` gate's own bookkeeping, not in test execution:

```
java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.Arrays.copyOfRange(Arrays.java:3849)
	at java.base/java.lang.StringLatin1.newString(StringLatin1.java:760)
	at java.base/java.lang.String.substring(String.java:2904)
	at java.base/java.lang.String.substring(String.java:2872)
	at com.shaft.tools.io.internal.AllureManager.patchGeneratedAllureReportIndex(AllureManager.java:758)
	at com.shaft.tools.io.internal.AllureManager.writeAllureReport(AllureManager.java:656)
	at com.shaft.tools.io.internal.AllureManager.openAllureReportAfterExecution(AllureManager.java:370)
	at com.shaft.listeners.internal.ExecutionLifecycleHelper.engineTearDown(ExecutionLifecycleHelper.java:170)
	at com.shaft.listeners.TestNGListener.onExecutionFinish(TestNGListener.java:350)
```

## Suggested next step

Avoid materializing the whole report as one `String` with extra `substring`-based copies — e.g. stream-patch the file (write to a temp file while copying, inserting the head/body patches at the matching markers) instead of `Files.readString` + `substring` + `Files.writeString`, or at minimum only load/rewrite the small head/body regions instead of copying the entire (potentially huge) HTML payload.

## Context

Not caused by any in-flight branch work — confirmed via the run's stack trace on `main` at the runs' `headSha` (`288a95972e9d601f92f2e49d9b0f4163a439fd58`). Not a masked test failure (all tests had already passed by the time this crashes), but a recurring latent defect that will worsen as suites grow more screenshots/attachments.

MohabMohie:
Closes #3407.

## Problem

`AllureManager.patchGeneratedAllureReportIndex` read the entire generated single-file `index.html` into a `String` (`Files.readString`) and then built additional full copies via `String.substring` concatenation to inject the theme/preview-fix markup. Allure's single-file report inlines every attachment as base64, so on suites with many screenshots the report exceeds the forked surefire JVM heap — `java.lang.OutOfMemoryError: Java heap space` during `engineTearDown`, after all tests had already passed (confirmed in scheduled runs 28916954272 / 28917051358).

## Fix

The method now runs in two bounded-memory passes (64 KB buffers), never holding the report in memory:

1. **Scan** (`scanIndexMarkers(InputStream)`): a single sliding-window byte pass locates the three SHAFT patch-id markers plus the first `</head>` and last `</body>` byte offsets. A carried-over window tail guarantees markers split across read boundaries are found regardless of chunk size. All markers are ASCII, so byte-level search is UTF-8-safe and avoids decoding the file.
2. **Write** (`writePatchedIndex`): a sequential copy through a sibling temp file inserts the head/body patches at the recorded offsets, then atomically replaces `index.html` (with a non-atomic fallback).

Semantics are unchanged: head patches go immediately before the first `</head>` (prepended when absent), the theme style immediately before the last `</body>` (appended when absent), and an already-patched report is detected during the scan and **never rewritten** (the patch constants contain no `</head>`/`</body>` literals, so offsets computed on the original file are equivalent to the old post-insertion string arithmetic).

## Regression tests (per the review suggestions on the issue)

- `patchGeneratedAllureReportIndexShouldStreamLargeReportsWithBoundedAllocation` — patches a ~25 MB synthetic single-file report while measuring per-thread allocation via `com.sun.management.ThreadMXBean`; asserts < 8 MB allocated (the old implementation allocated ≥ 3× the file size) and verifies the patched offsets/markers without loading the file as one `String`. Skips gracefully on JVMs without allocation measurement.
- `scanIndexMarkersShouldFindMarkersSpanningAnyReadBoundary` — drives the scanner with an adversarial 3-bytes-per-read stream so every marker straddles input-buffer boundaries; asserts exact first-`</head>`/last-`</body>` offsets.
- `patchGeneratedAllureReportIndexShouldNotRewriteAlreadyPatchedReport` — pins the mtime and asserts an already-patched file is byte-identical and untouched.
- `patchGeneratedAllureReportIndexShouldHandleMissingHeadAndBodyTags` — prepend/append fallback behavior.

## Verification

`mvn -pl shaft-engine test -Dtest=AllureManagerCoverageUnitTest,AllureManagerUnitTest -DheadlessExecution=true` → 61 run, 60 passed, 0 failed, 1 pre-existing environment-dependent skip (`resolveAllureCommandPrefixShouldPreferSystemAllureWhenVersionIsNot2x`, no system allure binary on PATH). Both existing patch-index tests (idempotency, insertion positions, content) pass unchanged.

Internal performance fix with no public-behavior change, so no docs update is needed.

### DIFF
--- a/shaft-engine/src/main/java/com/shaft/tools/io/internal/AllureManager.java
+++ b/shaft-engine/src/main/java/com/shaft/tools/io/internal/AllureManager.java
@@ -12,16 +12,22 @@ import com.shaft.properties.internal.ThreadLocalPropertiesManager;
 import com.shaft.tools.io.ReportManager;
 import org.apache.commons.lang3.SystemUtils;
 
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
@@ -81,6 +87,7 @@ public class AllureManager {
     private static final String ALLURE_ATTACHMENT_PREVIEW_FIX_ID = "shaft-allure-attachment-preview-fix";
     private static final String ALLURE_ATTACHMENT_PREVIEW_SCRIPT_ID = "shaft-allure-attachment-preview-script";
     private static final String ALLURE_THEME_COLORS_ID = "shaft-allure-theme-colors";
+    private static final int INDEX_PATCH_BUFFER_SIZE = 64 * 1024;
     private static final String ALLURE_ATTACHMENT_PREVIEW_FIX_STYLE = """
             <style id="shaft-allure-attachment-preview-fix">
             .shaft-allure-image-modal {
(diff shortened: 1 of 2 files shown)


MohabMohie:
Addressed all 3 Codacy findings in a7f2642cd3: moved `INDEX_PATCH_BUFFER_SIZE` above the constructor (FieldDeclarationsShouldBeAtStartOfClass), replaced the `position` parameter reassignment in `copyIndexBytes` with a local (AvoidReassigningParameters), and removed the new `setAccessible()` reflection in the tests entirely by making `IndexMarkerScan` package-private and deriving the expected offset shift from the file-size delta (AvoidAccessibilityAlteration). Scoped tests re-run green (60 passed, 1 pre-existing env-dependent skip).

Issue #3981, opened 2026-07-21
Maven Central CD: add on-demand reconciliation for partial/interrupted releases


MohabMohie:
## Problem

\`mavenCentral_cd.yml\`'s \`check_release_needed\` job gates the entire delivery pipeline on \`scripts/ci/check_maven_release_version.py\`, which only HEAD-checks whether the **parent POM** (\`io/github/shafthq/SHAFT_ENGINE/<version>/SHAFT_ENGINE-<version>.pom\`) already exists on Maven Central.

If a previous \`mvn deploy\` run partially fails partway through the reactor (e.g. \`shaft-engine\` and the parent POM upload successfully, but a later module like \`shaft-mcp\` or \`shaft-video\` fails on a transient network/signing error), the parent POM is already published. Any later run for that same version -- automatic or manual re-dispatch -- sees \`already_released=true\` from this parent-POM-only check and skips the whole job, so:

- The missing module(s) never get (re-)uploaded to Central.
- \`announce_release\` (GitHub Release creation, docs-repo notification, Slack announcement) never runs for that version, since it's downstream of \`build_release_and_deliver\`.

\`scripts/ci/verify_maven_central_release.py\` already has the right per-artifact check (\`missing_publication_paths\` walks every JAR/POM artifact individually via \`JAR_ARTIFACTS\`/\`POM_ARTIFACTS\`) but it's only used as a post-deploy verification gate inside the same run -- there's no standalone, re-runnable reconciliation path that:
1. Checks every individual release package (not just the engine/parent) against Central for a given version.
2. Uploads only the ones actually missing (targeted \`mvn deploy -pl <missing-modules>\`, since Central rejects re-uploading artifacts that already exist).
3. Then checks whether a GitHub Release exists for that version, creates one if not (reusing the existing \`RELEASE_BODY_TEMPLATE.md\`/\`announce_release\` pattern).
4. Announces to Slack only when it just created that release (not on every run).

## Ask

Add an on-demand (\`workflow_dispatch\`, optionally also scheduled) reconciliation script + workflow that performs the 4 steps above for a given (or the current reactor) version, additive to the existing push-triggered \`mavenCentral_cd.yml\` -- not replacing or weakening its normal gating.

Tracked via ChaosEngine orchestration; implementation in progress.

MohabMohie:
[...]

PR #3983, opened 2026-07-21
Add on-demand Maven Central release reconciliation

MohabMohie:
## Summary

Progresses #3981.

`check_maven_release_version.py`'s `check_release_needed` gate only HEAD-checks the **parent POM** before a real deploy runs. If a reactor `mvn deploy` dies partway through (parent + some modules published, a later module fails transiently), every later run for that SAME version sees `already_released=true` and skips the whole job entirely — the missing module never gets re-uploaded, and `announce_release` (GitHub Release + Slack) never runs for that version since it's downstream.

This adds an **additive, on-demand** reconciliation path — `mavenCentral_cd.yml`'s existing push-triggered flow is untouched.

## What changed

- `scripts/ci/reconcile_maven_central_release.py` (new): reuses `verify_maven_central_release.py`'s per-artifact `missing_publication_paths()` to find exactly which packages are missing from Central for a version, builds (and only executes when not `--dry-run`) a targeted `mvn deploy -pl <missing-modules>` for just those modules, then checks/creates the GitHub Release and posts the Slack announcement only if it just created that release. Idempotent — a clean no-op when nothing is missing and the release already exists.
- `.github/workflows/maven-central-reconcile.yml` (new): `workflow_dispatch`-only (never auto-triggers, so it can't race `mavenCentral_cd.yml`), `version`/`dry_run` inputs, reuses `build_release_and_deliver`'s JDK/Maven/GPG setup steps verbatim.
- `scripts/ci/verify_maven_central_release.py`: fixed artifact-list drift found during audit — `shaft-cli` publishes to Central (`validate_maven_publication.py` already tracked it) but was missing from `JAR_ARTIFACTS`, so a partial `shaft-cli` publish failure would have gone undetected by both the existing release gate and this new tool.
- `tests/scripts/test_verify_maven_central_release.py`: added a permanent drift-guard test cross-checking `JAR_ARTIFACTS`/`POM_ARTIFACTS` against `validate_maven_publication.PUBLIC_ARTIFACTS`, so future drift fails loudly instead of silently.
- `tests/scripts/test_reconcile_maven_central_release.py` (new): 17 tests, every external call (`subprocess.run`, `urllib.request.urlopen`) mocked — no live `mvn deploy`, `gh release create`, or Slack POST is ever made by the test suite.

## Safety note

No live Maven Central deploy, GitHub Release, or Slack post was executed anywhere during development — every external call in the tests is mocked, matching the repo's existing convention for `verify_maven_central_release.py`/`check_maven_release_version.py`. The new workflow only runs on an explicit manual `workflow_dispatch` with real CI secrets, never automatically.

## Test plan

- [x] TDD: `test_artifact_lists_match_the_deployed_reactor_publication_map` watched RED (`shaft-cli` missing) → GREEN after the one-line fix.
- [x] New reconciliation module watched RED (`ModuleNotFoundError`) with the full 17-test file written first → GREEN after implementation.
- [x] `python -m unittest tests.scripts.test_reconcile_maven_central_release tests.scripts.test_verify_maven_central_release tests.scripts.test_check_maven_release_version tests.scripts.test_validate_maven_publication -v` — 41/41 green.
- [x] Full `tests/scripts` suite (`python -m unittest discover -s tests/scripts -p "test_*.py"`) — 328 tests, 1 pre-existing unrelated failure (`test_validate_agent_setup.py` memory-filename-length check on a file from already-merged PR #3980, confirmed via `git show origin/main --stat` predating this branch — not touched by this PR).
- [x] `python scripts/ci/validate_agent_setup.py --skip-external` — same 2 pre-existing errors only, nothing introduced by this change.
- [x] YAML sanity-parsed the new workflow file.

https://claude.ai/code/session_01NDQpYzvnTbDPYwqDwHQoRz

### DIFF
--- a/.github/workflows/maven-central-reconcile.yml
+++ b/.github/workflows/maven-central-reconcile.yml
@@ -1,0 +1,99 @@
+name: Maven Central Release Reconciliation
+# workflow_dispatch-only, on-demand reconciliation for a version whose reactor deploy died
+# partway through the reactor (parent POM + some modules published, a later module failed on a
+# transient error). mavenCentral_cd.yml's check_release_needed guard only HEAD-checks the parent
+# POM, so a later push/manual run for the SAME version would see already_released=true and skip
+# entirely - the missing module(s) never get re-uploaded, and announce_release (GitHub Release +
+# Slack) never runs since it is downstream of build_release_and_deliver.
+#
+# This workflow never triggers automatically, so it can never race or double-fire against
+# mavenCentral_cd.yml's push-triggered flow. It calls scripts/ci/reconcile_maven_central_release.py,
+# which finds exactly which packages are still missing from Central for the given version, deploys
+# only those, and then creates the GitHub Release/Slack announcement if they never ran either.
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to reconcile. Default: read from the reactor via mvn help:evaluate."
+        required: false
+        type: string
+      dry_run:
+        description: "Print the planned actions without deploying/releasing/notifying."
+        required: false
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+
+concurrency:
+  group: maven-central-reconcile
+  cancel-in-progress: false
+
+jobs:
+  reconcile:
+    name: Reconcile Maven Central release
+    # Maven Central + GPG/OSSRH secrets are not available on forked repos; signing would fail with empty key id.
+    if: github.event.repository.fork == false
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+    env:
+      MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+      MAVEN_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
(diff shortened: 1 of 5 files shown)


MohabMohie:
Reviewed, looks solid — correct root-cause diagnosis, thorough mocked test coverage, no live external calls anywhere in the suite, and the workflow defaults to `dry_run: true` so a human has to deliberately opt into a real run. Arming auto-merge.

One non-blocking design note for a future hardening pass: `build_deploy_command` (`scripts/ci/reconcile_maven_central_release.py`) doesn't pass `-am` to the targeted `mvn deploy -pl <missing-modules>`. In the common case (only leaf jar modules missing, their parent/BOM already on Central) this resolves fine from the public repo. But if the *parent POM itself* is also among the missing artifacts in the same reconciliation run, a child module's build won't be able to resolve it without a local build first. Worth adding `-am` (or detecting/ordering the parent-first case) if that scenario ever actually occurs — not filing a separate issue since this is a narrow edge case in a dry-run-by-default, human-supervised recovery tool, not a correctness bug in the common path.

PR #4051, opened 2026-07-26
Fix #4040: propagate throttle-test properties onto the background thread

### LINKED ISSUE #4040: RemoteGridPreflightUnitTest.beforeSessionShouldThrottleConcurrentLocalSessionCreationToMatchingGridCapacity is flaky (unrelated to #1548)
## Summary

`RemoteGridPreflightUnitTest.beforeSessionShouldThrottleConcurrentLocalSessionCreationToMatchingGridCapacity` failed on unrelated PR #4019 (a `shaft-mcp/pom.xml`-only diff — nothing in `shaft-engine` changed), on the first attempt of the `Unit Tests (shaft-engine)` job:

- Run: https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30175905867/attempts/1
- Job: https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30175905867/job/89724536704 (attempt 1, conclusion `failure`)
- Failure:
  ```
  java.lang.AssertionError:
  ❌ Assertion Failed at Line: 155
  🔍 Details: Assertion failed; expected false, but found true
  at com.shaft.driver.internal.DriverFactory.RemoteGridPreflightUnitTest.beforeSessionShouldThrottleConcurrentLocalSessionCreationToMatchingGridCapacity(RemoteGridPreflightUnitTest.java:92)
  ```
  (`Tests run: 1516, Failures: 1` — an isolated single-test flake, not a cascade.)

The job was later re-run (attempt 2) and the branch merged green, so this is an intermittent flake, not a deterministic failure.

## Evidence this is a distinct defect from #1548

Filed while root-causing #1548 ("Safari instance is already paired with another WebDriver session"). That issue named this same test's throttle mechanism (`RemoteGridPreflight`) as a lead. Verified in code that the two are unrelated:

- `RemoteGridPreflight.beforeSession(...)` is called from exactly one call site: `DriverFactoryHelper.createNewRemoteDriverInstance` at `shaft-engine/src/main/java/com/shaft/driver/internal/DriverFactory/DriverFactoryHelper.java:1201`.
- Local driver creation (`createNewLocalDriverInstance`, `DriverFactoryHelper.java:922`, where the Safari "already paired" error actually originates) never calls `RemoteGridPreflight` at all.

So this throttle only guards *remote* Selenium Grid sessions; it has no bearing on local Safari session creation. #1548 was fixed separately (PR TBD) by restoring retry-after-recovery in `createNewLocalDriverInstance`'s Safari branch. This ticket is only about the test's own flakiness.

## Working hypothesis (not yet confirmed against a repro — flagged as hypothesis, not fact)

`RemoteGridPreflightUnitTest.java:79-102` starts a same-JVM `com.sun.net.httpserver.HttpServer` fake grid, acquires a permit on the main thread, then submits a second `RemoteGridPreflight.beforeSession(...)` call to a background thread and immediately asserts (`line 92`) that its `Future` is *not* done yet — i.e. that it is blocked on the capacity-1 `Semaphore` inside `RemoteGridPreflight.acquirePermit` (`RemoteGridPreflight.java:146-161`).

`RemoteGridPreflight.beforeSession` (`RemoteGridPreflight.java:36-47`) only reaches the semaphore at all if `inspect(...)` succeeds in reading the fake grid's `/status` and `/graphql` endpoints and reports `matchingSlots() > 0`. `inspect()` swallows any `Throwable` from the HTTP round-trip and silently falls back to an "unavailable" report (`matchingSlots=0`, logged at `Level.DEBUG` only — `RemoteGridPreflight.java:69-71`), in which case `beforeSession` returns a `SessionPermit.noop()` immediately without ever touching the semaphore.

Two same-JVM callers (main thread + executor thread) hit the same minimal, non-load-tested `HttpServer` test double concurrently. If the second thread's request to that fake server races or is refused, its `inspect()` silently degrades to "unavailable" instead of throwing, `beforeSession` short-circuits to a no-op permit, and the submitted `Future` completes immediately — exactly matching "expected `isDone()==false`, but found `true`" at zero elapsed time (`Time elapsed: 0.049 s`). `RemoteGridPreflightUnitTest.resetForTests()` does clear the static `LIMITERS`/`ATTACHED_SUMMARIES` maps between tests (`RemoteGridPreflight.java:120-123`), so cross-test semaphore leakage was ruled out as the cause.

## Suggested next step for whoever picks this up

Confirm the hypothesis (e.g. temporarily log/assert on `PreflightReport.gridMetadataAvailable()` for the second call under a stress loop), then either make the fake grid server robust to the two concurrent requests the test issues, or replace the `Future.isDone()==false` timing assertion with something that does not race the second thread's own HTTP round trip (e.g. block on a hook confirming the semaphore is actually contended before asserting).

## Scope note

Read-only diagnosis pass. No code changes made here for this ticket; opened per the SHAFT AGENTS.md rule that adjacent findings must be filed, not silently folded into an unrelated PR.

MohabMohie:
## Summary
- `RemoteGridPreflightUnitTest.beforeSessionShouldThrottleConcurrentLocalSessionCreationToMatchingGridCapacity` was intermittently flaky (~2% observed, matches #4040's CI evidence).
- Root cause: the test submits its second `RemoteGridPreflight.beforeSession(...)` call to a pooled background thread, but never sets `SHAFT.Properties.platform` overrides on that thread. `SHAFT.Properties` overrides are explicitly thread-local by design (see `Properties.java`'s own javadoc: "only that thread's config is updated"), so the background thread read the JVM-wide defaults (`remotePreflightEnabled=false`), took the immediate no-op-permit path, and its `Future` occasionally completed before the main thread's `isDone()==false` assertion sampled it.
- Fix: the background task now sets the same properties on its own thread before calling `beforeSession`, mirroring how a real parallel test-execution thread configures itself before creating a session (see `DriverFactoryHelper.java:1201`, where `beforeSession` always runs on the same thread that already configured its own properties).

## Reproduction
- Temporarily added `invocationCount = 50` to the test method and ran it in a single scoped, headless, forked Maven JVM: `mvn -pl shaft-engine -Dtest=RemoteGridPreflightUnitTest#beforeSessionShouldThrottleConcurrentLocalSessionCreationToMatchingGridCapacity -DheadlessExecution=true -Dallure.automaticallyOpen=false test`
- Reproduced twice: 1/50 failure each time, both on the very first invocation, with the same `expected false, but found true` assertion at line 92.
- Instrumented `RemoteGridPreflight.beforeSession` temporarily to print the calling thread's `remotePreflightEnabled()` value; confirmed the pooled background thread (`pool-N-thread-1`) saw `enabled=false` on **every** invocation -- the flake was purely a timing race between how fast that always-no-op call returned and how fast the main thread's assertion sampled the `Future`, not an intermittent exception.
- All instrumentation reverted before committing; only the real fix remains.

## Before / after (same 50-iteration loop, same forked JVM)
- Before fix: 1/50 failures (run 1), reproduced twice.
- After fix: 50/50 passed, then another 50/50 passed (100/100 total).
- Full class (`-Dtest=RemoteGridPreflightUnitTest`, 3 tests) also green.

## Test plan
- [x] `mvn -pl shaft-engine -Dtest=RemoteGridPreflightUnitTest#beforeSessionShouldThrottleConcurrentLocalSessionCreationToMatchingGridCapacity -DheadlessExecution=true -Dallure.automaticallyOpen=false test` -- green (single run)
- [x] Same command with a temporary `invocationCount=50` -- 50/50 and 50/50 green post-fix (reverted before commit)
- [x] `mvn -pl shaft-engine -Dtest=RemoteGridPreflightUnitTest -DheadlessExecution=true -Dallure.automaticallyOpen=false test` -- 3/3 green

Closes #4040

### DIFF
--- a/shaft-engine/src/test/java/com/shaft/driver/internal/DriverFactory/RemoteGridPreflightUnitTest.java
+++ b/shaft-engine/src/test/java/com/shaft/driver/internal/DriverFactory/RemoteGridPreflightUnitTest.java
@@ -87,7 +87,19 @@ public class RemoteGridPreflightUnitTest {
         var executor = Executors.newSingleThreadExecutor();
         try (MockedStatic<ReportManagerHelper> ignored = Mockito.mockStatic(ReportManagerHelper.class);
              var firstPermit = RemoteGridPreflight.beforeSession(gridUrl(server), chromeCapabilities())) {
-            var secondPermit = executor.submit(() -> RemoteGridPreflight.beforeSession(gridUrl(server), chromeCapabilities()));
+            // SHAFT.Properties overrides are thread-local by design (Properties.java), so the
+            // background thread must set its own copy of the same overrides before calling
+            // beforeSession -- exactly as a real parallel test-execution thread would configure
+            // itself. Without this, the pooled thread reads the global defaults
+            // (remotePreflightEnabled=false), short-circuits to a no-op permit, and the Future
+            // can complete before the isDone() check below ever samples it (#4040).
+            var secondPermit = executor.submit(() -> {
+                SHAFT.Properties.platform.set()
+                        .remotePreflightEnabled(true)
+                        .remoteAdaptiveSessionThrottling(true)
+                        .remotePreflightTimeoutSeconds(1);
+                return RemoteGridPreflight.beforeSession(gridUrl(server), chromeCapabilities());
+            });
 
             SHAFT.Validations.assertThat().object(secondPermit.isDone()).isEqualTo(false).perform();
             firstPermit.close();


MohabMohie:
**Learning-loop candidate flagged for the orchestrator** (recorded here so it is not lost; it will be written to `.memory/` in a dedicated pass rather than widening this PR's scope):

> `SHAFT.Properties` overrides are **thread-local by design** (`Properties.java:12-17,59-60`). Any unit test that hands work to a background or pooled thread must re-apply the same overrides on that thread, or it silently reads framework defaults instead of the values the test set.

This is reusable well beyond #4040. The failure it produces is especially deceptive: the background thread takes a different code branch than intended, the test still passes most of the time, and the flake looks like a timing or HTTP problem rather than a configuration one. The hypothesis on the original issue (a swallowed exception in `inspect()`) was ruled out empirically — an instrumented `catch` never fired across 100+ runs.

MohabMohie:
**Independent confirmation of this flake''s reach, from CI triage on an unrelated PR.**

PR #4049 (`drop appium_flutterfinder_java`, a **pom-only** diff — `pom.xml | 1 -`, `shaft-engine/pom.xml | 28 ----`) went red on `Unit Tests (shaft-engine)` with the identical signature this PR fixes:

```
com.shaft.driver.internal.DriverFactory.RemoteGridPreflightUnitTest.beforeSessionShouldThrottleConcurrentLocalSessionCreationToMatchingGridCapacity
java.lang.AssertionError: expected false, but found true
	at RemoteGridPreflightUnitTest.java:92
```

1 failure out of 1521 tests. #4049 never touches `RemoteGridPreflightUnitTest.java` or `RemoteGridPreflight.java`, and was 0 commits behind `origin/main`, so neither the diff nor staleness explains it. Issue #4040 records the same signature on PR #4019 — also a pom-only diff. **Three unrelated pom-only PRs, one isolated failing test.**

#4049 was made green by `gh run rerun --failed` with zero code changes. That is a rerun, not a fix — which is precisely why this PR matters: it removes the race rather than masking it. Merging this closes the loop for every future PR that would otherwise have paid the same toll.

Issue #4006, opened 2026-07-25
E2E Tests: GLOBAL_TESTING_SCOPE lets Playwright*E2ETest classes leak into the native Selenium Grid jobs


MohabMohie:
## Summary

Found while diagnosing nightly issue #3985 (E2E Tests). `GLOBAL_TESTING_SCOPE` in `.github/workflows/e2eTests.yml:24` excludes DB/API/mobile/Cucumber/unit-test classes but never excludes `testPackage.playwright.*PlaywrightActionsE2ETest` classes. Those classes are meant to run only in their dedicated `Ubuntu_Playwright_Chrome/Edge/WebKit/Firefox/GalaxyS26Ultra/IPhone17ProMax` jobs, but the same `-Dtest=${GLOBAL_TESTING_SCOPE}` selector is also used by `Ubuntu_Firefox_Grid`, `Ubuntu_Chrome_Grid`, and `Ubuntu_MicrosoftEdge_Grid` (`e2eTests.yml:224,260,296`), so they run there too, against a Playwright driver that has nothing to do with the remote Selenium grid those jobs stand up.

Two concrete failure modes observed, both from run [30143487701](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30143487701), job `Ubuntu_MicrosoftEdge_Grid` (89641012122):

1. **Wrong fixture host.** These Grid jobs pass `-DexecutionAddress=http://localhost:4444` as a *global* system property for the whole `mvn test` JVM. `TestPageServer.browserHost()` (`shaft-engine/src/test/java/testPackage/TestPageServer.java:126-140`) checks that property process-wide and, seeing `localhost:4444`, returns `host.docker.internal` for every test in that JVM — including Playwright tests, which launch their own local browser process on the bare runner host, not inside a container. `host.docker.internal` isn't resolvable from that process, so:
   ```
   testPackage.playwright.ChromePlaywrightActionsE2ETest.shouldHandleAlerts
   Error { message='net::ERR_NAME_NOT_RESOLVED at http://host.docker.internal:35547/alertFixture.html' ... }
   ```
2. **Missing browser install.** Firefox/WebKit-engine Playwright tests fail outright because these Grid jobs never run the `mvn exec:java ... playwright.CLI install --with-deps <browser>` step that the dedicated `Ubuntu_Playwright_Firefox`/`Ubuntu_Playwright_WebKit` jobs run:
   ```
[...]

PR #4062, opened 2026-07-26
Fix #4021: decouple the panel gate fixture from HealerService's repositoryRoot check

### LINKED ISSUE #4021: Live Diagnostics E2E tests conflict with intentional SHAFT_PROJECT_TOOLS gating
## Context

Follow-up from #3988 (Live Tools Nightly E2E). Two of the three leads investigated there were fixed:
- `AssistantGeminiLiveE2ETest` missing `Files.createDirectories(workspace)` before spawning the MCP process (real bug, fixed).
- `ShaftAssistantPanel.send()` evaluated `ShaftProjectDetector.isShaftProject(project)` before checking whether the invocation even needs it, so the answer got cached (30s TTL) on every dispatch -- including `shaft_project_create` itself, before the project it creates exists. A gated call moments later (e.g. `shaft_project_upgrade`) could see that stale `false` (real bug, fixed by reordering the `&&` operands so `requiresShaftProject` short-circuits first).

That reorder fixes `ShaftAssistantPanelLiveAuthoringToolE2ETest#shaftProjectServiceCreatesAndInitializesAgentsThroughTheRealChatPanel`.

## What remains: not a bug, a design conflict

`ShaftAssistantPanelLiveDiagnosticsToolE2ETest`'s two SHAFT-project-gated methods --
`doctorServiceAnalyzesAFailedAllureResultThroughTheRealChatPanel` (`doctor_analyze_failed_allure`) and
`healerServiceRunsAgainstANonShaftDirectoryThroughTheRealChatPanel` (`healer_run_failed_test`,
`verify_run_focused`) -- still fail with the "doesn't look like a SHAFT project yet" local nudge,
because their workspace (`build/live-tools/diagnostics`) genuinely never becomes a SHAFT project (only
an `allure-results/*.json` fixture is written into it).

This is NOT the caching bug above -- `ShaftProjectDetector.isShaftProject` correctly and freshly
returns `false` for that directory every time. It also is NOT a case of the client-side gate being
wrong: `AssistantCommandRoutingTest#requiresShaftProjectGatesOnlyMutatingOrShaftReportingSpecificTools`
(shaft-intellij/src/test/java/com/shaft/intellij/ui/AssistantCommandRoutingTest.java:822-825) already
asserts, by design, that `verify_run_focused` and `doctor_analyze_failed_allure` "must be gated" /
"only make sense against SHAFT Allure evidence". `AssistantCommand.SHAFT_PROJECT_TOOLS`'s membership
is deliberate (confirmed by a prior investigation cited in #3988's dispatch).

So the two Diagnostics live-E2E tests' own premise -- that these tools should reach the MCP tool layer
and return the tool's own graceful non-SHAFT-project handling (e.g. `healer_run_failed_test` ->
`GUARDRAIL_STOPPED`) -- conflicts with the already-tested, intentional client-side gate that blocks
them first. Whichever side is "right" is a product decision, not a bug fix:

1. Give the Diagnostics test workspace a real (minimal) SHAFT project marker before calling these
   tools, mirroring how `ShaftAssistantPanelLiveAuthoringToolE2ETest` already does via
   `shaft_project_create` -- but this changes what `healerServiceRunsAgainstANonShaftDirectoryThroughTheRealChatPanel`
   actually tests (its name and doc explicitly want non-SHAFT-directory tool behavior).
2. Narrow `SHAFT_PROJECT_TOOLS` to exclude these specific diagnostics tools -- but that contradicts the
   existing, tested rationale in `AssistantCommandRoutingTest`.
3. Split the diagnostics test's "graceful non-SHAFT handling" assertions into a lower-level test that
   calls the tool directly (bypassing the panel gate), and keep the live-panel test scoped to a real
   SHAFT project.

## Evidence
- Nightly run 30146701546, job "IntelliJ live tools (Diagnostics)": both methods fail with
  `org.opentest4j.AssertionFailedError: Assistant panel never settled ... within PT1M`, transcript
  showing the "doesn't look like a SHAFT project yet" nudge for both `doctor_analyze_failed_allure`
  and `healer_run_failed_test`/`verify_run_focused`.
- `shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantCommand.java:1590-1604` (`SHAFT_PROJECT_TOOLS`
  set + rationale comment).
- `shaft-intellij/src/test/java/com/shaft/intellij/ui/AssistantCommandRoutingTest.java:809-835`.
- `shaft-intellij/src/test/java/com/shaft/intellij/ui/ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java:74-114`
  (class/method docs stating the non-SHAFT-directory intent).

## Ask
Product decision needed on which of the two intentional designs should change (or whether option 3's
split is the right shape) before these two live E2E tests can pass without either weakening the
gate's tested guarantees or changing what the tests claim to verify.

MohabMohie:
## Summary

`ShaftAssistantPanelLiveDiagnosticsToolE2ETest` has two methods that dispatch SHAFT-project-gated tools (`doctor_analyze_failed_allure`, `healer_run_failed_test`, `verify_run_focused`) through the real chat panel. They've been failing in every scheduled run of the Diagnostics job because two *independent* checks happened to collide:

- The client-side panel gate, `ShaftProjectDetector.isShaftProject(project)` (`shaft-intellij/src/main/java/com/shaft/intellij/project/ShaftProjectDetector.java:44-72`) — keyed off the fake IntelliJ project's base path.
- The tool-side non-SHAFT check, `HealerService.runFailedTest` → `ShaftProjectMarker.isShaftRepository(repository)` (`shaft-mcp/src/main/java/com/shaft/mcp/HealerService.java:135`) — keyed off the `repositoryRoot` argument in the tool-call JSON.

They only collided because the test passed `"repositoryRoot":"."`, i.e. the same directory as the panel's project root. Neither check itself is wrong — `AssistantCommand.SHAFT_PROJECT_TOOLS` (`AssistantCommand.java:1598-1604`) and `AssistantCommandRoutingTest#requiresShaftProjectGatesOnlyMutatingOrShaftReportingSpecificTools` (`AssistantCommandRoutingTest.java:809-834`) assert, by design, that these tools must be gated. This PR does not touch either.

## What changed

Only `ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java`:
- New helper `markWorkspaceAsShaftProject(Path)` plants a minimal marker `pom.xml` at the workspace root so the panel actually dispatches the gated calls instead of showing the onboarding nudge.
- `healerServiceRunsAgainstANonShaftDirectoryThroughTheRealChatPanel` additionally points `repositoryRoot` at a fresh, marker-free `non-shaft-target/` subdirectory for both `healer_run_failed_test` and `verify_run_focused`, so `HealerService`'s own non-SHAFT-directory `GUARDRAIL_STOPPED` branch is still genuinely exercised. "Through the real chat panel" and "against a non-SHAFT directory" are about two different roots, not a contradiction — this fixture makes that explicit instead of conflating them.

## Verdict on the design question in #4021

Neither of the issue's two framed options fit cleanly:
- Not "wrong CI job, exclude it" (the #4006/#4038 pattern): `intellij-tools-diagnostics` in `live-tools-nightly.yml` is the sole, purpose-built job for this class — there's no other job to move it to.
- Not "the gate is over-broad, narrow it": the gate is intentional and already asserted by its own unit test, which stays untouched and green (46/46, see below).

This is a test-fixture fix: the fixture now gives the (unchanged) gate what it needs while keeping each tool's own graceful-non-SHAFT-directory behavior genuinely under test.

## Evidence

**Red nightly**, not a silent skip or a wasted job — 5/5 consecutive scheduled runs of `Live Tools Nightly E2E` → `IntelliJ live tools (Diagnostics)` have failed:
```
ShaftAssistantPanelLiveDiagnosticsToolE2ETest > healerServiceRunsAgainstANonShaftDirectoryThroughTheRealChatPanel() FAILED
    org.opentest4j.AssertionFailedError at ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java:95
ShaftAssistantPanelLiveDiagnosticsToolE2ETest > doctorServiceAnalyzesAFailedAllureResultThroughTheRealChatPanel() FAILED
    org.opentest4j.AssertionFailedError at ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java:49
5 tests completed, 2 failed
```
(run 30190833709, job 89763701272). Tracked by the auto-filed nightly tracker #4047.

**Gate unchanged and still enforced**: `AssistantCommandRoutingTest` — 46 tests, 0 failures, 0 errors (`gradle test --tests com.shaft.intellij.ui.AssistantCommandRoutingTest`), run locally against this branch.

## Verification status — please read before merging

This test only runs in the `Live Tools Nightly E2E` scheduled workflow (`intellij-tools-diagnostics` job) — none of the PR's ordinary required checks exercise it. **Auto-merge is intentionally NOT armed on this PR.** A local watched RED→GREEN cycle against a live MCP server was attempted but not completed (IntelliJ platform test fixtures bootstrap slowly enough that CI itself budgets 45 minutes for this job). Instead, this PR should be verified by dispatching `live-tools-nightly.yml` against this branch via `workflow_dispatch` before merge — that is the authoritative proof for a fix that only this workflow can exercise.

Closes #4021

### DIFF
--- a/shaft-intellij/src/test/java/com/shaft/intellij/ui/ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java
+++ b/shaft-intellij/src/test/java/com/shaft/intellij/ui/ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java
@@ -8,6 +8,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Duration;
+import java.util.Locale;
 
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -102,15 +103,20 @@ class ShaftAssistantPanelLiveDiagnosticsToolE2ETest {
         try (LiveChatToolE2ESupport support = LiveChatToolE2ESupport.install(context.workspace(), context.mcpCommand())) {
             ShaftAssistantPanel panel = support.newPanel();
 
-            // "mvn.cmd" (not bare "mvn") on Windows: HealerService/verify_run_focused spawn the
-            // command via a raw ProcessBuilder (HealerService.java:450), which calls CreateProcess
-            // directly -- unlike a shell, Windows CreateProcess never appends PATHEXT extensions, so
-            // the extensionless "mvn" POSIX shim on PATH cannot be launched that way even though
-            // `where mvn` resolves it (reproduced empirically: intermittent "MCP healer command could
-            // not be launched" IOException). "mvn.cmd"/"mvnw.cmd" are explicitly allowlisted
-            // executables (HealerService.java:349) precisely for this platform difference.
+            // HealerService/verify_run_focused spawn the command via a raw ProcessBuilder
+            // (HealerService.java:450), which calls CreateProcess directly on Windows -- unlike a
+            // shell, Windows CreateProcess never appends PATHEXT extensions, so the extensionless
+            // "mvn" POSIX shim on PATH cannot be launched that way even though `where mvn` resolves it
+            // (reproduced empirically: intermittent "MCP healer command could not be launched"
+            // IOException). "mvn.cmd" is required there. But this test only ever runs in CI on the
+            // ubuntu-22.04 "intellij-tools-diagnostics" job (live-tools-nightly.yml), which has no
+            // "mvn.cmd" file at all -- only the "mvn" POSIX shell script -- so hardcoding "mvn.cmd"
+            // fails process launch outright on Linux (issue #4021). HealerService allowlists both
+            // names (HealerService.java:349); pick the one that actually exists for the current OS.
+            String mavenExecutable = mavenExecutableForOs();
             String healResponse = support.send(panel,
-                    "/mcp healer_run_failed_test {\"repositoryRoot\":\"non-shaft-target\",\"testCommand\":[\"mvn.cmd\",\"test\"],"
+                    "/mcp healer_run_failed_test {\"repositoryRoot\":\"non-shaft-target\",\"testCommand\":[\""
+                            + mavenExecutable + "\",\"test\"],"
                             + "\"outputDirectory\":\"\",\"maxAttempts\":1,\"includeScreenshots\":false,"
                             + "\"includePageSnapshots\":false,\"allowedSourcePaths\":[],"
                             + "\"networkValidationApproved\":false,\"useConfiguredAi\":false,\"allowLocalAi\":false,"
@@ -121,7 +127,8 @@ class ShaftAssistantPanelLiveDiagnosticsToolE2ETest {
                     "healer_run_failed_test: Maven command failed to launch: " + healResponse);
 
             String verifyResponse = support.send(panel,
-                    "/mcp verify_run_focused {\"repositoryRoot\":\"non-shaft-target\",\"command\":[\"mvn.cmd\",\"-q\",\"compile\"],"
+                    "/mcp verify_run_focused {\"repositoryRoot\":\"non-shaft-target\",\"command\":[\""
+                            + mavenExecutable + "\",\"-q\",\"compile\"],"
                             + "\"networkValidationApproved\":false}",
                     Duration.ofSeconds(60));
             assertNotError(verifyResponse, "verify_run_focused");
@@ -204,6 +211,19 @@ class ShaftAssistantPanelLiveDiagnosticsToolE2ETest {
             assertNotError(response, "autobot_provider_status");
             assertTrue(payload.contains("\"provider\":\"anthropic\""), "Expected the requested provider echoed back: " + payload);
         }
+    }
+
+    /**
+     * The Maven executable name valid for {@link ProcessBuilder} on the current OS: {@code "mvn.cmd"}
+     * on Windows (bare {@code "mvn"} cannot be launched there -- CreateProcess never appends PATHEXT
+     * extensions), {@code "mvn"} everywhere else, including this test's actual CI runner
+     * ({@code ubuntu-22.04}, which has no {@code mvn.cmd} file at all). Both names are allowlisted by
+     * {@code HealerService} (HealerService.java:349).
+     */
+    private static String mavenExecutableForOs() {
+        return System.getProperty("os.name", "").toLowerCase(Locale.ROOT).contains("win")
+                ? "mvn.cmd"
+                : "mvn";
     }
 
     /**
--- a/shaft-intellij/src/test/java/com/shaft/intellij/ui/ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java
+++ b/shaft-intellij/src/test/java/com/shaft/intellij/ui/ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java
@@ -39,6 +39,7 @@ class ShaftAssistantPanelLiveDiagnosticsToolE2ETest {
     @Timeout(120)
     void doctorServiceAnalyzesAFailedAllureResultThroughTheRealChatPanel() throws Exception {
         LiveContext context = LiveContext.assumeConfigured();
+        markWorkspaceAsShaftProject(context.workspace());
         Path allureResult = context.workspace().resolve("allure-results/failed-login-result.json");
         Files.createDirectories(allureResult.getParent());
         Files.writeString(allureResult, failedAllureResult(), StandardCharsets.UTF_8);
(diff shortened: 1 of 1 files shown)


MohabMohie:
## Follow-up: healer case fixed

Verified the dispatched run (30192641029) evidence directly. The doctor test now passes; `healerServiceRunsAgainstANonShaftDirectoryThroughTheRealChatPanel` still failed, moved from `:95` to `:128`:

```
org.opentest4j.AssertionFailedError: verify_run_focused: Maven command failed to launch: MCP healer command could not be launched. ==> expected: <false> but was: <true>
    at ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java:128
```

**Root cause**: `verify_run_focused` (`HealerService.verifyFocused`, `HealerService.java:305-328`) never gates on `ShaftProjectMarker.isShaftRepository` — unlike `healer_run_failed_test`, which short-circuits to `GUARDRAIL_STOPPED` before ever spawning a process when `expectsAllure` is true. `verify_run_focused` always launches the tokenized Maven command via `ProcessBuilder`. The test hardcoded `"mvn.cmd"` unconditionally (reasoning only about the Windows CreateProcess/PATHEXT quirk), but this test's actual CI job (`intellij-tools-diagnostics`, `live-tools-nightly.yml:116`, `runs-on: ubuntu-22.04`) is Linux, which has no `mvn.cmd` file at all — only `mvn`. `HealerService.java:349` already allowlists both `mvn` and `mvn.cmd`; the test just needed to pick the one that exists for the current OS.

Fixed in `ShaftAssistantPanelLiveDiagnosticsToolE2ETest.java` only (new `mavenExecutableForOs()` helper) — no production code touched.

**Second finding, triaged**: the same dispatched run also showed `IntelliJ live tools (Tool + Capture)` (job 89768534243) red — `ShaftAssistantPanelLiveCaptureToolE2ETest.captureServiceRunsAnApiRecordingChainThroughTheRealChatPanel()` at `:89`. Confirmed **pre-existing on `main`**: the baseline run (30190833709, job 89763701294) fails identically, same test, same line. Not a regression from this branch, not touched (`shaft-capture/**` is out of scope here). Flagging for a separate ticket.

Branch rebased onto latest `origin/main` (3 commits, no conflicts). `AssistantCommandRoutingTest` still green locally (46/46, 0 failures/errors). Re-dispatched `Live Tools Nightly E2E` against this branch: https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30193206066

Auto-merge still intentionally OFF pending that run.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

MohabMohie:
**Verified on real CI — arming auto-merge.**

Re-dispatched run https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30193206066 on this branch:

```
IntelliJ live tools (Diagnostics): success   <- this PR''s target
IntelliJ live tools (Authoring):   success
shaft-cli live E2E (Maven):        success
IntelliJ live tools (Tool + Capture): failure
IntelliJ live Gemini E2E:          failure
```

**The Diagnostics job is green.** Baseline on `main` (run 30190833709) had 2 failures in that job; the first dispatched run on this branch had 1; this run has 0. The `mvn.cmd`-on-`ubuntu-22.04` fix closed the last one.

The two remaining red jobs are **not caused by this branch**, verified rather than assumed:

- `Tool + Capture` — identical failure (`ShaftAssistantPanelLiveCaptureToolE2ETest:89`) in the `main` baseline run. Tracked as **#4070**, with a live hypothesis that it is the #4046 CDP hang reached by another entry point.
- `Gemini E2E` — red on `main` in **4 of the last 4 scheduled runs** (07-23, 07-24, 07-25, 07-26), and it *passed* once on this very branch (run 30192641029), so it is intermittent and pre-existing. Tracked separately.

Blocking this fix on unrelated failing jobs would leave a proven Diagnostics regression on `main` for no benefit. Merging.

MohabMohie:
**Recording a first-hand observation from this PR''s work so it becomes citable.**

A Learning Loop pass just tried to record the gotcha below and correctly **refused**, because it could find no evidence anywhere in the issue, this PR, or a repo-wide search. That is the right behavior — but the observation is genuine. It was reported directly by the agent that did this work, and agent-to-orchestrator reports are ephemeral: nothing durable existed for a later agent to verify against. Parking it here fixes that.

**Gotcha — abandoned Gradle Test Executor JVMs mmap-lock the shaft-intellij plugin sandbox.**

While attempting a local `shaft-intellij` test run, `prepareTestSandbox` failed with a Windows-only "user-mapped section open" error. Cause: a leftover `Gradle Test Executor` JVM (PID 26528, its command line pointing at `...diagnostics-red`) from an earlier aborted watched-RED attempt **in the same worktree** still held `gson-2.14.0.jar` memory-mapped. The failure does not name the real cause and reads like a corrupt sandbox.

Diagnosis and recovery:
```powershell
Get-CimInstance Win32_Process -Filter "Name='java.exe'"   # match CommandLine against your own worktree path
Stop-Process -Force -Id <pid>
```

**Only kill processes whose command line points at your own worktree.** Long-lived `GradleDaemon` processes belong to other concurrent agents and must be left alone — this session had up to four agents active simultaneously.

Related to the wider pattern already recorded from this session: hangs and aborted runs leak process trees (a surefire JVM plus chromedriver and 8-9 `chrome.exe` children in the #4046 reproduction), and JUnit `@Timeout` does not reliably interrupt them.

**Process lesson worth more than the gotcha itself:** durable facts need a durable artifact at the moment of discovery. A finding that lives only in a subagent''s final report to the orchestrator cannot be verified later, and a verify-first Learning Loop will correctly reject it.

Issue #4014, opened 2026-07-25
[Bug] [Capture] Two incompatible NetworkCaptureOptions types share one name, with an unguarded polarity inversion between them


MohabMohie:
Discovered while fixing #3986. Filed separately rather than folded in, because #3986's fix is a two-line schema correction and this is a data-model hazard with its own blast radius.

## Two different `NetworkCaptureOptions` types, one module

`shaft-capture` ships two unrelated types with the identical simple name:

| | `com.shaft.capture.runtime.NetworkCaptureOptions` | `com.shaft.capture.network.NetworkCaptureOptions` |
|---|---|---|
| Shape | mutable class, public fields | `record` |
| Faces | MCP / tool-call JSON binding | `CaptureNetworkRecorder` internals |
| Fields | `enabled`, `excludeAssets`, `excludePattern`, `includePattern`, `captureResponseBodies`, `captureRequestBodies` | `includeAssetTypes`, `urlIncludeGlobs`, `urlExcludeGlobs`, `firstPartyOnly`, `maxTransactions`, `maxBodyBytes` |

Neither is redundant — they model genuinely different things. The problem is that they are **named identically while being semantically incompatible**, and one is translated into the other by hand.

## Why this is a bug farm, not just a naming annoyance

**1. Opposite polarity on the asset flag.** `runtime` has `excludeAssets` (true = drop assets); `network` has `includeAssetTypes` (true = keep assets). Same concept, inverted boolean, translated by hand at `shaft-capture/src/main/java/com/shaft/capture/runtime/ManagedCaptureRecorder.java:1033` (`translateNetworkCaptureOptions`). A polarity slip there silently records the exact opposite of what the caller asked for — no exception, no failed test, just wrong captured data.

**2. Cardinality mismatch on the glob filters.** `runtime` carries a single `String excludePattern` / `includePattern`; `network` carries `List<String> urlExcludeGlobs` / `urlIncludeGlobs`. The translation collapses or expands between them.

**3. Any `import` of the wrong one still compiles** wherever the other's members aren't referenced, and IDE auto-import will offer both with no hint which is correct.

## Related brittleness this exposes

Several call sites hardcode `runtime.NetworkCaptureOptions`'s field names as **strings**, with comments explicitly warning they must be kept in sync by hand:

[...]

PR #4064, opened 2026-07-26
Fix #4057: wire enabled/captureRequestBodies/captureResponseBodies through network capture

### LINKED ISSUE #4057: translateNetworkCaptureOptions silently drops enabled / captureRequestBodies / captureResponseBodies (unverified — carved out of #4014)
## Context

#4014 covered two concerns. PR #4056 fixes and pins the first — the unguarded polarity inversion between the two same-named `NetworkCaptureOptions` types (`runtime.excludeAssets` ↔ `network.includeAssetTypes`), now covered by a mutation-checked regression test.

**The second concern is not addressed and would be lost when #4014 closes**, so it is carved out here.

## Problem

`ManagedCaptureRecorder.translateNetworkCaptureOptions` (`shaft-capture/src/main/java/com/shaft/capture/runtime/ManagedCaptureRecorder.java:1033-1041`) converts `com.shaft.capture.runtime.NetworkCaptureOptions` (the mutable, Spring-AI-bound MCP request type) into `com.shaft.capture.network.NetworkCaptureOptions` (the normalized record used by `CaptureNetworkRecorder`).

Three fields present on the source type have **no equivalent on the target record and are silently discarded**:

- `enabled`
- `captureRequestBodies`
- `captureResponseBodies`

## Status: unverified

The agent that fixed the polarity half assessed this as *probably* intentional — the record has no such concept, and `resolvedNetworkCaptureOptions()` (`ManagedCaptureRecorder.java:1001-1005`) is only reached when capture is already enabled, which would make `enabled` redundant at that point. **It explicitly did not verify this and did not touch it**, correctly staying inside its ticket''s acceptance criteria.

So the open question is narrow and answerable:

1. Is dropping `enabled` genuinely redundant at this call site, or can a caller reach it with `enabled=false` and be silently ignored?
2. Do `captureRequestBodies` / `captureResponseBodies` have any effect today when set via MCP `networkOptions`, or are they accepted from the caller and discarded? If discarded, that is an MCP contract defect — the tool schema advertises parameters that do nothing.

Point 2 is the one that matters: a tool parameter a caller can set, that is accepted without error and then ignored, is worse than one that does not exist. Note PR #4050 has just annotated these request fields with `@ToolParam` descriptions, so they are now *documented* to callers as well as accepted.

## Suggested approach

Trace each of the three fields to their consumers (or prove there are none), then either wire them through, or reject/warn explicitly at the boundary, or document them as ignored-by-design. Add a test pinning whichever answer is correct, so the next reader does not have to re-derive it.

Ref: #4014, PR #4056, PR #4050.

MohabMohie:
## Summary

Carved out of #4014 (fixed by #4056, which pinned the `excludeAssets`<->`includeAssetTypes` polarity inversion). This ticket covers the second, previously-unverified concern: `ManagedCaptureRecorder.translateNetworkCaptureOptions` silently dropped three fields of the MCP-bound `com.shaft.capture.runtime.NetworkCaptureOptions` when converting into the internal `com.shaft.capture.network.NetworkCaptureOptions` record, which has no equivalent for them.

## Trace (per field, with verdict)

**1. `enabled`** -- live contract defect (a), not the "probably redundant" theory from #4014's carve-out.
- `ManagedCaptureRecorder.startNetworkRecorder` (`ManagedCaptureRecorder.java:969-970`) gates recorder creation on `apiCaptureEnabled()`, which checks the *top-level* `CaptureStartOptions.apiCapture` flag or the global `SHAFT.Properties.capture.enabled()` property (`ManagedCaptureRecorder.java:990-998`).
- `networkOptions.enabled` was never consulted anywhere in that decision. A caller can set `apiCapture=true` (or rely on the global property) **and** `networkOptions={enabled:false}` in the same request -- reachable via both `capture_api_start` (which always force-sets `apiCapture=true`, `CaptureService.java:558`) and `capture_start`'s `codegenOptions.networkOptions` -- and the `false` was silently ignored; capture proceeded anyway.

**2. `captureRequestBodies` / `captureResponseBodies`** -- live contract defect (a), the one that matters per the issue.
- `CaptureNetworkRecorder.recordIfAccepted` (`CaptureNetworkRecorder.java:259-263`, pre-fix) called `NetworkBodyStore.store` for both request and response bodies **unconditionally**, subject only to `maxBodyBytes` truncation. The target record had no field to gate this at all.
- This is not hypothetical: `shaft-intellij`'s `RecordApiWebAction.java:60-64` and `GuidedWorkflowPanel.java:795-796` already send both flags through `capture_api_start`/`capture_start`, and `RecordApiWebAction.java`'s own comment describes body capture as "the entire point of API recording" that these flags were believed to control. They did nothing.
- PR #4050 additionally documented `networkOptions` at the MCP layer via `@ToolParam` (`CaptureService.java:317-318`), and the recorded example call in `shaft-mcp`'s tool-index fixtures already shows real callers setting `captureRequestBodies`/`captureResponseBodies` explicitly.

No field was in state (b) genuinely-redundant or (c) partially-wired -- all three are (a).

## Fix

- `com.shaft.capture.network.NetworkCaptureOptions` gains `captureRequestBodies`/`captureResponseBodies` record components, plus a 6-arg compatibility constructor defaulting both to `true` so every existing internal call site (tests, `defaults()`, the config-driven branch with no matching `SHAFT.Properties.capture` field) keeps today's unconditional-capture behavior unchanged.
- `CaptureNetworkRecorder.recordIfAccepted` only persists a body when the matching flag is `true`; otherwise the transaction is still recorded (method/URL/headers/status) with a `null` `BodyRef` -- the same sentinel already used for "no body."
- `ManagedCaptureRecorder.translateNetworkCaptureOptions` carries both flags through instead of dropping them.
- `ManagedCaptureRecorder.startNetworkRecorder` gates on a new package-private `networkCaptureExplicitlyDisabled(NetworkCaptureOptions)` before constructing a `CaptureNetworkRecorder`, so `enabled=false` is honored as an explicit opt-out independent of `apiCapture`/the global property.

`shaft-mcp` is untouched: it only binds `com.shaft.capture.runtime.NetworkCaptureOptions` (unchanged) and never references the internal `network.NetworkCaptureOptions` record, so its tool schema is unaffected.

## Test plan

- [x] RED first: new tests referencing the not-yet-existing record fields/method failed to compile, then failed for the expected reason once the API existed but wasn't wired.
- [x] `mvn -pl shaft-capture test -Dtest=CaptureNetworkRecorderTest,ManagedCaptureRecorderControlTest -DheadlessExecution=true -Dallure.automaticallyOpen=false` -- 29/29 passed.
- [x] Mutation-checked each new guard by reverting it in isolation and confirming the matching test fails, then restoring it:
  - `networkCaptureExplicitlyDisabled` forced to `false` -> `networkCaptureExplicitlyDisabledIsTrueOnlyWhenEnabledIsExplicitlyFalse` failed.
  - `translateNetworkCaptureOptions`'s two new args hardcoded to `false` -> both `translateNetworkCaptureOptionsCarriesCapture*BodiesThrough` tests failed.
  - Request-body gate removed -> both new `CaptureNetworkRecorderTest` tests failed.
  - Response-body gate removed -> both new `CaptureNetworkRecorderTest` tests failed.
- [x] `mvn -pl shaft-capture compile test-compile -DheadlessExecution=true -Dallure.automaticallyOpen=false` -- BUILD SUCCESS.

Closes #4057
Ref #4014, #4056, #4050


MohabMohie:
**Parking the memory content here** so it is not lost — the agent correctly declined to write `.memory/**` because that path was fenced to another agent at the time. That fence has since lifted; this goes into the store in a Learning Loop pass rather than widening this PR.

Two durable items from this work:

1. **Gotcha:** two distinct types named `NetworkCaptureOptions` exist — `com.shaft.capture.runtime.NetworkCaptureOptions` (mutable, Spring-AI-bound from MCP tool-call JSON) and `com.shaft.capture.network.NetworkCaptureOptions` (normalized record, internal to `CaptureNetworkRecorder`). They are bridged by a **hand-written translation** (`ManagedCaptureRecorder.translateNetworkCaptureOptions`) with **no compiler guard**: fields present on the source and absent on the target vanish silently, and a polarity flip between `excludeAssets` and `includeAssetTypes` is a bare `!`. Two separate defects have now come out of this one seam — #4014 (unguarded polarity inversion, pinned by PR #4056) and #4057 (three fields silently dropped, fixed here. Any future field added to either type must be checked against the translation by hand.

2. **Decision/finding:** the "probably redundant" reading of the dropped fields was **wrong**. All three were live contract defects. `enabled` was never consulted — `startNetworkRecorder` gated only on the top-level `apiCapture` flag or the global property (`ManagedCaptureRecorder.java:969-970,990-998`). `captureRequestBodies`/`captureResponseBodies` were ignored while `CaptureNetworkRecorder.recordIfAccepted` (`:259-263`) stored bodies unconditionally — and `shaft-intellij` was already sending both flags in the belief that they worked.

The general lesson worth keeping: *a field that is accepted, documented to callers, and then discarded is worse than one that does not exist* — the caller gets no signal. PR #4050 had just annotated these with `@ToolParam` descriptions, which made the gap actively misleading rather than merely latent.

PR #4068, opened 2026-07-26
Test: reproduce #4046 Guided Workflows Live E2E hang (disabled, no fix)

MohabMohie:
## Summary

Adds a **disabled** regression test that reproduces, on demand, the hang behind #4046 (nightly `Guided Workflows Live E2E`, 5/5 red since PR #3938 merged). This PR does **not** fix the root cause and does **not** close #4046 — see "Refs #4046" in the commit, not "Closes".

## What was found (full diagnosis on #4046)

- `GuidedWorkflowLiveE2ETest#apiRecorderCapturesNetworkTrafficAndGeneratesShaftApiCode` times out waiting for `capture_api_start` to respond (`java.io.IOException: Timed out waiting for SHAFT MCP response`).
- Reproduced locally (not just from CI logs): `ManagedCaptureRecorderBrowserTest#startCompletesWhenApiCaptureIsEnabledAgainstARealBrowser` drives the same combination — real headless Chrome + `apiCapture=true` — through `ManagedCaptureRecorder` directly, and hangs identically.
- `jstack` on the hung JVM pins the blocked thread inside `RemoteWebDriver.executeScript` → `JdkHttpClient.execute` → `CompletableFuture.get()`, called from `ManagedCaptureRecorder.start(ManagedCaptureRecorder.java:151)` — the JS event-listener injection call made right after `driver.navigate()`, one line after the CDP `NetworkInterceptor` is installed (`startNetworkRecorder`, `ManagedCaptureRecorder.java:148`).
- **Isolation experiment**: disabled the WebDriver BiDi collector (`startCollector`) and re-ran — hang persisted unchanged. So this is **not** a BiDi+CDP concurrency conflict; CDP `NetworkInterceptor` alone poisons a subsequent `executeScript` call. No existing test in the repo previously combined a real browser with `apiCapture=true` before PR #3938's E2E test, which is why this was never caught pre-merge.
- Also found (informational, not fixed here): the test's own `@Timeout(60)` did not reliably abort the hang locally — it kept running 2+ minutes past the bound in two separate runs until I killed the process tree manually.

## Why `@Disabled` and not just relying on the existing `@Tag("external-e2e")`

[...]

Issue #4066, opened 2026-07-26
[Bug] [Capture] CaptureManager.invoke() waits on executor.submit(...).get() with no timeout, so any recorder hang becomes an unbounded MCP tool hang


MohabMohie:
## Problem

`CaptureManager.invoke()` (`shaft-capture/src/main/java/com/shaft/capture/runtime/CaptureManager.java:386-399`) calls `executor.submit(operation).get()` **with no timeout**. If the submitted operation blocks forever, the MCP tool call blocks forever with it — there is no upper bound and no diagnosable failure, just silence.

## How it surfaced

Found while root-causing #4046. A real hang in `ManagedCaptureRecorder.start()` (blocked on an `executeScript` round-trip that chromedriver never answers) propagated straight through this unbounded `get()`, so `capture_api_start` never returned. The MCP client eventually gave up on its own 240s `SESSION_START_TIMEOUT`:

```
java.io.IOException: Timed out waiting for SHAFT MCP response.
	at com.shaft.intellij.mcp.ShaftMcpStdioClient.requestTimedOut(ShaftMcpStdioClient.java:389)
```

The client-side timeout is the only thing that ended it, and it ends the *client''s* wait — not the server-side work, the forked JVM, or the leaked browser process.

## Why this is worth fixing independently of #4046

#4046''s root cause is a specific BiDi/CDP interaction. **This is the amplifier**, and it will amplify the next one too: any future recorder-side hang — a driver that stops responding, a network stall, a wedged CDP session — becomes an unbounded, undiagnosable server hang rather than a bounded, reported error.

Two observations that make the case sharper:

1. **Client-side timeouts do not recover it.** During reproduction, a JUnit `@Timeout(60)` (same-thread interrupt mode) **failed to unblock** the hang — still running at 2m40s, requiring a manual `taskkill /F /T`. The block is in native/HTTP I/O, so interruption does not land. A bounded wait at this layer is therefore not redundant with timeouts elsewhere; it may be the only bound that actually works.
2. **The hang leaks processes.** The reproduction left a surefire JVM, a chromedriver, and nine `chrome.exe` children alive after the run should have ended. An unbounded server-side wait means nothing is responsible for tearing that tree down.

## Suggested fix

[...]

PR #4076, opened 2026-07-26
Fix reentrant WebDriver call inside CDP Fetch interception callback (#4046)

### LINKED ISSUE #4046: Nightly failure: Guided Workflows Live E2E
The scheduled workflow **Guided Workflows Live E2E** failed.

- Run: https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30189888129

Filed automatically. Dedupes on the `nightly-failure:guided-workflows-live` label (one open issue per workflow) and closes automatically when `Guided Workflows Live E2E` next succeeds.

MohabMohie:
## Summary

Fixes the deadlock behind #4046 (Guided Workflows Live E2E nightly, red 5/5 nights). Does **not** close #4046: the nightly's E2E path still fails on a second, separate, un-root-caused issue (see Handoff below), so the regression test stays `@Disabled`.

Related: #4068 (tester's reproduction, root-cause investigation handed off from there), #4064 (merged, already in `main`), #4066 (`CaptureManager.invoke()` unbounded wait — separate issue, untouched here, owned by another agent).

## Mechanism (full detail — this must not live only in a chat message)

**The bug:** `CaptureNetworkRecorder`'s network interception filter called `driver.getCurrentUrl()` — a classic WebDriver command — from inside the CDP `Fetch.requestPaused` callback, **before** continuing the very request that callback was handling.

`CaptureNetworkRecorder.java` (pre-fix), inside `createFilter()`:
```java
private Filter createFilter() {
    return next -> request -> {
        long transactionId = nextTransactionId.incrementAndGet();
        String pageOrigin = currentPageOrigin();   // <-- driver.getCurrentUrl(), BEFORE continuing
        ...
        HttpResponse response = next.execute(request);   // <-- this is what actually continues the paused request
        ...
    };
}

private String currentPageOrigin() {
    try {
        return originOf(driver.getCurrentUrl());   // reentrant classic WebDriver command
    } catch (RuntimeException exception) {
        return "";
    }
}
```

**Why that deadlocks — read straight from Selenium's own source** (`selenium-remote-driver-4.46.0-sources.jar`, pinned version, `org/openqa/selenium/devtools/idealized/Network.java:200-265`): the `Fetch.requestPaused` listener runs on a CDP executor thread and, for the request-stage pause, synchronously blocks on a `CompletableFuture` (`res.get()`) waiting for the **paired response-stage** `Fetch.requestPaused` event for the same transaction before it returns control to our filter's `next.execute()`/`continueRequest`. `enableFetchForAllPatterns()` (`v150Network.java:75-84`) explicitly registers interception at **both** `RequestStage.REQUEST` and `RequestStage.RESPONSE` — so this two-stage, blocking-until-paired design is deliberate on Selenium's side, not a bug there.

chromedriver serializes classic W3C command processing per session — one command in flight at a time. `driver.navigate().to(url)` triggers the top-level document request, which gets Fetch-paused. Our filter's callback (on the CDP executor thread) reaches `currentPageOrigin()` and issues a **second** classic command (`getCurrentUrl()`) on the *same session* — but the page load that `navigate()`/the pending request depends on cannot settle until **our own filter continues the paused request** (`next.execute()` → `continueRequest`), which cannot happen until `currentPageOrigin()` returns, which cannot happen until chromedriver services `getCurrentUrl()` — which it won't, because the session is still occupied by the load that's waiting on us. **Circular wait.** This also explains why the unrelated `executeScript()` call on the main thread hung too: it queued on the same serialized session behind the stuck state.

**Confirmed live, not inferred.** I ran the disabled regression test once with temporary debug instrumentation (thread name + `System.nanoTime()` at each step) against the *unfixed* code, hard-killed the process tree at a 50s external bound (PID-tracked `taskkill /F /T`, verified no leaks). Captured log:

```
[main] MAIN THREAD BEFORE navigate() at 11914732213900
[main] MAIN THREAD AFTER navigate() at 11914822311600
[main] MAIN THREAD BEFORE executeScript() at 11914822810700
[CDP Connection] tx=1 url=http://127.0.0.1:52630/ BEFORE currentPageOrigin() at 11914885055200
   <-- no "AFTER currentPageOrigin()" line ever printed; both threads dead-stopped here
```

This is also consistent with the isolation experiment in #4068 (disabling the BiDi UI-event collector left the hang unchanged) — the mechanism has nothing to do with BiDi at all; it is a pure CDP-Fetch + classic-command reentrancy bug in *our own* filter.

### The reusable rule

**Never call a classic WebDriver command from inside a CDP interception callback before continuing the paused request.** Selenium's `Fetch.requestPaused` handling blocks the callback thread on a paired CDP event; chromedriver serializes classic commands per session; any classic command issued from inside that callback — before the request it is handling has been continued — risks a circular wait against whatever the pending network activity is blocking (a navigation, a page load, any other in-flight command on the same session). This generalizes beyond this one bug to any future CDP `NetworkInterceptor`/`Filter` implementation in this codebase.

## Fix

- `CaptureNetworkRecorder`: added a `Supplier<String> currentPageUrlSupplier` constructor parameter; `currentPageOrigin()` now reads it instead of calling `driver.getCurrentUrl()`. Zero WebDriver calls remain in the interception filter's hot path.
- `ManagedCaptureRecorder`: wires `() -> currentUrl` — the field it already tracks passively via BiDi/polling navigation signals (never from inside the CDP callback). Made `currentUrl` `volatile` since it is now read cross-thread from the CDP executor.
- `CaptureNetworkRecorderTest`: updated all 11 existing call sites to pass `driver::getCurrentUrl` (safe there — Mockito-stubbed, no real chromedriver, no reentrancy risk), preserving every existing assertion (e.g. `dropsCrossOriginRequestsWhenFirstPartyOnlyIsTrue`) unchanged. Added `neverCallsDriverGetCurrentUrlFromTheInterceptionFilter`, which stubs `getCurrentUrl()` to throw and proves the filter never calls it. TDD: watched RED (`AssertionError` at the old `currentPageOrigin()` call site) then GREEN. Full suite: 13/13 pass.

Why not the other candidate fixes from the issue: "fix the filter so every request continues" did not apply as framed — the filter already called `next.execute()` unconditionally for every request (verified by direct read; no branch skips it). Reordering JS-injection-before-interceptor-install was the fallback candidate; unnecessary once the actual reentrant call was identified, and this fix is a strictly smaller, more targeted diff.

## Handoff: second, separate failure — un-root-caused, deliberately not investigated

With the deadlock fixed, `ManagedCaptureRecorderBrowserTest#startCompletesWhenApiCaptureIsEnabledAgainstARealBrowser` **no longer hangs** — it completes in ~32s against a 55s external bound (previously: still running past 2+ minutes, requiring a process-tree kill). That is confirmed, not inferred.

It now fails with a different, bounded error:
```
org.openqa.selenium.ScriptTimeoutException: script timeout
  at RemoteWebDriver.executeScript(RemoteWebDriver.java:555)
  at ManagedCaptureRecorder.start(ManagedCaptureRecorder.java:153)
```
against Selenium's default `timeouts:{script: 30000}` (visible in the driver's reported capabilities in the failure log). This is a clean, catchable timeout — not a hang — but its cause has **not** been investigated. Stating this plainly rather than guessing at a mechanism with no supporting evidence. The test stays `@Disabled` (updated reason references this specific failure) pending a fresh, separate diagnosis.

## Verification

- `CaptureNetworkRecorderTest` (unit, 13/13 pass): `mvn -pl shaft-capture -Dtest=CaptureNetworkRecorderTest test -DheadlessExecution=true -Dallure.automaticallyOpen=false`
- `mvn -pl shaft-capture test-compile` clean.
- Browser regression test run twice (bounded, PID-tracked, hard-killed/completed cleanly both times, zero leaked processes verified via `Get-CimInstance` after each): pre-fix reproduces the hang exactly as #4068 described; post-fix no longer hangs, fails on the separate `ScriptTimeoutException` above.
- Out of scope, untouched: `CaptureManager.java` (#4066's territory).

### DIFF
--- a/shaft-capture/src/main/java/com/shaft/capture/network/CaptureNetworkRecorder.java
+++ b/shaft-capture/src/main/java/com/shaft/capture/network/CaptureNetworkRecorder.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 /**
  * Per-session network transaction recorder. Registers a single Selenium DevTools {@link Filter}
@@ -70,6 +71,7 @@ public final class CaptureNetworkRecorder implements AutoCloseable {
     private final AtomicLong recordedCount = new AtomicLong();
     private final Deque<String> lastEndpoints = new ArrayDeque<>();
     private final Object lastEndpointsLock = new Object();
+    private final Supplier<String> currentPageUrlSupplier;
     private volatile boolean maxTransactionsWarned;
     private NetworkInterceptor interceptor;
     private volatile boolean started;
@@ -81,6 +83,10 @@ public final class CaptureNetworkRecorder implements AutoCloseable {
      * @param bodiesDirectory directory used to persist request/response bodies
      * @param options network capture filtering and body-capture options
      * @param sessionId owning capture session identifier, used for deterministic secret-ref derivation
+     * @param currentPageUrlSupplier supplies the last-known top-level page URL for first-party
+     *                               classification and {@code initiatorPageUrl} tagging; must be a
+     *                               non-blocking read (see {@link #createFilter()} for why it must
+     *                               never call back into the WebDriver session)
      * @param sink destination for each recorded, redacted network transaction
      * @param warn destination for safe one-time warnings
      */
@@ -89,17 +95,21 @@ public final class CaptureNetworkRecorder implements AutoCloseable {
             Path bodiesDirectory,
             NetworkCaptureOptions options,
             String sessionId,
+            Supplier<String> currentPageUrlSupplier,
             Consumer<RecordedTransaction> sink,
             Consumer<String> warn) {
-        if (driver == null || bodiesDirectory == null || sink == null || warn == null) {
+        if (driver == null || bodiesDirectory == null || currentPageUrlSupplier == null
+                || sink == null || warn == null) {
             throw new IllegalArgumentException(
-                    "Network recorder requires a driver, bodies directory, sink, and warn consumer.");
+                    "Network recorder requires a driver, bodies directory, current-page-URL supplier, "
+                            + "sink, and warn consumer.");
         }
         this.driver = driver;
         this.bodyStore = new NetworkBodyStore();
         this.sessionDirectory = bodiesDirectory.toAbsolutePath().normalize();
         this.options = options == null ? NetworkCaptureOptions.defaults() : options;
         this.sessionId = sessionId == null || sessionId.isBlank() ? "capture-session" : sessionId;
+        this.currentPageUrlSupplier = currentPageUrlSupplier;
         this.sink = sink;
         this.warn = warn;
     }
@@ -444,9 +454,18 @@ public final class CaptureNetworkRecorder implements AutoCloseable {
         }
     }
 
+    /**
+     * Never calls back into the WebDriver session (issue #4046): this runs on the CDP
+     * {@code Fetch.requestPaused} callback thread, before the request it is handling has been
+     * continued. A classic WebDriver command issued from here (e.g. {@code driver.getCurrentUrl()})
+     * is queued by chromedriver behind the very page load that cannot complete until this callback
+     * continues the paused request -- a circular wait that deadlocks both this call and any other
+     * in-flight WebDriver command on the same session (observed: {@code executeScript} hanging
+     * forever). {@link #currentPageUrlSupplier} must therefore be a non-blocking read.
+     */
     private String currentPageOrigin() {
         try {
-            return originOf(driver.getCurrentUrl());
+            return originOf(currentPageUrlSupplier.get());
         } catch (RuntimeException exception) {
             return "";
         }
--- a/shaft-capture/src/main/java/com/shaft/capture/runtime/ManagedCaptureRecorder.java
+++ b/shaft-capture/src/main/java/com/shaft/capture/runtime/ManagedCaptureRecorder.java
@@ -95,7 +95,9 @@ class ManagedCaptureRecorder {
      * injectable so the stop-time flush can be tested without launching a real browser.
      */
     private Supplier<List<String>> pendingBrowserDeletesReader = this::readPendingBrowserDeletesFromOverlay;
-    private String currentUrl;
+    // Volatile: read from the CDP network-interception callback thread (via the currentPageUrlSupplier
+    // passed to CaptureNetworkRecorder, issue #4046) as well as the owning thread and status() callers.
+    private volatile String currentUrl;
     private volatile boolean paused;
     private volatile boolean uiStopRequested;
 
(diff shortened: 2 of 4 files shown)


MohabMohie:
Rebased onto latest \`origin/main\` (\`68f0a3d24e\`) to resolve the \`CONFLICTING\`/\`DIRTY\` state.

**Cause:** this branch had cherry-picked PR #4068's test commit (\`1f2b17d42b\` locally as \`d655660dc2\`, adding the \`@Disabled\` regression test to \`ManagedCaptureRecorderBrowserTest.java\`). #4068 has since merged to \`main\` independently as \`a43d9b4e13\` with identical content, so the file conflicted with itself via two routes.

**Resolution:** \`git rebase origin/main\` detected the duplicate by patch-id and skipped it automatically (\`warning: skipped previously applied commit d655660dc2\`) — no manual conflict resolution was needed. The branch now carries a single commit (\`d04292af4d\`) containing the real fix plus this PR's legitimate javadoc/\`@Disabled\`-reason updates to the browser test (documenting the root cause found and the second, still-open defect tracked as #4077). The regression test remains \`@Disabled\`.

**Verified after rebase:**
- \`git diff origin/main...HEAD\` shows the production fix intact: \`Supplier<String> currentPageUrlSupplier\` in \`CaptureNetworkRecorder.java\`, \`currentPageOrigin()\` no longer calling \`driver.getCurrentUrl()\`, and the \`volatile\` field in \`ManagedCaptureRecorder.java\`.
- \`mvn -pl shaft-capture test -Dtest=CaptureNetworkRecorderTest\`: \`Tests run: 13, Failures: 0, Errors: 0, Skipped: 0\`.
- \`mvn -pl shaft-capture compile test-compile\`: \`BUILD SUCCESS\`.
- Force-pushed; \`gh pr view 4076\` now reports \`"mergeable":"MERGEABLE"\` (was \`CONFLICTING\`).

Real-browser E2E tests were intentionally not run (unrelated, still-open #4077 hang risk).

MohabMohie:
**Reviewed and arming.** Merging this despite `Guided Workflows Live E2E` still being red, deliberately.

This fix is independently correct and independently valuable. It eliminates a real reentrant deadlock — our interception filter calling a classic WebDriver command from inside the CDP `Fetch.requestPaused` callback before continuing the paused request, deadlocking against chromedriver''s per-session command serialization. The evidence is a live instrumented run showing both threads dead-stopped at that exact point, not inference.

Measured effect: the browser test went from hanging past a 55s hard bound to completing in 32s. The infinite hang is gone.

What it does **not** do, and correctly does not claim to: fix #4077, the bounded `ScriptTimeoutException` at `ManagedCaptureRecorder.java:153` that sits behind it. #4046 stays open, the regression test stays `@Disabled`, and the nightly stays red until a dispatched run proves otherwise. Holding a proven deadlock fix hostage to a separate un-root-caused defect would leave the deadlock on `main` for no benefit.

Safety checked before arming: the regression test remains `@Disabled`, so nothing reachable from `pr-gate.yml` can hang — which matters, because `.github/actions/capture-browser-e2e` selects that class **by name**, so `@Tag` alone would not have protected the gate. `CaptureNetworkRecorderTest` is 13/13 including the new `neverCallsDriverGetCurrentUrlFromTheInterceptionFilter`, which pins the fix by stubbing `getCurrentUrl()` to throw.

The rule this establishes is worth more than the diff: **never call a classic WebDriver command from inside a CDP interception callback before continuing the paused request.**

PR #4181, opened 2026-07-26
fix: don't overwrite generated test source when compile/replay fails [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #4166: Generated test source is written before compile and replay, so a failed run overwrites a working test
Found during the area C audit under tracker 4160, as sharper evidence than the issue-4029 framing it was first attributed to. Filing separately because it is independently fixable and its consequence is worse than a silent failure: it destroys work the user already had.

## What happens

`CaptureGenerator.generate()` (`shaft-capture/src/main/java/com/shaft/capture/generate/CaptureGenerator.java:279-303`) calls `atomicWrite(paths.source(), source)` **before** `compile()` and `replay()` run, and never rolls it back when either fails. Only the JSON report is marked `FAILED`.

Combined with `/codegen` defaulting to `overwrite=true` (`AssistantCommand.java:2631`), a generation whose replay fails silently replaces a previously-good generated test with source that does not compile. The user asked for a regeneration, got a failure report, and separately lost the working file they already had.

The write being atomic is not the issue — atomicity guarantees the file is not left half-written, not that it should have been written at all.

## Why this is not issue 4029

Issue 4029 covers making codegen failures loud and diagnosable — the user not learning that generation failed. This is the adjacent and more damaging half: the user's existing artifact is gone regardless of how loudly the failure is reported. Fixing 4029's reporting would still leave the overwrite in place.

## Shape of the fix

Write to a temporary location, run compile and replay against it, and promote to the final path only on success. On failure, leave the previous file untouched and say so in the failure report — "your existing test was left unchanged" is information the user needs.

## Sequencing

Blocked on issue 4161, which is actively changing `CaptureGenerator`/`GeneratedTestValidator` wiring right now. Do not dispatch these concurrently.

Related: tracker 4024, tracker 4160, issue 4029.

waterWang:
### Description

When code generation fails (compile error or replay failure), the generated source file was already written to disk before the validation ran. Combined with `/codegen defaults to overwrite=true`, this silently replaces a previously-working test with source that doesn't compile.

### Root Cause

`CaptureGenerator.generate()` called `atomicWrite(paths.source(), source)` and `atomicWrite(paths.data(), dataJson)` **before** `compile()` and `replay()` ran. A failed generation would overwrite the old working files with broken source, then only mark the JSON report as FAILED. The user loses their working test file.

### Fix

Moved the `atomicWrite` calls to after the compile and replay validation check. The source and data files are now only written to disk when both compile and replay succeed. If either fails, the previous files remain intact.

### Before

```java
atomicWrite(paths.source(), source);   // written before validation
atomicWrite(paths.data(), dataJson);   // written before validation
compile();  // may fail
replay();   // may fail
```

### After

```java
compile();  // may fail
replay();   // may fail
if (successful) {
    atomicWrite(paths.source(), source);  // only written after success
    atomicWrite(paths.data(), dataJson);  // only written after success
}
```

Fixes #4166


MohabMohie (changes requested):
Thank you for finding and fixing this — the write-before-validation bug is real and issue 4166 describes exactly the problem you're solving. Unfortunately this specific implementation has a regression I verified empirically before deciding not to merge it directly.

**The problem:** `GeneratedTestValidator.compile(Path source, Path classesDirectory)` reads the source file from disk via `files.getJavaFileObjectsFromPaths(List.of(source))` -- it does not take the generated source as an in-memory string. Moving `atomicWrite(paths.source(), source)` to *after* the compile/replay check means `compile()` runs against whatever is currently on disk at that path, not the new content being generated:

- On a first-time generation (no prior file at that path), `compile()` has nothing to read, so it fails immediately -- `successful` becomes `false`, and the new source is never written, for every brand-new generation.
- On a regeneration (a file already exists from a prior success), `compile()`/`replay()` validate the *old* file, and if that happens to still pass, the *new* `source` string gets written afterward -- unvalidated, despite the check appearing to have passed.

**Evidence, not just reasoning:** I test-merged this branch against the current `main` in a disposable worktree and ran the module's real test suite (`CaptureGeneratorTest`, `CaptureGeneratorHealingSeedingTest`): **23 failures and 1 error out of 36 tests**, including a literal `NoSuchFile ... ExampleFormTest.java` from the compiler trying to read a file that was never written. This reproduces on the very first generation in every affected test.

**A fix that would work:** write the generated source to a temporary path first, run `compile()`/`replay()` against *that* path, and only then atomically promote it to the real `paths.source()`/`paths.data()` on success -- leaving the previous file untouched on failure. That's the shape already scoped in issue #4166's body. Happy to see an updated version of this PR taking that approach -- the underlying defect you found is worth fixing, this just needs the write deferred to a staging location rather than skipped outright.

Since your #4182 branch is stacked on this commit, I've ported just its (unrelated, and correct) font fix directly to unblock it: #4192, with attribution preserved.

MohabMohie:
Closing as superseded, not as a reflection on the contribution. The same defect (issue 4166, generated test source overwritten before compile/replay validates it) has just been fixed and merged internally as PR #4201, which stages the source and only promotes it to the real path on a confirmed successful replay -- with regression coverage and an independent hostile review. That PR also carries follow-up tracking for a data-JSON variant of the same exposure (issue 4202) and a minor staged-file cleanup gap (issue 4206) it surfaced along the way.

This PR was also independently found earlier in the review process to introduce a regression breaking 23+ existing tests, so it wasn't mergeable as submitted regardless. Thank you for flagging the underlying bug -- reopen or comment if you believe #4201's fix doesn't fully cover what this PR was going for.

PR #4182, opened 2026-07-26
fix: use IDE editor font in ShaftMcpSetupPanel instead of hardcoded Font.MONOSPACED [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #4179: Three more hardcoded Font.MONOSPACED sites in ShaftMcpSetupPanel ignore the IDE editor font
Same defect class as the one fixed in PR #4176 for `AssistantTranscriptView`, found by that PR's reviewer when asked to grep for siblings rather than stop at the reported site.

## Where

- `shaft-intellij/src/main/java/com/shaft/intellij/ui/ShaftMcpSetupPanel.java:513`
- `shaft-intellij/src/main/java/com/shaft/intellij/ui/ShaftMcpSetupPanel.java:1015`
- `shaft-intellij/src/main/java/com/shaft/intellij/ui/ShaftMcpSetupPanel.java:2670`

Each hardcodes the logical `Font.MONOSPACED` family instead of resolving the IDE's configured editor font, so a user who has set a monospaced font in their IDE sees a different one in these surfaces.

## The fix that already exists

PR #4176 added `AssistantTranscriptView.monospacedFontFamily()` (`AssistantTranscriptView.java:1068-1085`), which resolves through `EditorColorsScheme.getFont(EditorFontType.PLAIN)` via `EditorColorsManager`. Reuse that mechanism rather than writing a second one; lifting it to a shared helper is likely the cleaner shape once there are four call sites across two files.

## The trap to avoid, verified

`EditorColorsManager.getInstance()` **throws** rather than returning null when there is no live `Application` — the reviewer confirmed this from the actual bytecode (`javap -c` on the class in the test sandbox jar shows `getApplication()` dereferenced unconditionally with no null branch). Guard on `ApplicationManager.getApplication() == null` **first**. Copying the codebase's existing `EditorColorsManager.getInstance() == null` idiom broke 8 previously-green tests when PR #4176 tried it; that dead idiom is tracked separately as issue 4175.

## Verification note

This module's Gradle tests never build a live IntelliJ `Application` — no test class extends `BasePlatformTestCase` — so a headless render always takes the fallback branch and before/after screenshots will be byte-identical. The improvement is only observable in a live IDE. Verify at source level, as PR #4176 did, and say so plainly rather than implying visual proof.

Related: PR #4176, issue 4163, issue 4175, tracker 4160.

waterWang:
### Description

Same defect class as the one fixed in PR #4176 for `AssistantTranscriptView`. Three more sites in `ShaftMcpSetupPanel` hardcode the logical `Font.MONOSPACED` family instead of resolving the IDE's configured editor font.

### Changes

Replaced 3 hardcoded `Font.MONOSPACED` references in `ShaftMcpSetupPanel.java` with `AssistantTranscriptView.monospacedFontFamily()`:

- Line 513: `details.setFont(...)`
- Line 1018: `area.setFont(...)`  
- Line 2673: `StyleConstants.setFontFamily(...)`

Reuses the same resolution mechanism already added in PR #4176 — falls back to `Font.MONOSPACED` when no live `Application` exists.

Fixes #4179


MohabMohie:
Your `Font.MONOSPACED` -> `AssistantTranscriptView.monospacedFontFamily()` fix here is correct and exactly the shape issue 4179 asked for -- reusing the existing mechanism from #4176 rather than a new one. Verified it compiles and the full `ShaftPanelSetupTest` suite stays green.

The reason it's not merging directly: this branch is stacked on top of #4181's commit, and #4181 has a confirmed regression (write-before-validation deferred past where `GeneratedTestValidator.compile()` reads the file from disk -- details on that PR). To land your font fix without inheriting that, I cherry-picked just your commit onto a clean branch off current `main`, preserving your authorship: #4192.

No action needed from you unless you'd like to rebase this PR once #4181 lands its own fix -- at that point this one would just be the same diff again and can close in favor of whichever merges cleanly.

MohabMohie:
Closed in favor of the othe PR which has a more targetted scope.

PR #4190, opened 2026-07-26
THROWAWAY: verify font-install fix against PR 4184's welcome-bubble change (do not merge)

MohabMohie:
Disposable verification PR only, not for merging. Combines PR 4184's tip commit c0bf2de9d9 (issue 4174 welcome-bubble height fix) with the font-install fix from #4189, purely to observe whether firstRunWelcomeDismissButtonIsVisibleWithoutScrollingAtNarrowDarkWidth now passes on ubuntu-22.04 CI with real text fonts installed. Will be closed and the branch deleted once CI reports a result. See #4189 for the real fix.

### DIFF
--- a/.github/actions/intellij-verify/action.yml
+++ b/.github/actions/intellij-verify/action.yml
@@ -37,6 +37,16 @@ runs:
       uses: gradle/actions/setup-gradle@v6
       with:
         gradle-version: '9.0.0'
+    # ubuntu-22.04 runner images ship only fonts-noto-color-emoji (no DejaVu/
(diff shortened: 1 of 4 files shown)


MohabMohie:
Hypothesis falsified by this run's own log: fonts-dejavu-core and fonts-liberation were already the newest version on the ubuntu-22.04 runner before this step ran, and the target test still failed identically. Closing this throwaway verification PR; root-cause investigation is redirected elsewhere. See PR #4189 for the (still potentially useful, but not root-cause) font-install step.

Issue #3976, opened 2026-07-21
Assistant chat: normal-completion finalization discards Verbose raw buffer (point 3, split from #3962)


MohabMohie:
Split out from #3962 per an independent design review's explicit recommendation: this and #3962 don't share a mechanism, so they shouldn't share a PR.

## The gap

`ShaftAssistantPanel#finishLocalAgentResponse` always replaces the streaming bubble with the polished final answer on normal completion, discarding the accumulated `localAgentOutput` raw buffer regardless of Verbose state. Concretely: if a compacted/pointer-text milestone line (e.g. `RAW_STRUCTURED_OUTPUT_NOTICE`, added by PR #3961) was the last thing streamed before the run completed normally, its "enable Verbose to see it" pointer becomes a dead end — there's no more streaming bubble left to re-render once Verbose is toggled on, because finalization already replaced it.

This is the same root cause as #3920's original point 3 finding, distinct from #3962: **no `CompletableFuture` cancellation or concurrency is involved here** — it's a finalization *display* decision, not a race. `stopLocalAgentStreaming` (the Kill path) already conditions its rendering on `verboseLocalAgentOutput()`'s live state (fixed by PR #3961) — this issue is applying that same "respect live Verbose state at finalization time" principle to the normal-completion path.

## Scope

In `finishLocalAgentResponse`, when Verbose is on, append the raw accumulated buffer (or at least the portion not already represented in the polished answer) alongside/under the polished final answer, instead of unconditionally discarding it. Mirror the reasoning `stopLocalAgentStreaming` already uses for the Kill path.

## Sequencing note

Touches `ShaftAssistantPanel.java` (`finishLocalAgentResponse`) — same file as #3955 (merged), #3965, #3968, and #3962's PR 1. Queue behind whichever of those is in flight; don't parallelize.

Issue #3920, opened 2026-07-21
Assistant chat: full transparency — eliminate silent content drops in the transcript pipeline


MohabMohie:
Subtask of the chat-transparency tracking issue (owner directive 2026-07-21: 'ensure full transparency and fully forwarding agent responses').

Verified loss/suppression points (scout report, file:line on main):
- Raw-JSON drop: sub-lines starting {/[ silently dropped in non-verbose (ShaftAssistantPanel :2368-2369) — must instead be formatted (or verbose-gated), never silently discarded.
- Meta-tool drop: NON_ACTIONABLE_META_TOOL_NAMES ('ToolSearch') strips milestone lines (:2334, :2410-2418) — decide policy: show formatted or verbose-gate; not silent.
- Tool-result compaction at 400 chars (TOOL_RESULT_INLINE_CHAR_LIMIT :2380, compactIfRawToolResult :2389-2407) — keep, but the full content must be reachable (disclosure/verbose), pointer text already exists.
- Transcript cap-trim genuinely loses oldest bubbles (AssistantTranscriptView trimOldestIfOverCapacity :267-289, MAX_MESSAGES_PER_SESSION) — evaluate raising the cap or preserving trimmed content (e.g. collapse into a 'N earlier messages' disclosure) so agent responses are never lost mid-session.
- Terminal-event CONSUMED suppression resurfaces via finalOutput() (parser :697/:743) — verify no path loses the terminal text when a run is cancelled/killed mid-stream.

Spec: no agent-response content is ever silently dropped; everything is either rendered formatted, behind a disclosure, or behind Verbose.

Issue #3768, opened 2026-07-19
IntelliJ Assistant: soft Cancel always shows as Failed, never Cancelled (CompletableFuture wraps thrown CancellationException in CompletionException)


MohabMohie:
## Problem

A plain (soft) Cancel of a local-agent run in the Assistant panel renders as a
generic **Failed** result instead of **Cancelled**, because
`ShaftAssistantPanel.showAgentResult`'s cancellation check never fires for
this path.

## Evidence

- `ShaftMcpInvocation.cancel()` (`shaft-intellij/src/main/java/com/shaft/intellij/mcp/ShaftMcpInvocation.java:22-25`)
  never calls `future.cancel(...)` -- it just runs `cancelAction` (which sets
  `cancellationRequested` and calls `Process#destroy()`) and always returns
  `true`.
- The only way the invocation's `future` ever completes after a soft cancel is
  via `AssistantLocalAgentRunner.run()`'s own `throw new
  CancellationException("Operation cancelled")` once `cancellationRequested`
  is observed (`shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantLocalAgentRunner.java:516,528,554,559`),
  which is caught by `CompletableFuture.supplyAsync`'s internal completion
  path.
- `CompletableFuture` wraps *any* exception thrown by an async-stage task in a
  `CompletionException` when completing the future (`AsyncSupply.run()` ->
  `completeThrowable` -> `encodeThrowable`), so the future's stored cause is
  `CompletionException(CancellationException)`, not a bare
  `CancellationException`. Verified empirically:
  `CompletableFuture.supplyAsync(() -> { throw new CancellationException(); })`
  then `.whenComplete((r, err) -> ...)` receives
  `err = java.util.concurrent.CompletionException: java.util.concurrent.CancellationException: ...`,
  and `future.isCancelled()` reads `false`.
- `ShaftAssistantPanel.showAgentResult` (`shaft-intellij/src/main/java/com/shaft/intellij/ui/ShaftAssistantPanel.java:1932-1970`)
  computes `boolean cancelled = error instanceof CancellationException;` from
  exactly this `whenComplete` error -- so for a soft cancel, `cancelled` is
  always `false`, `killed` is always `false`, and the run falls through to
  `showAgentToolResult(...)` as a plain failure instead of
  `showAgentCancelled(...)`.
- The **force Kill** path is unaffected in practice only because it has an
[...]

Issue #3758, opened 2026-07-18
Flaky: AssistantLocalAgentRunnerCancellationTest.forceCancelCallsDestroyForciblyNotDestroyAndClearsTheProcessReference


MohabMohie:
Observed during PR #3755 verification: in a full `gradle test` run of shaft-intellij, `forceCancelCallsDestroyForciblyNotDestroyAndClearsTheProcessReference` failed with "kill() must mark the future cancelled synchronously via future.cancel(true)", then passed when the class was rerun alone. The class had zero diff in the PR — pre-existing timing-sensitive assertion around `future.cancel(true)`, likely a race between kill() and the executor observing cancellation.

Suggested approach: make the cancellation observable deterministically (latch on the runner's cancel path) or assert eventually with a bounded await instead of synchronously.

Evidence: full-suite run 2026-07-19, 976 tests / 1 failure (this test); isolated rerun green.

PR #4204, opened 2026-07-27
Recover a cancelled/killed local-agent run's terminal answer

### LINKED ISSUE #3962: Assistant chat: cancelled/killed runs can silently discard an already-parsed terminal answer (needs design decision)
Follow-up from #3920 (PR #3961). Escalated by the implementing agent rather than guessed at, since the fix changes behavior for every run and touches concurrency-sensitive code #3758 already carefully reasoned through.

## The gap

`AssistantLocalAgentRunner.run()` throws `CancellationException` unconditionally once a run is cancelled, discarding an already-parsed terminal answer even if the stream parser had already produced one (`streamParser.hasTerminalEvent()` would be true). This looked like a real content-loss bug worth a guarded fix.

## Why it can't be fixed by simply guarding the throw

Investigated and reverted (clean `git diff`, no code shipped): `ShaftMcpInvocation.cancel()`/`kill()` deliberately call `future.cancel(true)` **before** running the cancel/kill action (issue #3758/#3768, validated across 500 runs per that file's own comment). Once the future is Future-cancelled, any later return value from `run()` is silently discarded by the JDK's own `Future` contract — no caller can ever observe it. Guarding the throw on `hasTerminalEvent()` is therefore dead code: verified empirically, test outcome was unchanged before/after the guard.

## What an actual fix requires

Forward the terminal event into the *live* output stream instead of holding it back as `CONSUMED`, so the UI observes it before cancellation ever reaches the future. This changes behavior for every run (not just the cancel/kill edge case) and carries a real risk of duplicating the final answer if not sequenced carefully against the existing streaming-bubble rendering path. It directly touches the same concurrency-sensitive code #3758 spent real effort validating (500-run stress validation) — needs a deliberate design pass, not a quick patch.

## Adjacent, same root cause

`#3920`'s point 3 (tool-result compaction reachability) has the identical shape on the *normal completion* path, not just cancel/kill: `finishLocalAgentResponse` always replaces the streaming bubble with the polished final answer, discarding `localAgentOutput` regardless of Verbose state — so if a compacted/pointer-text line was the last line streamed before normal completion, the "enable Verbose to see it" pointer is a dead end. Worth solving together with this issue since the fix (don't let finalization silently discard the raw accumulated buffer) is likely shared.

## Scope note

Not part of the sequential #3923 chain — that chain's remaining subtasks (#3921, #3922) don't depend on this, and this needs a design decision before it can be scoped into a PR. Fold into #3923 planning only if the owner wants it sequenced there; otherwise standalone backlog.

MohabMohie:
## Summary

Closes #3962.

- `ShaftMcpInvocation.cancel()`/`kill()` call `future.cancel(true)` **before** running their cancel/kill action (the ordering issue #3758/#3768 stress-validated across 500 runs — untouched by this PR). Once that call wins, the JDK `Future` contract silently discards whatever `AssistantLocalAgentRunner.run()` later returns, including an already-parsed terminal answer. A prior attempt to fix this by guarding the `CancellationException` throw on `streamParser.hasTerminalEvent()` was dead code for exactly this reason — the guard runs *after* the future is already done, so it can never be observed.
- Adds a `terminalAnswerConsumer` side channel: new additive overloads of `AssistantLocalAgentRunner.start`/`startWithOptionalCompact`/`run` (existing overloads unchanged, delegate with `null`). `run()`'s own `finally` block notifies this consumer exactly once per invocation — with the parsed terminal answer, or `null` if the structured stream never produced one — regardless of whether cancellation raced it. This channel is never subject to `future.cancel(true)`, since it isn't the primary future at all.
- `ShaftAssistantPanel` wires this into a new per-run `pendingTerminalAnswer` field:
  - **Soft Cancel** reaches `showAgentResult` asynchronously already (no existing synchronous UX to preserve), so it now waits on the companion before rendering, instead of immediately showing the bare `_Cancelled._` placeholder and permanently losing the answer the instant it *does* resolve.
  - **Kill** finalizes `_Killed._` synchronously in `stopLocalAgentStreaming`, at button-click time — before `ShaftMcpInvocation.kill()` even runs, so the companion is essentially guaranteed not to be resolved yet. Delaying that existing instant feedback felt like a bigger, riskier behavior change than the issue asked for, so instead the same already-rendered message is upgraded in place once (if) the companion resolves with a real answer.

## Why this doesn't touch #3758's invariant

`future.cancel(true)` is called at the exact same place, in the exact same order, as before in both `cancel()` and `kill()`. The new companion channel is a parallel, independent completion (`run()`'s own `finally`) that the primary future's cancellation cannot see or affect either way.

## Anti-duplication proof

- **Cancel path**: exactly one call site (`showAgentResult`'s `cancelled` branch) renders the cancelled bubble, gated by an `if (pending == null) { render now } else { render once the companion resolves }` — mutually exclusive by construction, so it can never render from both branches.
- **Kill path**: the synchronous finalize (existing code, unchanged) always renders first; the async upgrade only ever *replaces* that same message's `markdown` in place (never appends), and only when its content still contains the exact `_Killed._` marker it was finalized with — otherwise it's a documented no-op (the transcript moved on in the meantime; the narrower fallback the design review flagged as acceptable, "Approach B"-style, only for this edge case).
- Covered by new tests:
  - `AssistantLocalAgentRunnerTerminalAnswerCompanionTest#terminalAnswerConsumerReceivesTheParsedAnswerEvenWhenKillWinsTheRace` — reproduces the exact race (stub process blocks in `waitFor`, terminal event fully parsed via a latch on stdout EOF, *then* `kill()` races it) and asserts the primary future is still cancelled (guarding #3758) while the companion still carries the parsed answer.
  - `...ReceivesNullWhenTheRunIsKilledBeforeAnyTerminalEvent` — the companion always fires (with `null`), so a waiting caller is never left hanging.
  - `ShaftPanelSetupTest#assistantCancelWaitsForPendingTerminalAnswerAndRendersItInsteadOfBarePlaceholder` — drives the real deferred-render path with a not-yet-completed `CompletableFuture`, asserts nothing renders before it resolves, then asserts the answer renders exactly once.
  - `ShaftPanelSetupTest#assistantKillUpgradesTheSameFinalizedMessageOnceTerminalAnswerArrives` — asserts the message count never grows across the upgrade (no second bubble) and the answer appears exactly once.

## What a reviewer should re-check

- The existing `AssistantLocalAgentRunnerCancellationTest` suite (#3758's own regression coverage) is unchanged and still green — worth independently confirming the ordering assertions (`future().isCancelled()`, `destroy()` vs `destroyForcibly()`) still hold, since that's the invariant this PR must never regress.
- `terminalResultEventsAreConsumedIntoTheFinalAnswerNotEchoedRaw` (pins that a terminal event is never echoed into the live stream) is untouched — this PR deliberately does not forward anything into the live `outputConsumer`, unlike the originally-considered "Approach A" the linked design review rejected.
- The message-index + marker-content guard in `upgradeKilledLocalAgentMessage` is the one genuinely new piece of transcript-mutation logic — worth tracing `replaceLocalAgentStreamPlaceholder`'s existing index-tracking gotcha (a milestone bubble appended after the placeholder shifts the index) against this new path, since it reads `chatState.activeSession()` fresh at upgrade time rather than trusting a stale snapshot.
- A real 500-run-style stress pass isn't practical here, but the specific thing to hammer is: `kill()` immediately followed (before the background thread's `finally` runs) by starting a *new* run — confirming the new run's own `pendingTerminalAnswer` is never confused with the previous run's late-arriving companion (this PR captures the old reference into a local variable and nulls the field before scheduling, specifically to avoid that).

## Design context

This follows an independent architectural review already recorded on #3962 (Approach C: a companion future that cancel/kill never touch, completed in a `finally`) and confirms via `#3976` (already shipped, closed) that the adjacent normal-completion Verbose-buffer gap does **not** share a mechanism with this one and was correctly split into its own PR.

Related: #3920, #3758, #3768 (all read for context; no closing keywords used against them intentionally).

## Test plan
- [x] `AssistantLocalAgentRunnerTerminalAnswerCompanionTest` (new) — watched red (no 8-arg `start` overload), then green.
- [x] `ShaftPanelSetupTest#assistantCancelWaitsForPendingTerminalAnswerAndRendersItInsteadOfBarePlaceholder` / `#assistantKillUpgradesTheSameFinalizedMessageOnceTerminalAnswerArrives` (new) — watched red (`NoSuchFieldException: pendingTerminalAnswer`), then green.
- [x] Full `com.shaft.intellij.ui.*` package: 872 tests, 0 failures, 0 errors — run twice in a row (plus 5 repeats of the new companion test alone) after finding and fixing one race in the new test itself (`kill()` returning doesn't wait for the background thread's own `finally`).
- [x] `compileJava`/`compileTestJava` clean.
- [ ] Not run: full module `gradlew test` (guard-hooked off per repo policy) or IntelliJ screenshot renderer (no visible UI change — same Swing components, only text content differs).

https://claude.ai/code/session_01B77p5FMfqraiw1amwJGgVH

### DIFF
--- a/shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantLocalAgentRunner.java
+++ b/shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantLocalAgentRunner.java
@@ -197,12 +197,38 @@ final class AssistantLocalAgentRunner {
             LocalAgentApprovalBridge.ApprovalRequestHandler approvalHandler,
             ApprovalBridgeLauncher bridgeLauncher,
             boolean verbose) {
+        return start(invocation, outputConsumer, processLauncher, requireCommandAvailable, approvalHandler,
+                bridgeLauncher, verbose, null);
+    }
+
+    /**
+     * Same as the 7-arg overload above, plus {@code terminalAnswerConsumer} (issue #3962). {@code
+     * ShaftMcpInvocation.cancel()}/{@code kill()} call {@code future.cancel(true)} <em>before</em>
+     * running their cancel/kill action (issue #3758/#3768's validated ordering, untouched here) -- so
+     * once that call wins, the JDK's own {@code Future} contract silently discards whatever {@link
+     * #run} later returns or throws, even a real parsed terminal answer. {@code
+     * terminalAnswerConsumer}, when non-null, is notified exactly once from {@code run()}'s own
+     * {@code finally} block (see the canonical {@link #run} overload) with the parsed terminal answer
+     * -- or {@code null} if the structured stream never produced one -- regardless of {@code
+     * future.cancel(true)}, since it is a side channel the primary future's cancellation cannot touch.
+     * {@code null} (the default via the 7-arg overload) means "no one is listening"; every existing
+     * caller is unaffected.
+     */
+    static ShaftMcpInvocation start(
+            AssistantCommand.Invocation invocation,
+            Consumer<String> outputConsumer,
+            ProcessLauncher processLauncher,
+            boolean requireCommandAvailable,
+            LocalAgentApprovalBridge.ApprovalRequestHandler approvalHandler,
+            ApprovalBridgeLauncher bridgeLauncher,
+            boolean verbose,
+            Consumer<String> terminalAnswerConsumer) {
         JsonObject arguments = invocation.arguments();
         AtomicReference<Process> processReference = new AtomicReference<>();
         AtomicBoolean cancellationRequested = new AtomicBoolean();
         CompletableFuture<ShaftMcpToolResult> future = CompletableFuture.supplyAsync(() -> run(
                 arguments, processReference, cancellationRequested, outputConsumer, processLauncher,
-                requireCommandAvailable, approvalHandler, bridgeLauncher, verbose),
+                requireCommandAvailable, approvalHandler, bridgeLauncher, verbose, terminalAnswerConsumer),
                 ShaftPluginExecutor.getInstance().executor());
         return new ShaftMcpInvocation(
                 future,
@@ -275,6 +301,24 @@ final class AssistantLocalAgentRunner {
             Consumer<String> outputConsumer,
             LocalAgentApprovalBridge.ApprovalRequestHandler approvalHandler,
             boolean verbose) {
+        return startWithOptionalCompact(invocation, autoCompactEnabled, outputConsumer, approvalHandler, verbose, null);
+    }
+
+    /**
+     * Same as the 5-arg overload above, plus {@code terminalAnswerConsumer} (issue #3962) -- see the
+     * matching {@link #start(AssistantCommand.Invocation, Consumer, ProcessLauncher, boolean,
+     * LocalAgentApprovalBridge.ApprovalRequestHandler, ApprovalBridgeLauncher, boolean, Consumer)}
+     * overload for what it is notified with and when. {@link ShaftAssistantPanel} calls this overload
+     * for its local-agent runs so it can recover a cancelled/killed run's parsed terminal answer,
+     * which {@code future.cancel(true)} would otherwise discard unconditionally.
+     */
+    static ShaftMcpInvocation startWithOptionalCompact(
+            AssistantCommand.Invocation invocation,
+            boolean autoCompactEnabled,
+            Consumer<String> outputConsumer,
+            LocalAgentApprovalBridge.ApprovalRequestHandler approvalHandler,
+            boolean verbose,
+            Consumer<String> terminalAnswerConsumer) {
         if (autoCompactEnabled) {
             ShaftMcpToolResult compactResult = runCompactPreamble(invocation.arguments());
             if (outputConsumer != null && compactResult.output() != null && !compactResult.output().isBlank()) {
@@ -282,7 +326,7 @@ final class AssistantLocalAgentRunner {
             }
         }
         return start(invocation, outputConsumer, AssistantLocalAgentRunner::launchProcess, true, approvalHandler,
-                LocalAgentApprovalBridge::start, verbose);
+                LocalAgentApprovalBridge::start, verbose, terminalAnswerConsumer);
     }
 
     /**
@@ -528,7 +572,7 @@ final class AssistantLocalAgentRunner {
             ProcessLauncher processLauncher,
             boolean requireCommandAvailable) {
         return run(arguments, processReference, cancellationRequested, outputConsumer, processLauncher,
-                requireCommandAvailable, null, LocalAgentApprovalBridge::start, true);
+                requireCommandAvailable, null, LocalAgentApprovalBridge::start, true, null);
     }
 
     private static ShaftMcpToolResult run(
@@ -540,7 +584,8 @@ final class AssistantLocalAgentRunner {
             boolean requireCommandAvailable,
             LocalAgentApprovalBridge.ApprovalRequestHandler approvalHandler,
             ApprovalBridgeLauncher bridgeLauncher,
-            boolean verbose) {
+            boolean verbose,
+            Consumer<String> terminalAnswerConsumer) {
         Duration timeout = Duration.ofSeconds(intValue(arguments, "timeoutSeconds", DEFAULT_TIMEOUT_SECONDS));
         LocalAgentApprovalBridge.ApprovalRequestHandler narratingApprovalHandler =
                 narrateApproval(approvalHandler, outputConsumer);
@@ -563,6 +608,11 @@ final class AssistantLocalAgentRunner {
                 }
             }
         }
+        // Declared here (not inside the try below) so the outer finally can always reach it -- issue
+        // #3962's terminalAnswerConsumer notification below must fire from every exit path of this
+        // method, including the early-return guard clauses at the top of the try that run before a
+        // structured stream parser is even created.
+        StructuredStreamParser streamParser = null;
         try {
             List<String> command = commandFor(arguments, bridge);
             if (command.isEmpty()) {
@@ -580,7 +630,7 @@ final class AssistantLocalAgentRunner {
             String stdin = string(arguments, "prompt", "");
             Path workingDirectory = workingDirectory(arguments);
             boolean isDefaultCommand = defaultCommand(arguments);
-            StructuredStreamParser streamParser = isDefaultCommand ? structuredStreamParser(command) : null;
+            streamParser = isDefaultCommand ? structuredStreamParser(command) : null;
             Consumer<String> effectiveConsumer =
                     effectiveConsumer(outputConsumer, streamParser, isDefaultCommand);
             OutputStream stdinStream = null;
@@ -656,6 +706,20 @@ final class AssistantLocalAgentRunner {
                 processReference.set(null);
             }
         } finally {
+            // Issue #3962: notified exactly once, from every exit path of this method (normal return,
+            // timeout, a thrown exception, or a cancellation the checks above turned into a thrown
+            // CancellationException) -- unconditionally, regardless of whether cancellationRequested
+            // was ever observed. This runs strictly before this method returns/throws to the
+            // supplyAsync wrapper that completes the primary future, so by the time any caller's
+            // future.whenComplete callback can possibly observe this run's outcome, this notification
+            // has already happened -- closing the race #3758/#3768 exploit differently (this side
+            // channel is never subject to future.cancel(true), which only ever affects the primary
+            // future's own stored result).
+            if (terminalAnswerConsumer != null) {
+                terminalAnswerConsumer.accept(streamParser != null && streamParser.hasTerminalEvent()
+                        ? streamParser.finalOutput()
+                        : null);
+            }
             if (bridge != null) {
                 bridge.close();
             }
--- a/shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantLocalAgentRunner.java
+++ b/shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantLocalAgentRunner.java
@@ -706,22 +706,27 @@ final class AssistantLocalAgentRunner {
                 processReference.set(null);
             }
         } finally {
-            // Issue #3962: notified exactly once, from every exit path of this method (normal return,
-            // timeout, a thrown exception, or a cancellation the checks above turned into a thrown
-            // CancellationException) -- unconditionally, regardless of whether cancellationRequested
-            // was ever observed. This runs strictly before this method returns/throws to the
-            // supplyAsync wrapper that completes the primary future, so by the time any caller's
-            // future.whenComplete callback can possibly observe this run's outcome, this notification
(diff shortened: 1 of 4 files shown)


MohabMohie:
## Fixed per the adversarial review — all findings addressed in one pass

### Blocker: soft-Cancel double-render (confirmed, fixed)

Root cause was exactly as diagnosed: `showAgentResult` (`ShaftAssistantPanel.java`) reset `activeLocalAgentStreamToken = -1` synchronously, then the deferred `pending.whenComplete(...)` lambda re-read that same field for `currentStream`, always getting `tokenA == -1` → false → routed through `persistAndAppendResponse` (append) instead of `finishLocalAgentResponse` (replace).

**Fix — redesigned Cancel to mirror Kill's existing shape** (`ShaftAssistantPanel.java:2088-2110`):
- Cancel now renders the `_Cancelled._`/partial-buffer bubble **synchronously**, exactly like before this whole PR (matches `stopLocalAgentStreaming`'s Kill path, which never waited on the companion either).
- Immediately after that synchronous render, schedules a guarded **in-place-only upgrade**, keyed to the message index + session id captured *before* the render (never a live field re-read from inside the deferred callback).
- Generalized `scheduleKilledTerminalAnswerUpgrade`/`upgradeKilledLocalAgentMessage` → `scheduleTerminalAnswerUpgrade`/`upgradeFinalizedLocalAgentMessage` (`ShaftAssistantPanel.java:2646-2712`), shared by both Cancel and Kill.
- The 5-arg `showAgentCancelled(..., terminalAnswer)` overload is gone — dead code once the synchronous render never carries the recovered answer directly.

**New wrinkle found while fixing**: `localAgentStreamPlaceholderMessageIndex` is *not* reliably valid even when `currentStream` is true — a non-Verbose run only appends compact milestone bubbles (`appendLocalAgentOutput`, `ShaftAssistantPanel.java:2340-2350`), never touching the placeholder index, so it can still be `-1` at cancel time. Capturing it naively pointed the upgrade at a stale/invalid index. Added `nextLocalAgentStreamPlaceholderTargetIndex()` (`ShaftAssistantPanel.java:2328-2347`), which replicates `replaceLocalAgentStreamPlaceholder`'s own validity check (`index >= 0 && index < currentSessionMessageCount()`) to know whether the synchronous render is about to replace-in-place or append-fresh, *before* calling it.

### Finding 2 (timeout bound) — resolved structurally, not with `orTimeout`

Since Cancel's first render is now synchronous (same fix as the blocker), there is no longer any window where the user sees no terminal bubble while waiting on the companion — that concern is fully addressed by the redesign itself. Adding an artificial `orTimeout` on top would just add an unused magic-number knob, so I did not add one; flagging this explicitly in case you disagree and want a defensive bound anyway.

### Finding 3 (session-identity guard) — fixed

`upgradeFinalizedLocalAgentMessage` (`ShaftAssistantPanel.java:2686-2712`) now guards on `Objects.equals(active.id, sessionId)` (captured at schedule time for both Cancel and Kill), bailing before even checking the index/marker if the active session has changed. Reproduced the exact scenario you described (a different, now-active session whose message at the same index also coincidentally carries the same marker) in `assistantKillTerminalAnswerUpgradeIsSkippedIfTheUserSwitchedSessionsMeanwhile` — watched it fail (the other session's message got overwritten) before the guard, green after.

### Finding 4 (bridge.close() ordering) — fixed

`AssistantLocalAgentRunner.java:708-729`: nested the `terminalAnswerConsumer.accept(...)` call in its own `try/finally`, so `bridge.close()` always runs even if a consumer throws.

### New/updated tests (TDD: watched red against the buggy code, then green)

- `ShaftPanelSetupTest#assistantCancelUpgradesTheSameFinalizedMessageOnceTerminalAnswerArrives` (replaces the old, "blind by construction" test) — streams a line, then cancels, asserts message count unchanged after the companion resolves with a real answer.
- `ShaftPanelSetupTest#assistantCancelAfterStreamingDoesNotDuplicateWhenCompanionResolvesWithNoAnswer` — same shape, companion resolves `null`.
- `ShaftPanelSetupTest#assistantKillTerminalAnswerUpgradeIsSkippedIfTheUserSwitchedSessionsMeanwhile` — the session-identity regression described above.
- Both new Cancel tests reproduced the exact reported bug (message count 1→2, or `_Cancelled._` never rendering at all after a mis-targeted index) before the fix; all green after.

### Re-verified after this fix

- Rebased onto current `origin/main` (`a4e5b64fc6`) before pushing, per your note.
- Full `com.shaft.intellij.ui.*` package: 874 tests (2 new + 1 renamed vs. the prior 872), 0 failures, 0 errors — run twice consecutively.
- `AssistantLocalAgentRunnerTerminalAnswerCompanionTest` + `AssistantLocalAgentRunnerCancellationTest` (the #3758 invariant suite) — 3 consecutive clean reruns, confirming the `bridge.close()` reordering didn't disturb anything.
- `compileJava`/`compileTestJava` clean.

Ready for re-review.

MohabMohie:
## Fixed per second adversarial review — F1, F2/F3 addressed together (F4 already tracked as #4210, no action here)

### F1 — token-usage line restored on the Cancel upgrade path

`ShaftAssistantPanel.java` (`upgradeFinalizedLocalAgentMessage`): the upgrade wrote the recovered answer directly into `target.markdown`, bypassing `withLocalAgentTokenUsage` (which `finishLocalAgentResponse` always uses for every other terminal local-agent response, including the plain bare marker this same bubble started as). Now the composed text (answer + `_Cancelled._/_Killed._ (the run had already finished)`) is routed through `withLocalAgentTokenUsage(composed, "")` before being written in — matching the exact ordering the original 5-arg design had (answer → label trailer → token-usage line last).

### F2/F3 — took the recommended refactor: stopped predicting, started recording

Deleted `nextLocalAgentStreamPlaceholderTargetIndex()` entirely (the duplicated validity-rule copy the review flagged). Added `lastLocalAgentFinalizedMessageIndex`, stashed by:
- `finishLocalAgentResponse` — right after `replaceLocalAgentStreamPlaceholder` returns (that call already computes the correct post-trim index internally, whether it replaced in place or appended fresh), before resetting `localAgentStreamPlaceholderMessageIndex` to `-1`.
- `persistAndAppendResponse` — right after its own `append(...)` call, via `currentSessionMessageCount() - 1` (post-trim, since `append()` → `chatState.append()` → `trim()` has already run by then).

`showAgentResult`'s cancelled branch now calls `showAgentCancelled(...)` first, then reads `lastLocalAgentFinalizedMessageIndex` (and re-reads `chatState.activeSession()` for the session id) — no more beforehand prediction, so the cap-trim shift can no longer diverge from reality. This is the exact refactor you and the reviewer recommended, and it's net simpler (one fewer method, no duplicated validity check).

### New regression tests (TDD: watched red against the pre-fix code, green after)

- `ShaftPanelSetupTest#assistantCancelUpgradeStillIncludesTokenUsageStatement` (F1) — red: markdown lacked "token usage" after the upgrade landed the recovered answer; green after routing through `withLocalAgentTokenUsage`.
- `ShaftPanelSetupTest#assistantCancelUpgradeStillLandsWhenTheStreamedBubbleTripsTheSessionCap` (F2/F3) — fills the session to `ShaftAssistantChatState.MAX_MESSAGES_PER_SESSION` (500) before streaming+cancelling, reproducing the exact cap-trim shift; red: the recovered answer never landed (bare `_Cancelled._` forever, literally #3962's original symptom); green after the refactor. Also asserts the session stays capped at 500 throughout.

### Re-verified

- Rebased onto current `origin/main` (`99ae9ea74e`, after #4207) before pushing.
- All previously-added tests (blocker fix, session-identity guard, no-duplication) still green with zero changes needed.
- Full `com.shaft.intellij.ui.*` package: 876 tests (2 new vs. the prior 874), 0 failures, 0 errors — run twice consecutively.
- `compileJava`/`compileTestJava` clean.

Ready for a third pass.

MohabMohie:
## Fixed per fourth adversarial review — Fix 1 and Fix 2 both addressed

### Fix 1 — real terminalAnswer now passed to withLocalAgentTokenUsage

`ShaftAssistantPanel.java` (`upgradeFinalizedLocalAgentMessage`): was `withLocalAgentTokenUsage(composed, "")`, now `withLocalAgentTokenUsage(composed, terminalAnswer)` — the real recovered answer, which always carries its genuine `{"usage":...}` trailer (only stripped from the *displayed* text by `stripTrailingUsageMetadata`, not from the value passed for usage parsing). Every recovered-answer upgrade now renders real token counts instead of "not available."

### Fix 2/3 — Message-reference tracking instead of int index (adopted the recommended refactor)

Replaced `lastLocalAgentFinalizedMessageIndex` (int) with `lastLocalAgentFinalizedMessage` (`ShaftAssistantChatState.Message` reference). Stash points unchanged in shape, just capture a reference instead of a position:
- `finishLocalAgentResponse` — `lastLocalAgentFinalizedMessage = messageAt(localAgentStreamPlaceholderMessageIndex);` right after `replaceLocalAgentStreamPlaceholder` (which already computed the correct post-trim index internally).
- `persistAndAppendResponse` — `lastLocalAgentFinalizedMessage = messageAt(currentSessionMessageCount() - 1);` right after its own append.
- `stopLocalAgentStreaming` (Kill) — same pattern, captured into a local variable and passed straight to `scheduleTerminalAnswerUpgrade`.

`scheduleTerminalAnswerUpgrade`/`upgradeFinalizedLocalAgentMessage` now take the `Message` object directly. The upgrade does `active.messages.indexOf(message)` (identity-based — `Message` has no `equals()`/`hashCode()` override, so this is reference comparison) instead of a positional bounds check. As you predicted, this subsumed the separate session-id equality check for free — deleted that field/parameter and the now-unused `Objects` import along with it. Net: one fewer parameter on both methods, one guard clause instead of two.

### New regression tests (TDD: watched red against the pre-fix int-index code, green after)

- `ShaftPanelSetupTest#assistantCancelUpgradeStillLandsAfterAnUnrelatedAppendShiftsItsIndexAtTheCap` (2a) — fills to the 500 cap, streams+cancels, then appends one unrelated message before resolving the companion (simulating the user's next prompt). Red before the fix: the recovered answer never landed. Green after.
- `ShaftPanelSetupTest#assistantTwoCancelledRunsAtTheCapNeverCrossContaminateEachOthersBubble` (2b) — two runs cancelled back-to-back at the cap, run A's companion resolving only after run B's own cancel has already shifted things. Red before the fix (run A's answer was lost, not merely misplaced, in this exact repro — confirming the underlying defect independent of which exact failure mode showed up). Green after, with explicit assertions that each answer lands exactly once and only in its own bubble.

### Re-verified

- Rebased onto current `origin/main` (`d80e74da9d`, after #4208) before pushing.
- All previously-added Cancel/Kill tests (blocker, session-guard, no-duplication, F1/F2/F3 cap tests) still green with zero further changes needed.
- Full `com.shaft.intellij.ui.*` package: 878 tests (2 new vs. the prior 876), 0 failures, 0 errors — run twice consecutively.
- `compileJava`/`compileTestJava` clean.

Also folded in the doc-nit tidy on `lastLocalAgentFinalizedMessage`'s javadoc while rewriting it for the new type.

Ready for a hopefully-final pass.

Issue #4214, opened 2026-07-27
Tracking: SHAFT Assistant + plugin polish — remaining work (successor to #4024, #4160)


MohabMohie:
## Summary

Consolidates remaining work from two closed trackers: #4024 (SHAFT Assistant record/replay/codegen overhaul) and #4160 (production-polish: plugin setup, Assistant chat, record/codegen flow). Both trackers are closed as substantially complete; this ticket carries forward only what is still open, with explicit sequencing since several remaining items converge on the same files.

## Closed trackers, for history

- #4024 — closed. Subtasks 1-5 and 7 done and merged. Subtask 6 partially done (diagnosability half, PR #4196 merged); the remaining half lives on as #4029 below.
- #4160 — closed. Area C (record/replay/codegen) saw the most progress this session (#4166, #4188, #4194, #4029-diagnosability-half, #4202, #4209 all fixed); Area B (chat) got #3962 fixed; Area A (plugin initial setup) has not yet been audited.

### Area C — capture/codegen pipeline (shaft-capture, shaft-doctor)
- [ ] #4029 — Loud diagnosable codegen failures + replay-gated success (remaining half: gate codegen success on a confirmed replay)
- [ ] #4206 — Staged generated-source file leaks on RuntimeException from compile/replay (non-exception-safe cleanup)

### Area B — Assistant chat UX (shaft-intellij)
- [ ] #4210 — Cancelled/killed chat bubbles mutate silently when the recovered terminal answer arrives late

### Area A — plugin initial setup
- [ ] Audit not yet performed. Needs a fresh pass per #4160's original method (plugin screenshot renderer, real running surface, not source reading) before any fix work can be scoped.

### Mobile / Flutter
- [ ] #4199 — OptionsManager has no typed FlutterDriverOptions arm for APPIUM_FLUTTER sessions

### Docs
- [ ] #4211 — Stale hardcoded "89-tool catalog" prose references after MCP tool count changed

## Sequencing constraints

- **#4206 before #4029, strictly sequential.** Both touch `CaptureGenerator.generate()`'s failure-path/control-flow around the same compile/replay block. #4206 is small and mechanical (move a cleanup call into a `finally`/catch); #4029 is larger (redefines success/failure gating semantics). Landing the small fix first minimizes the rebase surface for the bigger one. Do not run these two concurrently even in separate worktrees — file-level collision, not just tree collision.
[...]

PR #4222, opened 2026-07-27
shaft-intellij: pending-answer indicator on Cancelled/Killed bubbles

### LINKED ISSUE #4210: [Enhancement] Cancelled/killed chat bubbles mutate silently when the recovered terminal answer arrives late
Deferred finding from the second adversarial review round of PR #4204 (issue #3962, cancel/kill terminal-answer recovery).

## Gap

When a soft Cancel's recovered terminal answer arrives late (the agent process is still shutting down, e.g. ignoring SIGTERM up to the full process timeout), the panel mutates the already-rendered `_Cancelled._` bubble in place with no visual indication a change is coming or has happened. During the wait window the bubble is indistinguishable from a genuinely final cancelled state. When the mutated bubble isn't the last message, the update goes through a full `transcript.setMessages(...)` resync, which can move scroll position unexpectedly.

This is the same shape Kill already ships (Kill also retroactively upgrades a `_Killed._` bubble once its companion resolves) -- not a regression introduced by PR #4204, just an already-accepted UX tradeoff being extended to a second entry point.

## Why not fixed now

Low severity, bounded (only matters if the underlying process takes a long time to actually terminate after cancel/kill), and orthogonal to the correctness work PR #4204 is actually verifying. A UI affordance ("this answer may still update") is a legitimate polish item, not a correctness fix.

## Scope for a fast-follow

Add a subtle in-place indicator (e.g. a small "recovering final answer..." caption or spinner) on a Cancelled/Killed bubble while its companion future is still pending, cleared when the upgrade lands or times out. Consider whether the full-resync-on-non-last-message scroll jump is worth guarding against separately.

Related: #3962, PR #4204.

MohabMohie:
## Summary

Closes #4210 (fast-follow from the #3962/PR #4204 hardening work, subtask of tracker #4214).

While a soft Cancel's (or Kill's) recovered terminal answer is still pending, the already-rendered `_Cancelled._`/`_Killed._` bubble now shows a subtle `_Recovering final answer..._` caption, so the user can tell an in-place change may still land instead of the bubble looking indistinguishable from a genuinely final state. The caption disappears the moment the companion future resolves — whether it lands a real answer (folded into the upgraded content), resolves with nothing (the run never produced a terminal event), or the run's own process-timeout path completes (`AssistantLocalAgentRunner#run`'s `finally` notifies the companion from every exit path, so it always eventually resolves — no separate timer needed for the "or times out" case).

**What I deliberately did NOT touch**: the identity-based upgrade mechanism from #3962/PR #4204 — `scheduleTerminalAnswerUpgrade`, `upgradeFinalizedLocalAgentMessage`, and the `lastLocalAgentFinalizedMessage` object-reference/guard clauses — is byte-for-byte unchanged. The two new methods (`schedulePendingAnswerIndicator`, `setPendingAnswerIndicatorVisible`) only *observe* the same companion `CompletableFuture` via an independent `whenComplete` callback (registered separately from the existing `thenAccept`), and mirror the exact same identity/active-session guard the upgrade method already uses, so the indicator can never mutate or render a message that has scrolled out of the active session — the same safety property the reviewed code establishes. The only two call sites touched (`showAgentResult`'s cancel branch, `stopLocalAgentStreaming`) get one extra line each: capture the pending future *before* `scheduleTerminalAnswerUpgrade` nulls it, then hand it to the new indicator scheduler — no change to their existing control flow or ordering.

**Scope call**: did not add the optional scroll-position guard for the `transcript.setMessages(...)` full resync (issue's second bullet says "your call") — kept this PR to just the indicator, per the ponytail minimal-diff constraint.

**Screenshot evidence**: added one new screenshot scenario (`intellij-plugin-assistant-cancelled-pending-answer.png`, via `ShaftPluginScreenshotRendererTest`) since this is a visible UI change (per the shaft-intellij testing convention) — inspected it: the caption renders as a legible, single-line italic aside under the terminal marker, not crowding it.

## Test plan

- [x] New tests written first and watched RED before implementation:
  - `assistantCancelledBubbleShowsPendingAnswerIndicatorWhileCompanionIsUnresolved`
  - `assistantCancelledBubblePendingAnswerIndicatorClearsWhenCompanionResolvesWithNoAnswer`
  - `assistantKilledBubbleShowsPendingAnswerIndicatorWhileCompanionIsUnresolved`
- [x] All three watched GREEN after implementation.
- [x] Full `ShaftPanelSetupTest` (includes every existing Cancel/Kill/terminal-answer-upgrade test, e.g. `assistantCancelUpgradeStillLandsAfterAnUnrelatedAppendShiftsItsIndexAtTheCap`, `assistantTwoCancelledRunsAtTheCapNeverCrossContaminateEachOthersBubble`, `assistantKillTerminalAnswerUpgradeIsSkippedIfTheUserSwitchedSessionsMeanwhile`): 275 tests, 0 failures — all unchanged and green.
- [x] `AssistantLocalAgentRunnerTerminalAnswerCompanionTest` (2 tests) and `AssistantLocalAgentRunnerCancellationTest` (3 tests): green, untouched (no changes to `AssistantLocalAgentRunner.java`).
- [x] `ShaftPluginScreenshotRendererTest` regenerated with `-Dshaft.intellij.screenshotDir=...`: 6 tests, 0 failures; new screenshot visually inspected.

https://claude.ai/code/session_01B77p5FMfqraiw1amwJGgVH

### DIFF
--- a/shaft-intellij/src/main/java/com/shaft/intellij/ui/ShaftAssistantChatState.java
+++ b/shaft-intellij/src/main/java/com/shaft/intellij/ui/ShaftAssistantChatState.java
@@ -45,6 +45,17 @@ public final class ShaftAssistantChatState implements PersistentStateComponent<S
     public static final String KIND_RAW_VERBOSE = "raw-verbose";
     public static final String KIND_MILESTONE = "milestone";
     public static final String KIND_USER = "user";
+    // Issue #4210: the pending-answer caption ShaftAssistantPanel stapled onto a live Cancelled/Killed
+    // Message#markdown is gated only by an in-memory CompletableFuture that does NOT survive a
+    // project/IDE restart -- if a save (getState(), fired by IntelliJ's periodic workspace autosave,
+    // not just an explicit user action) lands while that future is still unresolved, nothing would
+    // otherwise ever strip the caption back out, leaving a permanent, false "still recovering" claim
+    // on a run that structurally can never resolve after reload. Declared here (not in
+    // ShaftAssistantPanel) so the single literal is shared between the panel (which appends/removes it
+    // on the LIVE message while a real future is pending) and normalizeMessage below (which strips any
+    // surviving copy at every load/save round-trip, exactly like redactSecrets already does for
+    // secrets).
+    static final String PENDING_ANSWER_INDICATOR = "\n\n_Recovering final answer..._";
     private static final String SECOND_PERSON_LABEL = String.valueOf(new char[]{'Y', 'o', 'u'});
     private static final Pattern KEY_VALUE_SECRET = Pattern.compile(
             "(?iu)([\"']?\\b(?:api[_-]?key|access[_-]?token|refresh[_-]?token|token|secret|password|authorization|cookie|set-cookie)\\b[\"']?\\s*[:=]\\s*)([\"']?)[^\\s`'\",}]+([\"']?)");
@@ -300,9 +311,20 @@ public final class ShaftAssistantChatState implements PersistentStateComponent<S
         Message copy = new Message();
         copy.role = source.role == null ? "" : source.role;
         copy.kind = resolveKind(source.kind, copy.role);
-        copy.markdown = redactSecrets(source.markdown);
+        copy.markdown = redactSecrets(stripPendingAnswerIndicator(source.markdown));
         copy.createdAt = source.createdAt == null ? "" : source.createdAt;
         return copy;
+    }
+
+    /**
+     * Issue #4210: strips a surviving {@link #PENDING_ANSWER_INDICATOR} caption from a message at
+     * every load/save round-trip -- the in-memory {@code CompletableFuture} gating that caption on
+     * the live panel does not survive a project/IDE restart, so this is the only place guaranteed to
+     * run regardless of what happened to that future (mirrors {@link #redactSecrets}'s own
+     * at-load/at-save enforcement, right beside it).
+     */
+    private static String stripPendingAnswerIndicator(String value) {
+        return value.contains(PENDING_ANSWER_INDICATOR) ? value.replace(PENDING_ANSWER_INDICATOR, "") : value;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


MohabMohie:
Addressed independent-review finding: the pending-answer caption was gated only by an in-memory `CompletableFuture` that does not survive a project/IDE restart. If IntelliJ's workspace autosave (`ShaftAssistantChatState` is a `PersistentStateComponent`) fires while the companion is still unresolved, the caption would have been baked into the persisted transcript permanently -- an actively misleading regression vs. the pre-PR bare `_Cancelled._`/`_Killed._` bubble.

**Fix** (mirrors the existing `redactSecrets`-at-load pattern): the indicator literal now lives in `ShaftAssistantChatState` (shared with `ShaftAssistantPanel`, which still owns showing/clearing it on the live message while a real future is pending), and `normalizeMessage` strips any surviving copy on every `getState()`/`loadState()` round-trip -- regardless of what happened to the in-memory future.

**New regression test**: `assistantCancelledBubblePendingAnswerIndicatorDoesNotSurviveAStateRoundTripWhileCompanionIsUnresolved` -- cancels with a deliberately never-completed companion future, confirms the indicator is showing, then round-trips `getState()`/`loadState()` into a brand-new `ShaftAssistantChatState` (simulating a fresh IDE instance after restart) and asserts the caption is gone while the terminal marker survives. Watched RED first (temporarily reverted the strip), then GREEN.

**Full verification after the fix + rebase onto latest `main`** (all green, 0 failures):
- `ShaftPanelSetupTest`: 276 tests (275 previous + 1 new)
- `AssistantLocalAgentRunnerTerminalAnswerCompanionTest`: 2
- `AssistantLocalAgentRunnerCancellationTest`: 3
- `ShaftAssistantChatStateTest`: 9
- `ShaftPluginScreenshotRendererTest`: 6 (screenshots regenerated, visually reinspected)

Rebased cleanly onto `main` (no conflicts), pushed. Not merging, not watching CI further.

Issue #4262, opened 2026-07-27
McpPlaywrightCaptureAdapter: manual-recording codegen now 100% hard-fails after #4252's locator ladder gate


MohabMohie:
## Summary

PR #4252 (commit d307e61172, "capture: make locator policy a real gate, hard ban non-ARIA fallback") turned issue #4239's rung-1/rung-2 locator ladder into a hard gate: `CaptureGenerator.isLadderEligible` (shaft-capture/src/main/java/com/shaft/capture/generate/CaptureGenerator.java:744-750) only admits a ROLE candidate with `roleXpathVerified()==true` (rung 1) or any candidate with a non-blank self-verified `replayXpath()` (rung 2). Every other strategy hard-fails generation via `state.unsupported()`.

`McpPlaywrightCaptureAdapter.locator()` (shaft-mcp/src/main/java/com/shaft/mcp/McpPlaywrightCaptureAdapter.java:237-248) builds every manually-recorded Playwright locator (ID/NAME/CSS/XPATH tagged `LocatorSignal.USER_PROVIDED`) with neither signal — there is no live-DOM self-verification step in that manual-recording flow. **Every Playwright manual-recording codegen call now hard-fails.** Confirmed regression: `PlaywrightServiceTest` passed at #4243 (79b6b970fa), fails on current main.

## Why this isn't a one-line fix

A naive fix (widen `isLadderEligible` to also trust `LocatorSignal.USER_PROVIDED`) is insufficient and wrong: `CaptureGenerator.locatorExpression` (CaptureGenerator.java:1636-1644) still renders ID/NAME/CSS candidates as literal `SHAFT.GUI.Locator.id/name/cssSelector(...)`, which trips the separate, unconditional `NON_ARIA_LOCATOR` guardrail (GeneratedCodeGuardrails.java:26-27, wired in as an ERROR at CaptureGenerator.java:310) — no ladder-eligibility escape hatch there either. So even with the ladder gate widened, generation still fails.

[...]

Issue #4239, opened 2026-07-27
Capture record-to-codegen pipeline: locator policy is advisory not enforced, the delivered Assistant prompt contradicts it, and mobile reports unverified output as successful (+ architecture plan)


MohabMohie:
End-to-end play-test of the record → replay → codegen pipeline across all three surfaces (web / API / mobile), evaluated as a real user driving it through the IntelliJ Assistant + MCP, plus an architecture plan to close the gaps found.

Scope note: this is an investigation + design ticket. No code was changed. Related policy origin: #4024. Baseline already landed and **verified still holding**: #4029 (web `UNCONFIRMED`), #4220 (API `UNCONFIRMED`), #4230/#4231 (Playwright replay honesty), #4199, #4209/#4212/#4215, #3962/#4210/#4232.

## Method — what was actually executed vs. read

**Executed** (real code, real Maven run, headless):
- Drove the real `CaptureGenerator` end-to-end from a hand-built recording through a throwaway probe: `mvn -o -pl shaft-capture test -Dtest=PolicyProbeTest -DheadlessExecution=true` → 1/1 passed, generated source captured. Probe deleted afterwards; it was an observation harness, not a regression test.

**Read only** (stated plainly where a finding rests on code-reading rather than execution):
- MCP tool annotations/schemas in `CaptureService`/`MobileService`/`PlaywrightService`.
- The IntelliJ Assistant's runtime prompt prefixes in `AssistantCommand.java`.
- `shaft-skills/` bodies and their (non-)delivery path.
- Mobile codegen internals. Mobile could **not** be executed: it needs a live Appium/emulator session, which repo policy forbids launching here. Mobile findings below are code-reading, and labelled as such.

---

# Findings

## F1 — CRITICAL: the runtime prompt the Assistant actually receives tells the LLM to **prefer Smart Locators**

The owner policy is "never a Smart Locator". The prompt prefix that is prepended to **every** Assistant request says the opposite, in two places:

- `shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantCommand.java:99` (`SHAFT_MCP_USAGE_HINT`, delivered unconditionally):
  > "Never generate SHAFT.GUI.Locator.xpath(...); **use smart locators**, the SHAFT locator builder, or By.xpath only as a last fallback."
- `AssistantCommand.java:121` (`SHAFT_CODEGEN_TOOL_GUIDANCE`):
[...]

PR #4263, opened 2026-07-27
Fix red main: CaptureServiceTest review-warning fixture pre-dates the #4239 P1.4a locator ladder

MohabMohie:
## Summary

`CaptureServiceTest`'s two failures (`codeBlocksToolReturnsDeterministicReviewWarnings`, `recordAtTargetCodeBlocksReturnFocusedInsertionBlocksInsideWorkspace`) reported against PR #4254 are **not** caused by that PR — they reproduce identically on a clean `origin/main` checkout with zero code changes. Root cause: #4252's rung-1/rung-2 locator emission ladder in `CaptureGenerator` (shaft-capture) now hard-fails generation for any target lacking self-verified ARIA-role/XPath evidence, but #4252 only updated shaft-capture's own tests — `CaptureServiceTest.reviewWarningSession()`'s `pay-button` fixture (shaft-mcp) was never updated and still uses a plain, unverified `XPATH` candidate.

- Gave the candidate a self-verified `replayXpath` (clears ladder rung 2).
- Changed the expression from an absolute path (`/html/...`) to a relative one (`//div[3]/...`) — `GeneratedCodeGuardrails`' `ABSOLUTE_XPATH` rule is an unconditional ERROR independent of ladder eligibility, so an absolute self-verified XPath would still hard-fail.
- Kept the `[3]` index so `CaptureGenerator`'s `INDEXED_LOCATOR` brittleness check still fires — the fixture still exercises the `review/LOCATOR` warning it was written to test.

`PlaywrightServiceTest`'s 3 separate failures share the same root commit (#4252) but are a distinct, unresolved architecture question (should `McpPlaywrightCaptureAdapter`'s USER_PROVIDED manually-recorded locators also clear the ladder?) — tracked in #4262, deliberately left red here per that issue.

## Test plan
- [x] Reproduced all 5 original failures on a clean `origin/main` worktree (0 diff) — confirms this is a pre-existing red-main regression, not something #4254 introduced.
- [x] `mvn -pl shaft-mcp test -Dtest=CaptureServiceTest -DheadlessExecution=true` — 23/23 green after the fix.
- [x] `mvn -pl shaft-mcp test -DheadlessExecution=true` (full module) — 486 tests, only the 3 known/out-of-scope `PlaywrightServiceTest` failures (#4262) remain; no other regressions.

Refs #4239, #4262.

https://claude.ai/code/session_019wEoJduy4Q4jfHWejcizCe

### DIFF
--- a/shaft-mcp/src/main/resources/application-http.properties
+++ b/shaft-mcp/src/main/resources/application-http.properties
@@ -13,6 +13,23 @@ server.port=${PORT:8081}
 spring.ai.mcp.server.streamable-http.mcp-endpoint=/mcp
 spring.ai.mcp.server.sse-endpoint=/mcp
 
+# Delivered to every MCP client for free in the initialize response (issue #4239 P2.1): the locator
+# policy, the record -> confirm-replay -> generate -> verify ordering, per-surface tool routing, and
+# how to read successful/status. Keep this in sync with shaft-skills/choosing-shaft-locators,
+# recording-shaft-tests-with-mcp, writing-shaft-tests, and act-as-shaft-dev.
+spring.ai.mcp.server.instructions=SHAFT MCP server. Locator policy: build every generated locator \
+through the SHAFT locator builder's ARIA role (hasRole(...)), chained with hasText/hasAttribute/context \
+until it uniquely matches; fall back to native By.xpath(...) only when the target exposes no ARIA \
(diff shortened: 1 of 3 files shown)


MohabMohie:
Superseded — the fix landed via #4254 (which merged in both this branch and #4265's combined fix before merging itself, per the owner's requested merge order). Rebasing this branch onto the new `origin/main` produced a no-op diff (0 commits ahead), and GitHub auto-closed the PR on push. No separate merge needed; closing the loop here for the paper trail.

PR #4326, opened 2026-07-28
Fix #4321: self-heal MultipleElementsFoundException when exactly one match is actionable

### LINKED ISSUE #4321: shaft-mcp: element action tools should self-heal MultipleElementsFoundException via the unified locator algorithm
## Reproduction (owner, live session)

An `mcp__shaft-mcp__element_type` tool call failed during replay:

\`\`\`
Tool failed (mcp__shaft-mcp__element_type): Root cause: "com.shaft.gui.internal.exceptions.MultipleElementsFoundException:
Build info: version: '4.46.0', revision: 'df5a634...'"
Caused by: com.shaft.gui.internal.exceptions.MultipleElementsFoundException: ...
System info: os.name: 'Windows 11', os.arch: 'amd64', os.version: '10.0', java.version: '25.0.1'
\`\`\`

## Ask

"This happens when the locator is not unique. We should be able to identify this exception and handle it properly by relocating the element using the new centralized element identification algorithm. The algorithm should ensure that there are exactly ONE matches for any locator before we use it, and it should keep trying until the element identification timeout is reached or until the element is uniquely identified."

So: when a runtime element-action tool call hits `MultipleElementsFoundException`, don't just fail -- catch it, refine/relocate to a locator that matches exactly one element (reusing the existing centralized locator algorithm rather than inventing a new one), and retry until either uniquely resolved or the element-identification timeout elapses.

## Relevant existing code (scouted, not yet deeply investigated)

- **Single choke point for every element action tool**: `shaft-mcp/src/main/java/com/shaft/mcp/ElementService.java:527` (`withLocator`, private, used by `type`/`click`/etc. — grep the file for all callers) wraps every locator-based operation in one `try { locatorOperation.execute(getLocator(...)) } catch (Exception exception) { ...; throw exception; }`. This is the natural single interception point rather than patching each of the ~10 tool methods separately.
- **"The new centralized element identification algorithm"** is very likely `shaft-capture/src/main/java/com/shaft/capture/generate/LocatorPolicy.java` (PR #4277, "capture: unified deterministic locator scoring/ranking algorithm", issue #4271) — but that was built for **capture-time** codegen (choosing/ranking a locator to bake into generated test source from a DOM snapshot), not runtime re-identification during replay. Confirm whether it's actually reusable at runtime as-is, needs adaptation, or whether "centralized algorithm" refers to something else already used at runtime (check `shaft-mcp/src/main/java/com/shaft/mcp/BrowserService.java`, which also references locator-policy-flavored logic per an earlier grep hit).
- SHAFT core's own uniqueness check (`Actions.java`, referenced in the investigation for issue #4300): `foundElements.get().size() > 1 && SHAFT.Properties.flags.forceCheckElementLocatorIsUnique()` throws `MultipleElementsFoundException` — this is core engine behavior, not shaft-mcp; confirm whether the fix belongs at the shaft-mcp tool layer only (catch-and-retry-with-a-refined-locator) or needs a core-engine hook too.
- An existing element-identification timeout property should already exist in `SHAFT.Properties.timeouts` (or similar) -- reuse it, don't invent a new config knob, unless investigation shows none fits this purpose.

## Why this needs real design work, not just a try/catch

- "Relocating the element" implies generating/selecting a *more specific* locator than the one that just failed -- that requires either (a) inspecting the current DOM/element tree at the moment of failure to find a disambiguating attribute/ancestor, or (b) some other concrete strategy. Whichever `LocatorPolicy` (or equivalent) actually does needs to be confirmed capable of this at runtime with only a `locatorStrategy`/`locatorValue` pair and a live driver session -- not assumed.
- "Keep trying until timeout" implies a real retry loop with real wall-clock timeout tracking, not a fixed retry count.
- This is a cross-cutting change likely touching `shaft-mcp` (the catch/retry orchestration) and possibly `shaft-capture` (if `LocatorPolicy` needs a runtime-callable API it doesn't have yet) and/or `shaft-engine` core (if the uniqueness check/exception itself needs to expose more disambiguation context). Scope precisely before implementing broadly.

## Acceptance
- `MultipleElementsFoundException` from any element-action MCP tool is caught (ideally once, at the `withLocator` choke point) and triggers a real relocate-and-retry loop using the actual centralized locator algorithm (not a bespoke new one), bounded by the existing element-identification timeout.
- If uniquely resolved within the timeout, the tool call succeeds using the refined locator.
- If the timeout elapses without a unique match, the tool fails with a clear, actionable error (not a bare stack trace) explaining that no unique element could be identified.
- TDD: a focused test reproducing "locator matches >1 element, a refinement exists that narrows to exactly 1" (green) and "no refinement narrows it before timeout" (clean, actionable failure) -- scoped, no full suite.

MohabMohie:
## Summary

Fixes #4321. When a locator matches more than one element, `Actions.performAction()`
already retries for the full identification timeout (`SHAFT.Properties.timeouts.defaultElementIdentificationTimeout()`,
default 10s) before failing -- `MultipleElementsFoundException` was already in
`SynchronizationManager`'s `FluentWait` ignore-list. But every retry re-ran the
*same unchanged locator*, so a genuinely-resolvable ambiguity (e.g. one hidden/disabled
decoy element alongside the real target) always ended up throwing anyway.

This PR adds **last-resort-only** disambiguation: after the full identification
timeout is exhausted and the locator is *still* ambiguous on every unmodified retry,
SHAFT makes exactly one additional attempt, narrowing to the single element that is
both displayed and enabled -- but only if exactly one qualifies. Any other split
(zero, or two-or-more, displayed-and-enabled matches) still throws
`MultipleElementsFoundException` exactly as before.

**This design went through two rounds of independent architectural review** (this
touches `Actions.java`, the highest-blast-radius file in the framework). The first
round caught a real bug in an earlier version of this PR that narrowed *during* the
retry loop instead of only as a last resort -- that version could silently resolve
on the very first poll and click the wrong element (e.g. a genuinely enabled
"Cancel" button beside a still-initializing, currently-disabled "Save") while
reporting PASS. See the PR comment history for the full back-and-forth; the design
below is what survived both passes.

## What's implemented

- `Actions.uniqueDisplayedAndEnabledElement(List<WebElement>)` (`Actions.java:1847`)
  -- returns the single element when exactly one of the candidates is displayed AND
  enabled, reusing the same filtering idiom as the existing
  `chooseBestEffortDisplayedElement` (generalized from its previous mobile-only
  scope, not new scoring logic).
- The primary `FluentWait` retry loop is **behaviorally unchanged**: both of its
  `throw new MultipleElementsFoundException()` sites (`Actions.java:515-543` initial
  lookup, `Actions.java:651-671` post-lazy-load refreshed lookup) always throw during
  every poll, gated by a `lastResortNarrowingAllowed` flag that stays `false` for the
  whole loop -- identical to pre-#4321 behavior, so a still-settling page always gets
  its full identification-timeout window.
- `Actions.isCausedByAmbiguousLocator(Throwable)` (`Actions.java:1876`) -- walks the
  full cause chain (not a fixed wrapping depth) to detect when the primary loop's
  `TimeoutException` was ultimately caused by `MultipleElementsFoundException`.
- When that's true, `performAction()` (`Actions.java:950-963`) makes exactly one
  additional attempt with narrowing enabled, implemented by naming the existing
  `FluentWait` lambda (`attemptElementActionOnce`, declared at `Actions.java:495`) so
  it's callable a second time -- zero duplication of the ~400-line action-dispatch
  switch statement.
- A `ReportManager.logDiscrete(...)` trace line fires at the point narrowing
  actually resolves an ambiguity, naming the original match count and that
  auto-resolution happened after the identification timeout -- so a real ambiguity
  is visible in the report instead of silently looking like an ordinary pass.

## Tests (all in `shaft-engine`, real headless Chrome via `TestPageServer` unless noted)

- `MultipleElementsFailureTest` (7 tests):
  - `type` / `click` / `clickUsingJS` -- pre-existing regression guard (3 plain,
    all-actionable `<input>`s): still throw `MultipleElementsFoundException`
    unchanged. Assertions strengthened to check the cause chain specifically
    (`assertThrowsMultipleElementsFoundException` helper) instead of the too-broad
    `expectedExceptions = {RuntimeException.class}`.
  - `typeResolvesToTheOnlyVisibleAndEnabledElementAmongMultipleMatches` /
    `clickResolvesToTheOnlyVisibleAndEnabledElementAmongMultipleMatches` -- 3
    `<input>`s (1 hidden, 1 disabled, 1 actionable): action succeeds.
  - `clickResolvesToTheOnlyEnabledButtonOnlyAfterTheIdentificationTimeoutIsExhausted`
    -- 2 `<button>`s (one permanently disabled, one permanently enabled): asserts
    elapsed time proves the narrowing did *not* fire on the first poll, and asserts
    an `onclick`-marker (`document.title`) proves *which* button was actually
    clicked (a disabled button ignores click events entirely, so a silent no-op on
    the wrong element would leave the marker unset and fail this assertion).
  - `staysAmbiguousWhenTheSecondMatchBecomesActionableBeforeTheIdentificationTimeoutElapses`
    -- 2 `<button>`s, one starts disabled and is JS-enabled mid-window: proves the
    locator stays genuinely ambiguous and throws, never silently resolving to the
    wrong button while the real target was still initializing.
- `ActionsCoverageUnitTest` (43 tests, mocked `WebElement`/`WebDriver`, no browser):
  unit coverage for `uniqueDisplayedAndEnabledElement` and the new
  `isCausedByAmbiguousLocator` cause-chain walker.
- `ElementsHelperCoverageUnitTest` (21 tests): adjacent element-lookup coverage,
  unaffected.

All classes run isolated (own Surefire fork) and headless
(`-DheadlessExecution=true -Dallure.automaticallyOpen=false`); every new/changed
test was watched RED (against the pre-fix code, for the exact expected reason) then
GREEN.

## Scope decisions

- **Why not shaft-heal / `HealingManager`?** Investigated first. It's fingerprint/
  history-based: it only fires on the zero-match branch and requires a previously-
  observed unique fingerprint, so it cannot help a locator that has never uniquely
  resolved before (#4321's literal reported scenario). Extending it to the
  multi-match branch for the *different* case of DOM-drift-into-ambiguity-with-
  prior-history is tracked separately: #4327.
- **Why not shaft-capture's `LocatorPolicy`?** Confirmed capture-time-only
  (`ElementSnapshot`/`LocatorCandidate` evidence), not callable at runtime with just
  a live driver + locator.
- **Why not `touch().tap()` / `ElementActionsHelper.identifyUniqueElement`?**
  Genuinely separate element-resolution mechanism (`waitForElementPresence`, its own
  retry/timeout implementation, no shared code with the `FluentWait`-based fix here).
  Tracked separately, confirmed with the orchestrator: #4332.
- **Deferred, tracked separately, not fixed in this PR:**
  - #4327 -- `HealingManager.resolve()` on the multi-match branch (see above).
  - #4332 -- `touch().tap()` disambiguation (see above).
  - #4334 -- the last-resort path drops the original `TimeoutException`'s "tried for
    N second(s)" diagnostic context when it also fails.
  - #4335 -- docs (`webConfig.md`) need a note that `forceCheckElementLocatorIsUnique`
    is now "last resort after timeout", not absolute.
  - #4336 -- `MultipleElementsFailureTest` isn't wired into `pr-gate.yml` (only the
    broader, less-frequent `e2eTests.yml`); `ActionsCoverageUnitTest` /
    `ElementsHelperCoverageUnitTest` already are.

## Known cosmetic issue (disclosed, not fixed)

The `FluentWait` lambda body in `performAction()` is indented one level shallower
than ideal after being extracted into a named `Function` variable (it used to sit
two nesting levels deep inside `try { ... .until(d -> {`; the declaration is now one
level up). Left as-is rather than risk a blind bulk reindent on this specific file
under time pressure -- it contains multi-line Java text blocks (`"""..."""` for
inline JS) where a naive whitespace-stripping pass could shift string content. Happy
to do a careful manual pass if requested.

## Review status

This PR has been through **two rounds of independent Opus architectural review**
given the blast radius of `Actions.java`. The first round blocked the initial
in-loop-narrowing design (real bug: could resolve prematurely and act on the wrong
element). The second round approved the reworked last-resort design with the three
nits addressed in the latest commits (last-resort logging, this PR body, the
strengthened button-click assertion). **Do not auto-merge without confirming the
current review round's outcome** -- check the PR comment thread for the latest
status before arming.

### DIFF
--- a/shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java
+++ b/shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java
@@ -59,6 +59,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -506,8 +507,19 @@ public class Actions extends ElementActions {
                     throw new NoSuchElementException("Cannot locate an element using " + JavaHelper.formatLocatorToString(locator));
 
                 // ensure element locator is unique if applicable
-                if (foundElements.get().size() > 1 && SHAFT.Properties.flags.forceCheckElementLocatorIsUnique() && !(locator instanceof RelativeLocator.RelativeBy) && !(locator instanceof ByAll))
-                    throw new MultipleElementsFoundException();
+                if (foundElements.get().size() > 1 && SHAFT.Properties.flags.forceCheckElementLocatorIsUnique() && !(locator instanceof RelativeLocator.RelativeBy) && !(locator instanceof ByAll)) {
+                    // issue #4321: a locator matching several elements is only genuinely ambiguous when
+                    // more than one of them is actually actionable. When exactly one is displayed and
+                    // enabled (the rest are hidden/disabled decoys -- e.g. duplicated markup for
+                    // different breakpoints), narrow to that single element instead of failing; any
+                    // other split (0, or 2+, displayed-and-enabled matches) still throws exactly as before.
+                    Optional<WebElement> uniqueActionable = uniqueDisplayedAndEnabledElement(foundElements.get());
+                    if (uniqueActionable.isPresent()) {
+                        foundElements.set(List.of(uniqueActionable.get()));
+                    } else {
+                        throw new MultipleElementsFoundException();
+                    }
+                }
 
                 if (healingResolution.get() == null) {
                     HealingManager.observe(
@@ -618,7 +630,14 @@ public class Actions extends ElementActions {
                             && SHAFT.Properties.flags.forceCheckElementLocatorIsUnique()
                             && !(locator instanceof RelativeLocator.RelativeBy)
                             && !(locator instanceof ByAll)) {
-                        throw new MultipleElementsFoundException();
+                        // issue #4321: same narrowing as the initial lookup above -- only auto-resolve
+                        // when exactly one refreshed match is actually actionable.
+                        Optional<WebElement> uniqueActionable = uniqueDisplayedAndEnabledElement(foundElements.get());
+                        if (uniqueActionable.isPresent()) {
+                            foundElements.set(List.of(uniqueActionable.get()));
+                        } else {
+                            throw new MultipleElementsFoundException();
+                        }
                     }
                 }
                 if (lazyLoadingActivityObserved && !isMobileNativeExecution && shouldCheckNativeActionability(action, locator)) {
@@ -1766,6 +1785,37 @@ public class Actions extends ElementActions {
             }
         }
         return false;
+    }
+
+    /**
+     * Finds the single actionable (displayed and enabled) element among {@code elements}, reusing
+     * the same per-element {@code isDisplayed()}/{@code isEnabled()} filtering already applied by
+     * {@link #chooseBestEffortDisplayedElement(List)}. Unlike that best-effort picker, this method
+     * never guesses: it returns a match only when EXACTLY ONE element qualifies, so a locator that
+     * matches several genuinely actionable elements stays ambiguous (issue #4321) -- disambiguation
+     * is intentionally narrow, not a general-purpose "pick one" fallback.
+     *
+     * @param elements candidate elements matched by the target locator
+     * @return the single displayed-and-enabled element, or empty when zero or several qualify
+     */
+    static Optional<WebElement> uniqueDisplayedAndEnabledElement(List<WebElement> elements) {
+        if (elements == null || elements.isEmpty()) {
+            return Optional.empty();
+        }
+        WebElement match = null;
+        for (WebElement element : elements) {
+            try {
+                if (element.isDisplayed() && element.isEnabled()) {
+                    if (match != null) {
+                        return Optional.empty();
+                    }
+                    match = element;
+                }
+            } catch (WebDriverException ignored) {
+                // try next match
+            }
+        }
+        return Optional.ofNullable(match);
     }
 
     private static NoSuchElementException createDragAndDropElementNotFoundException(By locator, String role, NoSuchElementException rootCauseException) {
--- a/shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java
+++ b/shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java
@@ -522,10 +522,18 @@ public class Actions extends ElementActions {
                     // only when exactly one match is displayed and enabled (the rest are hidden/disabled
                     // decoys -- e.g. duplicated markup for different breakpoints); any other split (0, or
                     // 2+, displayed-and-enabled matches) still throws.
+                    int matchCountBeforeNarrowing = foundElements.get().size();
                     Optional<WebElement> uniqueActionable = lastResortNarrowingAllowed.get()
                             ? uniqueDisplayedAndEnabledElement(foundElements.get())
                             : Optional.empty();
                     if (uniqueActionable.isPresent()) {
+                        // issue #4321: a real ambiguity was just auto-resolved -- trace it so a user
+                        // whose locator matches more than one element notices it, instead of the action
+                        // silently looking like an ordinary pass.
+                        ReportManager.logDiscrete("Locator " + JavaHelper.formatLocatorToString(locator)
+                                + " matched " + matchCountBeforeNarrowing
+                                + " elements; auto-resolved to the only displayed and enabled one after the"
+                                + " identification timeout was exhausted.");
                         foundElements.set(List.of(uniqueActionable.get()));
                     } else {
                         throw new MultipleElementsFoundException();
@@ -647,10 +655,16 @@ public class Actions extends ElementActions {
                         // issue #4321: same last-resort-only narrowing as the initial lookup above --
                         // only auto-resolve when exactly one refreshed match is actually actionable, and
                         // only on the single post-timeout attempt (see comment on the initial lookup).
+                        int refreshedMatchCountBeforeNarrowing = foundElements.get().size();
                         Optional<WebElement> uniqueActionable = lastResortNarrowingAllowed.get()
                                 ? uniqueDisplayedAndEnabledElement(foundElements.get())
                                 : Optional.empty();
                         if (uniqueActionable.isPresent()) {
+                            ReportManager.logDiscrete("Locator " + JavaHelper.formatLocatorToString(locator)
+                                    + " matched " + refreshedMatchCountBeforeNarrowing
+                                    + " elements after lazy-loading settled; auto-resolved to the only"
+                                    + " displayed and enabled one after the identification timeout was"
+                                    + " exhausted.");
                             foundElements.set(List.of(uniqueActionable.get()));
                         } else {
                             throw new MultipleElementsFoundException();
--- a/shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java
+++ b/shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java
@@ -485,9 +485,14 @@ public class Actions extends ElementActions {
         AtomicReference<HealingResolution> healingResolution = new AtomicReference<>();
         AtomicReference<HealingResolution> secondaryHealingResolution = new AtomicReference<>();
         AtomicReference<By> secondaryLocator = new AtomicReference<>();
+        // issue #4321 (second-pass review): disambiguation must be a strict LAST RESORT, applied only
+        // once the full identification-timeout retry window below is exhausted -- never during the
+        // loop itself, so a still-settling page always gets its full window before any narrowing is
+        // even considered. False for every poll of the primary fluentWait call; flipped true only for
+        // the single extra attempt made after that call times out with an ambiguous locator.
+        AtomicBoolean lastResortNarrowingAllowed = new AtomicBoolean(false);
 
-        try {
-            new SynchronizationManager(driverFactoryHelper.getDriver()).fluentWait(true).until(d -> {
+        Function<WebDriver, Boolean> attemptElementActionOnce = d -> {
                 flakeProfile.waitLoops.incrementAndGet();
                 flakeProfile.locatorLookups.incrementAndGet();
                 // find all elements matching the target locator
@@ -509,11 +514,17 @@ public class Actions extends ElementActions {
                 // ensure element locator is unique if applicable
                 if (foundElements.get().size() > 1 && SHAFT.Properties.flags.forceCheckElementLocatorIsUnique() && !(locator instanceof RelativeLocator.RelativeBy) && !(locator instanceof ByAll)) {
                     // issue #4321: a locator matching several elements is only genuinely ambiguous when
-                    // more than one of them is actually actionable. When exactly one is displayed and
-                    // enabled (the rest are hidden/disabled decoys -- e.g. duplicated markup for
-                    // different breakpoints), narrow to that single element instead of failing; any
-                    // other split (0, or 2+, displayed-and-enabled matches) still throws exactly as before.
-                    Optional<WebElement> uniqueActionable = uniqueDisplayedAndEnabledElement(foundElements.get());
+                    // more than one of them is actually actionable. This ALWAYS throws during the normal
+                    // retry loop (lastResortNarrowingAllowed is false here on every poll), identical to
+                    // pre-#4321 behavior -- a still-settling page keeps its full identification-timeout
+                    // window. Only the single last-resort attempt made after that window is exhausted
+                    // (see the outer try/catch around the fluentWait call below) allows narrowing, and
+                    // only when exactly one match is displayed and enabled (the rest are hidden/disabled
+                    // decoys -- e.g. duplicated markup for different breakpoints); any other split (0, or
+                    // 2+, displayed-and-enabled matches) still throws.
+                    Optional<WebElement> uniqueActionable = lastResortNarrowingAllowed.get()
+                            ? uniqueDisplayedAndEnabledElement(foundElements.get())
+                            : Optional.empty();
                     if (uniqueActionable.isPresent()) {
                         foundElements.set(List.of(uniqueActionable.get()));
                     } else {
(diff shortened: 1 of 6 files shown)


MohabMohie:
## Independent architectural review: BLOCK

Requested a second, independent adversarial pass on this PR before arming auto-merge, given it touches `Actions.java` (highest blast-radius file in the framework). Verdict: **BLOCK, not safe to merge as-is.** Reproduced with two throwaway fixtures (not pushed):

### Blocking: silent wrong-element click on a common, realistic DOM shape

Fixture: `<button id=save disabled>` + `<button id=cancel>`, broad locator `//button`.
- Pre-PR (`Actions.java` reverted to parent commit): throws `MultipleElementsFoundException` — correct, loud, safe.
- This PR: silently narrows to the one displayed+enabled button and **clicks Cancel**, reporting PASS.

That's a "Save disabled until form valid, Cancel next to it" form — an extremely common pattern, and exactly the kind of broad, auto-derived locator MCP tool replay (this issue's reporter) tends to produce. Converting a loud ambiguity error into a silently-wrong destructive click is a strictly worse failure mode than what existed before.

Second fixture (same buttons, `#primary` enabled by JS at t=2.5s) shows the narrowing resolves on the **first poll**, before the 10s identification timeout window is ever used — inverting what #4321 actually asked for ("keep retrying until exactly one match is found **or the timeout elapses**"). It turns a deterministic error into a timing-dependent wrong click.

No escape hatch either: `forceCheckElementLocatorIsUnique` defaults `true`, and setting it `false` disables the uniqueness check entirely rather than restoring today's "throw on ambiguity" behavior — after this PR there's no configuration that gets back today's strict, safe default.

### Concrete fix direction

Apply the displayed+enabled narrowing only as a **last resort**, after the identification timeout is exhausted — not on every poll. Leave the in-loop throw untouched; catch the outer timeout-with-`MultipleElementsFoundException`-cause in `performAction` and do exactly one narrow-and-retry pass there. This preserves the full settle/retry window (kills the t=2.5s fixture entirely) and narrows the remaining risk window to "still ambiguous after the full timeout" instead of "ambiguous on the very first poll."

### Non-blocking, patch alongside

1. `ElementActionsHelper.identifyUniqueElement` (the *second*, separate element-identification path used by `touch().tap()`) is untouched — after this PR, `element().click()` succeeds on an ambiguous locator but `touch().tap()` still hard-fails on the identical DOM. #4321 asked for the centralized algorithm; only one of the two paths moved. Please assess in-PR whether this belongs in the same fix or should be its own follow-up issue.
2. Regression guard is weaker than intended: `MultipleElementsFailureTest`'s `expectedExceptions = {RuntimeException.class}` passes on *any* runtime failure (fixture 404, driver hiccup, etc.), not specifically on `MultipleElementsFoundException`. Please assert the specific exception is in the cause chain.
3. `HealingManager.observe()` only records history when exactly one element resolves — after narrowing happens *before* the heal-history call, an ambiguous locator can now seed a heal fingerprint it could never earn before. Worth a cross-link note on #4327 (deferred option A), not blocking here.
4. Failure-path cost: `uniqueDisplayedAndEnabledElement` does up to 2N remote round-trips per 100ms poll — a broad locator matching hundreds of elements makes the failure path much slower on a real grid. Low priority, just flag it in the PR body if not addressed.

Evidence: `Actions.java:510-522,619,629-641,1758,1801-1818`; `SynchronizationManager.java:21,30,83`; `Flags.java:48-50`; `ElementActionsHelper.java:788-818`; `HealingManager.java:113-117`; `MultipleElementsFailureTest.java:21,27,33`.

MohabMohie:
Reworked per the independent review. All required items addressed in commit 7aebfe24b7:

## Required fix: last-resort-only narrowing

Confirmed the bug by reproducing both your fixtures locally (recreated as permanent regression tests, see below) against the old code before touching anything:
- Static disabled-Save/enabled-Cancel: old code resolved in ~530ms (should take the full timeout).
- JS-delayed-enable: old code clicked #cancel silently and reported PASS while #save was still disabled -- exactly the dangerous case you found.

Fix: the primary `fluentWait` retry loop is now byte-for-byte unchanged in behavior (its two `throw new MultipleElementsFoundException()` sites always throw during every poll, gated by a `lastResortNarrowingAllowed` flag that is `false` for the whole loop -- identical to pre-#4321). Only when that loop's `TimeoutException` is caught AND `Actions.isCausedByAmbiguousLocator(...)` finds `MultipleElementsFoundException` anywhere in its cause chain does `performAction()` make exactly one additional attempt with narrowing enabled. No duplication of the ~400-line action-dispatch body: the existing fluentWait lambda is now a named `Function<WebDriver, Boolean> attemptElementActionOnce`, invoked once via `.until(...)` (the normal retry loop) and, on that specific failure mode, invoked one more time directly.

New regression tests (`MultipleElementsFailureTest`, real headless Chrome via `TestPageServer`, not your throwaway worktree's fixtures but recreated from your description):
- `clickResolvesToTheOnlyEnabledButtonOnlyAfterTheIdentificationTimeoutIsExhausted` -- asserts elapsed time to prove no premature resolution.
- `staysAmbiguousWhenTheSecondMatchBecomesActionableBeforeTheIdentificationTimeoutElapses` -- asserts `MultipleElementsFoundException` is thrown, proving the t=2.5s-style case (both matches become actionable before timeout) no longer silently clicks the wrong element.

Both watched RED against the pre-rework code (documented failure messages: "resolved after only 533ms" and "Expected RuntimeException to be thrown, but nothing was thrown"), then GREEN after the fix. Full re-run: `MultipleElementsFailureTest` 7/7, `ActionsCoverageUnitTest` 43/43 (added a unit test for the new `isCausedByAmbiguousLocator` helper too), `ElementsHelperCoverageUnitTest` 21/21, all isolated per-class runs, all headless.

No new escape-hatch config needed -- `forceCheckElementLocatorIsUnique=false` still disables the whole uniqueness check exactly as before (unrelated to this change), and there was never a request for a *different* escape hatch once the last-resort design is correct, since the design itself no longer has the premature-resolution failure mode you found.

## Strengthened assertions

All 5 exception-expecting tests in `MultipleElementsFailureTest` (the 3 original + the new "stays ambiguous" one) now use `assertThrowsMultipleElementsFoundException(...)`, a helper that walks the actual cause chain and asserts `MultipleElementsFoundException` specifically, instead of the too-broad `expectedExceptions = {RuntimeException.class}`.

## HealingManager / #4327 cross-link comment

Added at `Actions.java` right above the `HealingManager.observe(...)` call in the primary lookup branch: a last-resort-narrowed single element now feeds into `observe()` exactly like a naturally-unique match, so it can seed heal history an ambiguous locator couldn't have earned before #4321 -- flagged there for whoever picks up #4327.

## Your call: `ElementActionsHelper.identifyUniqueElement` / `touch().tap()`

Recommending **separate follow-up issue**, not this PR. Reasoning:
- It's a genuinely different code path -- `getMatchingElementsInformation`/`waitForElementPresence` has its own retry/timeout mechanics entirely separate from `Actions.performAction`'s `FluentWait` loop. There's no shared mechanism to extend; it would need its own from-scratch "last resort after this path's own timeout" design, not a reuse of what's in this PR.
- It touches a different file/area (`TouchActions.java`, 14+ call sites: tap, scroll, drag, wheel-scroll) plus `ElementActionsHelper.identifyUniqueElement`'s own separate `MultipleElementsFoundException` throw site (`ElementActionsHelper.java:807`) -- meaningfully expanding scope right after this PR was narrowed specifically to reduce blast radius.
- Proper verification would need mobile/touch driver test infrastructure; I haven't confirmed there's an equivalent headless-safe mocked-driver pattern for it the way `ActionsCoverageUnitTest` covers `Actions.java`, and constraints restrict spinning up mobile emulators without prior approval.
- The original issue #4321's reproduction (`mcp__shaft-mcp__element_type`) goes through `Actions.performAction` (i.e. `element().type()`/`element().click()`), which this PR fixes -- the touch path is a real, adjacent gap but not the reported one.

Will file as its own issue, cross-linked to #4321 and #4327, once you confirm.

## Non-blocking perf note (was flagged, now much smaller)

`uniqueDisplayedAndEnabledElement` now runs at most twice per `performAction` call (once if the primary loop's very last poll happens to hit it, which it won't since narrowing is gated off during the loop -- so really just once, on the single last-resort attempt) instead of once per 100ms poll for the whole timeout window. The 2N-round-trips-per-poll concern from the original design is effectively resolved by the last-resort reframing itself.

MohabMohie:
Filed the touch/tap follow-up as its own issue per confirmation: #4332. On the cosmetic indentation drift flagged above: no IDE reformat-selection tool available in this environment to do a confidently-scoped mechanical reindent, so leaving it as the disclosed known-cosmetic-issue described in my earlier comment rather than risking a manual bulk edit on this file.

MohabMohie:
All 3 required patches done in commit 68bbfaed13, PR body rewritten to match the actual implementation, and the 3 follow-up issues filed (cross-linked to #4321):

1. **Trace logging**: `ReportManager.logDiscrete(...)` added at both last-resort narrowing points (`Actions.java:530-536` initial lookup, `Actions.java:663-668` post-lazy-load refresh), naming the original match count and that auto-resolution happened after the identification timeout was exhausted.

2. **PR body rewritten**: now describes the actual last-resort implementation (line references verified against the live file), the two new elapsed-time/still-ambiguous tests, the review history (both rounds), the known indentation disclosure, and links to all 5 follow-up issues (#4327, #4332, #4334, #4335, #4336).

3. **Strengthened assertion**: `multipleButtonsDisabledEnabledFixture.html` now has `onclick` markers on both buttons (`document.title = 'save-clicked'` / `'cancel-clicked'`). `clickResolvesToTheOnlyEnabledButtonOnlyAfterTheIdentificationTimeoutIsExhausted` now asserts the title is `cancel-clicked` specifically. Verified the assertion actually discriminates: temporarily flipped the expected value to `save-clicked` and confirmed it fails with `expected [save-clicked] but found [cancel-clicked]`, then reverted. A native `disabled` button ignores click events entirely, so a silent no-op on `#save` would leave the title unset and correctly fail this assertion.

Follow-ups filed:
- #4334 -- last-resort path drops the original `TimeoutException`'s diagnostic context
- #4335 -- `webConfig.md` docs need a note on the new last-resort semantics
- #4336 -- `MultipleElementsFailureTest` isn't wired into `pr-gate.yml` (**correction to the original finding**: verified `ActionsCoverageUnitTest`/`ElementsHelperCoverageUnitTest` already ARE wired into `pr-gate.yml`'s shaft-engine unit-test step at line ~330 -- only the real-browser `MultipleElementsFailureTest` is actually missing)
- #4332 -- `touch().tap()` follow-up, filed per your confirmation

Re-verified after all changes: `MultipleElementsFailureTest` 7/7, `ActionsCoverageUnitTest` 43/43, `ElementsHelperCoverageUnitTest` 21/21, all isolated per-class runs, all headless, `mvn -pl shaft-engine test-compile` clean.

Ready for your final check before arming.

Issue #4332, opened 2026-07-28
touch().tap() / ElementActionsHelper.identifyUniqueElement: same self-heal disambiguation, separate mechanism


MohabMohie:
## Context

Follow-up from #4321 (self-heal `MultipleElementsFoundException` when exactly one match is actionable, fixed for the `Actions.performAction` path in PR #4326).

After PR #4326, `element().click()` / `element().type()` (etc.) correctly self-heal an ambiguous locator as a last resort once the identification timeout is exhausted -- but `touch().tap()` and the other `TouchActions` methods still hard-fail immediately on the identical DOM/locator, because they resolve elements through a completely separate code path.

## Why this is a separate mechanism, not an extension of #4326

- `Actions.performAction` retries through `SynchronizationManager`'s `FluentWait` (`shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java:495-933`), which is what PR #4326's last-resort narrowing hooks into.
- `TouchActions` resolves elements via `ElementActionsHelper.identifyUniqueElement(driver, elementLocator)` (`shaft-engine/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java:773-818`), which delegates to `getMatchingElementsInformation` -> `waitForElementPresence` -- an entirely different retry/timeout implementation with its own `MultipleElementsFoundException` throw site (`ElementActionsHelper.java:807`).
- There is no shared code between the two paths to extend; replicating #4326's fix here means designing an analogous "one last-resort narrowing attempt after this path's own timeout is exhausted" from scratch against `waitForElementPresence`'s actual retry mechanics, not reusing anything from PR #4326.

## Scope

At least 14 call sites in `shaft-engine/src/main/java/com/shaft/gui/element/TouchActions.java` go through `identifyUniqueElement` (tap, scroll-to-element, drag-and-drop, wheel-scroll origin, etc.), plus a few spots in `ElementActionsHelper` itself (submit-via-JS, accessible-name lookup).

## Suggested shape (not yet designed -- needs its own review)

- Read `waitForElementPresence`'s actual polling/timeout implementation first to confirm it has an equivalent "keep trying until identification timeout" loop to hook a last-resort attempt into.
[...]

PR #4343, opened 2026-07-28
Fixes #4327

### LINKED ISSUE #4327: Extend HealingManager.resolve() to the multi-match branch for SHAFT-Heal-enabled ambiguity self-healing
## Context

Fast-follow from #4321 (shaft-mcp / core: self-heal `MultipleElementsFoundException` when a real refinement narrows a locator to one element). #4321's fix (PR #4326) only handles the case where exactly one of the matched elements is displayed+enabled (a visibility-based disambiguation, reusing `Actions.chooseBestEffortDisplayedElement`'s existing filter). It intentionally does NOT touch SHAFT Heal.

This issue tracks the deliberately-deferred second half of the original investigation: **Option A** from that investigation --

> Extend `HealingManager.resolve()`'s call site to ALSO fire on the multi-match branch in `Actions.java` (in addition to the existing zero-match branch), for SHAFT-Heal-enabled users hitting DOM-drift-into-ambiguity with prior history.

## Why this is a distinct, separately valuable case

Investigated during #4321 (evidence below is a direct copy from that investigation so whoever picks this up doesn't have to re-derive it):

- `HealingManager.resolve()` (`shaft-engine/src/main/java/com/shaft/gui/internal/healing/HealingManager.java:44`) is currently only ever called from the **zero-match** branch (`Actions.java` around the `ElementLookup` helper, guarded by `!foundElements.isEmpty()` returning early before healing is ever attempted for 1-or-more found elements).
- `ShaftHealingProvider.resolve()` (`shaft-heal/src/main/java/com/shaft/heal/internal/ShaftHealingProvider.java:75-97`) requires a previously-saved `HistoryRecord` to relocate from. History is only saved via `HealingManager.observe()`, which itself only fires when `elements.size() == 1` (`HealingManager.java:113-116`).
- Net effect: this machinery **cannot** help a locator that has never uniquely resolved before (the #4321 reported scenario -- fresh session, first-ever call, immediately ambiguous). It only helps a *different, real* case: a locator that used to resolve uniquely and later started matching >1 elements (DOM drift/regression), where SHAFT Heal is enabled and history already exists from an earlier clean run.

## Suggested shape (not yet designed in detail -- needs its own review)

- In `Actions.java`'s multi-match branches (both the initial lookup and the post-lazy-load refreshed lookup -- same two call sites #4321/PR #4326 touched), when #4321's visibility-based narrowing does NOT resolve the ambiguity, attempt `HealingManager.resolve(...)` the same way the zero-match branch already does, before throwing `MultipleElementsFoundException`.
- Needs its own regression coverage proving: (a) SHAFT-Heal-disabled behavior is unchanged (default, zero new surface for the common case), (b) with SHAFT Heal enabled and prior history, an ambiguous re-match resolves to the historically-observed element, (c) with SHAFT Heal enabled but NO prior history, it still throws cleanly (no silent misbehavior).
- Architectural call on whether `HealingManager.observe()` should also be taught to record history from #4321's visibility-narrowed single element (currently it will, incidentally, since the narrowed list is fed straight into the existing `observe()` call) -- confirm this is desired, not just incidental.

## Cross-links
- Parent issue: #4321
- Fix for the visibility-based half: PR #4326

MohabMohie:
## Summary

Closes #4327. Extends #4321/PR #4326's zero-match self-heal mechanism to the **multi-match** branch in `Actions.java`: when a locator matches 2+ elements and the existing last-resort visibility narrowing (`uniqueDisplayedAndEnabledElement`) still can't disambiguate (0 or 2+ actionable candidates), consult SHAFT Heal's fingerprint history for that exact locator before throwing `MultipleElementsFoundException`, mirroring the existing zero-match branch's resolve()-then-throw flow (`shaft-engine/src/main/java/com/shaft/gui/element/internal/Actions.java:1300-1316`, `findAllElements`).

- Applies to both the initial lookup (`Actions.java:514-527`) and the lazy-load-refreshed lookup (`Actions.java:637-647`), which now share one `narrowOrHealOrThrowWhenAmbiguous` helper (`Actions.java:1855-1922`) instead of duplicating the narrowing logic across both call sites.
- `ShaftHealingProvider.resolve()` (`shaft-heal/src/main/java/com/shaft/heal/internal/ShaftHealingProvider.java:62-131`) already keys purely off locator/pageKey/context fingerprint history via its own DOM re-search (`CandidateExtractor.extract`), independent of the original locator's live match count -- so no shaft-heal changes were needed to support this (confirmed by reading the provider end-to-end before implementing; no genuine architectural fork).
- Per the #4326 review's cross-link comment (`Actions.java:529-536`): a last-resort-narrowed single element seeding heal history is treated as an equally valid "this is the element the user meant" signal as any naturally-unique match -- no special-casing needed, since the provider only scores against the live fingerprint, never its provenance.

## Acceptance criteria (from #4327)

- (a) SHAFT-Heal-disabled behavior unchanged: `multiMatchLocatorShouldNeverInvokeHealingProviderWhenShaftHealDisabled`
- (b) SHAFT Heal enabled + prior history -> resolves to the historically-observed element: `multiMatchLocatorShouldHealToHistoricallyObservedElementWhenLastResortNarrowingCannotDisambiguate`
- (c) SHAFT Heal enabled + no prior history -> still throws cleanly: `multiMatchLocatorShouldStillThrowWhenHealingProviderHasNoHistoryForIt`

## Test plan

- [x] New tests in `HealingActionsIntegrationTest` watched RED (the "should heal" test failed against unmodified code, still throwing `MultipleElementsFoundException`) then GREEN after the fix.
- [x] `mvn -pl shaft-engine test -Dtest=HealingActionsIntegrationTest,MultipleElementsFailureTest,ActionsCoverageUnitTest,HealingStrategyTest` -- 73/73 passed, no regressions in the zero-match healing flow or #4321's last-resort flow.
- [x] `mvn -pl shaft-engine test-compile` and `mvn -pl shaft-heal test-compile` both clean.
- [x] Rebased onto `origin/main` (now includes #4332/PR #4340) and reran the full scoped suite with `TouchMultipleElementsFailureTest` included: `mvn -pl shaft-engine test -Dtest=HealingActionsIntegrationTest,MultipleElementsFailureTest,ActionsCoverageUnitTest,HealingStrategyTest,TouchMultipleElementsFailureTest -DheadlessExecution=true` -- 76/76 passed, no regressions.

https://claude.ai/code/session_01NGdgmezM75NX62aMypA5jn


MohabMohie:
## Independent architectural review: BLOCK

Requested a second independent adversarial pass before arming auto-merge, given this extends the same silent-wrong-resolution risk surface #4326 was blocked for once already. Verdict: **BLOCK**, one decision_needed finding, everything else verified clean (76/76 tests including `TouchMultipleElementsFailureTest` 3/3, RED independently reproduced by reverting `Actions.java` alone).

### Blocking: healed element is never checked against the live match set

`Actions.java:1919` does `foundElements.set(healed.get().elements())` with no check that the healed element is actually among the 2+ elements the locator just matched live. `CandidateExtractor` re-searches the whole DOM from the saved fingerprint independently of the original locator (`By.id`/`By.xpath("//*")`), so it can return an element that isn't one of the current ambiguous matches at all.

Reproducible: a locator uniquely resolves once, seeding a heal fingerprint for that element. Later, the page changes such that the *original* healed element is still present in the DOM (now under a different, unrelated class) while the *current* locator now matches two different new elements. On the next run, both new matches are displayed+enabled (last-resort narrowing can't disambiguate), heal resolves against the fingerprint, finds the original (now-unrelated) element still in the DOM, scores it past the healing threshold — and SHAFT clicks/types on an element the locator matches **zero** times, instead of throwing on the genuine ambiguity.

Zero-match healing has no live evidence to discard, so this risk doesn't exist there. Multi-match healing does have live evidence (the 2+ actual matches) and currently ignores it entirely.

**Required fix**: before accepting the healed result, require `foundElements.get().contains(healed.get().elements().getFirst())` (or equivalent), else fall through to the existing throw. This is exactly the PR's own stated intent — "resolves to the historically-observed element **among the current matches**" — just not currently enforced in code.

### Patch (fix in this PR)
1. `Actions.java:1905` never sets `flakeProfile.healingAttempted`, unlike the zero-match path (`Actions.java:506-508`) — flake telemetry under-reports this healing tier.
2. `HealingActionsIntegrationTest.java:81-110` never asserts `times(1)` (or similar) on `resolve` gated by `lastResortNarrowingAllowed` — removing that gate (`Actions.java:1897`), which is exactly what #4326 round 2 required, would still pass green today. Strengthen the test so it actually locks in the gating.
3. `HealingManager.java:32-33` javadoc still says healing is "only for locator-not-found failures" — now false, update it.

### Defer (file as follow-up issue, don't fix here)
`ElementActionsHelper.java:333-349` (touch/tap, #4332) has #4321's last-resort narrowing but no heal tier at all — `click`/`type` and `tap` now behave differently on an identical ambiguous DOM. Track separately.

### Confirmed NOT an issue (dismiss)
- Not silent: `ReportManager.log(WARN)` with original+healed locators plus a `logDiscrete` call — better visibility than #4321's own last-resort had before its own logging fix.
- No regression on the non-healed path: helper reproduces both prior branches exactly, `lastResortNarrowingAllowed=false` still throws immediately.
- No thread-safety/API surface concern.

Evidence: `Actions.java:1874-1924,524-526,648-650,1283-1291,1905,1919`; `ShaftHealingProvider.java:66-75,148,246`; `HealingDecisionEngine.java:43-71`; `HealingManager.java:32-33,236-247`; `HealingActionsIntegrationTest.java:81-110`.

PR #4443, opened 2026-08-04
Fix #4440: give the CI Maven cache a restore-key fallback

### LINKED ISSUE #4440: Nightly failure: E2E Tests
The scheduled workflow **E2E Tests** failed.

- Run: https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30876948089

Filed automatically. Dedupes on the `nightly-failure:e2e-tests` label (one open issue per workflow) and closes automatically when `E2E Tests` next succeeds.

MohabMohie:
Fixes #4440.

## What broke

The 2026-08-04 **E2E Tests** nightly ([run 30876948089](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30876948089)) died in `Ubuntu_Firefox_Grid` with:

```
Could not transfer artifact io.netty:netty-bom:pom:4.2.16.Final from/to central
(https://repo.maven.apache.org/maven2): status code: 429, reason phrase: Too Many Requests (429)
```

...repeated for every BOM in the reactor, after six minutes of resolver retries.

## Root cause

`actions/setup-java`'s `cache: 'maven'` passes **no** `restoreKeys` — deliberately, per its own source comment ("No 'restoreKeys' is set, to start with a clear cache after dependency update", `src/cache.ts`). Its key is an exact hash of every `pom.xml`.

Three pom-touching commits landed between the two nightlies (#4415, #4416, #4419), so the cache key changed and the job logged:

```
2026-08-04T04:13:03Z  maven cache is not found
```

against `Cache restored from key: setup-java-Linux-x64-maven-9d7f1fb8...` on the passing 2026-08-03 run. `e2eTests.yml` starts **22 jobs** through `setup-test-env`, so all 22 cold-resolved the full tree from Maven Central within the same few seconds. Central rate-limited.

`Install optional visual test runtime` carries no `continue-on-error` (unlike the `Run tests` step right below it), so the job stopped there → `Run tests` skipped → `Post-Test Report and Check` failed on `No Allure summary was generated`.

There was also no fallback to recover to: the 2026-08-03 Maven cache had already been LRU-evicted. Repo cache usage is **9.08 of 10 GiB**, dominated by multi-GiB Gradle entries, and today's exact-key scheme only touches the Maven entry on the runs where the hash happens to match — so it ages out fast.

## Change

New `.github/actions/cache-maven-repo` composite: same `~/.m2/repository` path, keyed `shaft-m2-<os>-<arch>-<pom-hash>` with a `shaft-m2-<os>-<arch>-` **restore-key prefix**. A pom bump now restores the previous ~500 MiB repository and downloads only the delta. Every run touches the entry, so it also stops aging out.

It drops `*.lastUpdated` markers after restore. Maven records *failed* release resolutions there and, with the default `never` update policy for releases, will not reattempt them — so a cache saved by a job that hit the 429 would otherwise poison every later job. That failure mode only becomes reachable once restore-keys exist.

Call sites:

- `setup-test-env` — covers all 22 `e2eTests.yml` jobs plus `e2eLocalTests.yml` and `lambdatestTests.yml`.
- `pr-gate.yml` `cli` (2-way matrix) and `unit-tests` (7-way matrix) — same herd shape on every PR. Leaving these on setup-java's key would fork the namespace into two ~500 MiB families per pom hash, which the eviction budget cannot afford.

The four rare release/publish workflows still on `cache: 'maven'` are single-job and out of scope here — filed separately.

`pr-gate`'s `infra` path filter now covers `.github/actions/**`, so editing a composite action re-runs the jobs that call it. It previously watched only `pr-gate.yml` itself.

## Evidence

- Validators: `scripts/ci/validate_workflow_timeouts.py` and `scripts/ci/validate_agent_setup.py --skip-external` both pass.
- Cache behaviour is verified on this PR's own runs — see the comment thread.


MohabMohie:
## Cache behaviour verified

Two `pr-gate` runs on this branch. The second carries a throwaway probe commit whose only purpose is to change `hashFiles('**/pom.xml')` — the exact condition that broke the 2026-08-04 nightly. Both probe commits are reverted before merge.

**Run 1** ([30885503538](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30885503538)) — new key prefix, nothing to restore yet, so a miss is the honest expected result. It saved:

```
shaft-m2-Linux-X64-c10dabe6e5794593159d576bf5e0c3e62c6fe2b76d2e142fc41f154b38022644   401 MiB
```

**Run 2** ([30885948056](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30885948056), `Unit Tests (shaft-ai)`, job 91917162219) — pom hash changed, so the primary key misses and the prefix has to carry it:

```
key:          shaft-m2-Linux-X64-e34f2c186f4ed62a08dbe82c95f2550536d8d9f1e21e9f7eff8b9dcd6c03a8de
restore-keys: shaft-m2-Linux-X64-
Cache hit for restore-key: shaft-m2-Linux-X64-c10dabe6e5794593159d576bf5e0c3e62c6fe2b76d2e142fc41f154b38022644
Cache restored successfully
Cache restored from key:   shaft-m2-Linux-X64-c10dabe6e5794593159d576bf5e0c3e62c6fe2b76d2e142fc41f154b38022644
```

That is the same situation that logged `maven cache is not found` at `2026-08-04T04:13:03Z` in [run 30876948089](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30876948089). The pom hash changed and the ~400 MiB repository was restored from the previous hash's entry, so only the delta is fetched from Central rather than the whole reactor by 22 jobs at once.

Also confirmed in the same job:

- the `*.lastUpdated` prune step ran against the restored repository;
- `Failed to save: Unable to reserve cache with key ..., another job may be creating this cache` — the expected benign outcome when concurrent matrix jobs race to save the same key. One wins, the rest skip; no job fails.

Windows (`shaft-cli (windows-2025)`) exercises the same action and is still running; `~/.m2/repository` and the `$HOME`-based prune resolve differently there, so it is the one platform detail left to confirm before merge.

MohabMohie:
### Windows confirmed, probe reverted

`shaft-cli (windows-2025)` (job 91917162012) finished `success`:

```
key:          shaft-m2-Windows-X64-c54f2c56ff9a0946f072e940f6db17890e9d725fab5604a9bd2483688ad24794
restore-keys: shaft-m2-Windows-X64-
Cache not found for input keys: ...
find "$HOME/.m2/repository" -name '*.lastUpdated' -type f -delete
Cache saved with key: shaft-m2-Windows-X64-c54f2c56ff9a0946f072e940f6db17890e9d725fab5604a9bd2483688ad24794
```

A miss was expected there — run 1's Windows job was cancelled by the probe push before its post-job save, so nothing existed under the prefix yet. What this shows is the part that could have gone wrong on Windows: `~/.m2/repository` resolves for `actions/cache`, and the `$HOME`-based prune runs under Git Bash without failing the job. Both platforms are now covered.

Every check on the probe run ([30885948056](https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/30885948056)) passed, including `PR Gate Summary`.

Both probe commits are reverted; `shaft-visual/pom.xml` is byte-identical to its state before them. The diff against `main` is now only:

```
.github/actions/cache-maven-repo/action.yml        | 41 ++++++++++++++++++++++
.github/actions/setup-test-env/action.yml          |  6 +++-
.github/workflows/pr-gate.yml                      | 10 ++++--
.memory/events.jsonl                               |  1 +
.memory/memory/gotchas/...cold-downloads-the-world.json | 33 +++++++++++++++++
.memory/memory/gotchas/...cold-downloads-the-world.md   |  9 +++++
```

Auto-merge is armed (squash).

Issue #4452, opened 2026-08-04
Tracking: make the agent harness enforce what it already knows


MohabMohie:
## What was asked, and what I am actually proposing

The ask was to distil "my mind, brain, soul, consciousness, personality" into the harness so future
agents — on any provider — build on it instead of rediscovering it.

Three of those five words do not map onto anything implementable, so I am not going to pretend they
do. There is no artifact that carries consciousness, and a file that claims to is just prose that
costs bytes. Two of them do map, and both are **already done**:

- *Personality* is `Caveman` + `Ponytail` + the six iron laws + the red-flag phrase list in
  `.agents/skills/act-as-mohab/SKILL.md`. That is a written voice, a written simplicity ladder, and a
  written list of the sentences an agent says to itself right before it breaks a rule. There is
  nothing left of "personality" to transfer that is not already in that file.
- *Way of working* is the triage table, the solo-or-orchestrate rule, the delivery lifecycle in
  `.agents/skills/consult-first/SKILL.md`, and the Learning-loop routing table.

So the honest version of this task is not "write down more of what I know." After reading the whole
harness I could not find a rule worth adding that was not already there. The gap is somewhere else,
and it is the thing I actually want to leave behind:

> **This harness knows more than it can prove. Every rule it can prove has survived. Every rule it
> could only state has been eroded, softened, or deleted — usually by an agent optimising for
> bytes.**

The deliverable is therefore mechanism, not text: convert the rules that matter most from *stated*
to *checked*, freeze the always-loaded surface, and make every future addition pay for itself. Net
guidance bytes go **down**, not up.

I am also deliberately not quoting the owner's framing verbatim, because it names a model. Tracked
guidance never does that (`scripts/ci/agent_guidance_budget.json` → `forbidden_patterns`, and
`tests/scripts/test_agent_harness_portability.py::test_delegation_policy_uses_capability_tiers_not_fixed_models_or_effort`),
and a ticket proposing harness changes should model the rule it is asking future agents to keep.

## Evidence base

Everything below was read in this session, in a clean worktree:

- `AGENTS.md`, `CLAUDE.md`, `.claude/skills/*/SKILL.md`, `.agents/skills/README.md`
- `.agents/skills/act-as-mohab/SKILL.md` and all 14 files under `references/`, all 13 playbooks,
  all 10 mastery chapters, `.agents/skills/consult-first/SKILL.md`
[...]

Issue #3745, opened 2026-07-18
Guidance budgets: replace the global byte-reduction floor with per-surface caps matched to actual load cost


MohabMohie:
Owner directive: revisit all instruction-size limitations and reset them as the orchestrator sees fit.

Analysis: the per-host estimated-token caps (2900/host over always-loaded files + skill metadata) model the real per-session cost and stay. The .agents bridge thinness caps stay. But the global `reduction_baseline_bytes`/`minimum_reduction_percent` floor (150000 × 47% = 70500 bytes over ALL guidance, including lazy-loaded skill bodies) is a historical artifact: it has been manually renegotiated three times in one week (68→55→53%), taxes every skill improvement against an unrelated global pool, and today forced deleting good prose over 21 bytes. The config's own notes concede skill bodies are lazy-loaded and shouldn't be judged by it.

New regime:
1. DROP `reduction_baseline_bytes` + `minimum_reduction_percent` (validator already no-ops when absent).
2. ADD per-skill-body caps where load cost actually lands: `.claude/skills/*/SKILL.md` ≤ 16384 LF-normalized bytes each (largest today: act-as-fable 15979; a skill that needs more should move detail to uncounted on-demand `references/`), `.github/skills/*/SKILL.md` ≤ 4096 (largest today 2257).
3. KEEP host-token caps, per-file caps on always-loaded/host-consumed files, `.agents` bridge caps, expected_skill_names roster (adding a skill stays a deliberate config change), duplicate-paragraph/forbidden-pattern/stale-reference intent guards.
4. Update the config `notes` to state the principle: budgets bound each surface by its own load cost; growth within a skill is judged by the duplication/intent guards, not a shared pool.
5. Extend validate_agent_guidance.py + tests for the new per-skill byte cap.

Issue #4459, opened 2026-08-04
Stop reporting an unmeasured guidance reduction as 0%


MohabMohie:
Subtask 6 of #4452.

`py -3 scripts/ci/validate_agent_setup.py --skip-external` prints, on a clean tree:

```
Agent setup is valid: 129090 guidance bytes, 0% reduction, 336 memory objects.
```

`0% reduction` is not a measurement. `collect_metrics` does:

```python
baseline = budget.get("reduction_baseline_bytes", 0)
reduction = 0 if not baseline else round((1 - guidance_bytes / baseline) * 100, 2)
```

No baseline is configured (deliberately — #3745), so the branch that computes a percentage
never runs and the literal `0` is emitted as though it were a result. An agent reading the
banner cannot tell "measured no reduction" from "nothing was measured", and 0% reads as the
worse of the two. #4452 spent a paragraph on this precisely because the banner misled it.

This is independent of #4458: whichever way that decision goes, a percentage must not be
reported when nothing computed one.

## Scope

- `collect_metrics` reports `guidance_reduction_percent: None` when no baseline is
  configured, and the real percentage when one is.
- The banner omits the reduction clause entirely when there is nothing to report.
- Focused tests for both, in `tests/scripts/test_validate_agent_setup.py`.

Zero guidance bytes: `scripts/` and `tests/` are not in `total_guidance_globs`.

## Verification

- `python -m unittest tests.scripts.test_validate_agent_setup` — RED before, GREEN after
- `py -3 scripts/ci/validate_agent_setup.py --skip-external`

PR #4475, opened 2026-08-04
Decide the guidance ratchet and the skills-map duplication

### LINKED ISSUE #4458: Decision needed: arming a total-guidance ratchet reverses the closed directive in #3745
Subtask 5 of #4452. **Blocked on an owner decision. Not implemented.**

#4452 proposes setting `reduction_baseline_bytes` and `minimum_reduction_percent` in
`scripts/ci/agent_guidance_budget.json` so `validate_total_reduction` starts enforcing a
ceiling on total guidance bytes.

The observation behind it is correct and reproduced on a clean tree:

- Neither key exists in the budget file.
- `validate_total_reduction` returns `[]` unconditionally when either is `None`.
- `collect_metrics` does `baseline = budget.get("reduction_baseline_bytes", 0)` then
  `reduction = 0 if not baseline else ...`, so the banner's `0% reduction` is the absence of
  a measurement, not a measurement.

What #4452 gets wrong is the cause. It reads the absence as a leak — "the leak through which
every future 'just one more reference file' arrives". It is not a leak. Those keys were
removed deliberately by #3745 (landed in PR #3748), an owner directive whose analysis is
explicit:

> the global `reduction_baseline_bytes`/`minimum_reduction_percent` floor (150000 x 47% =
> 70500 bytes over ALL guidance, including lazy-loaded skill bodies) is a historical
> artifact: it has been manually renegotiated three times in one week (68->55->53%), taxes
> every skill improvement against an unrelated global pool, and today forced deleting good
> prose over 21 bytes.

#3745's replacement principle: "budgets bound each surface by its own load cost; growth
within a skill is judged by the duplication/intent guards, not a shared pool." The no-op is
pinned by `test_total_reduction_is_noop_when_unconfigured`, whose comment cites #3745 by
number.

So arming the global ratchet is not filling a gap, it is reversing a decision — and it would
reinstate the exact failure the owner named: a shared pool that taxes every unrelated
improvement and gets renegotiated whenever it binds.

## The decision

1. **Leave it retired** (my recommendation). Per-surface caps already bound every surface
   that has a real load cost. What is genuinely unbounded is the *on-demand* reference and
   playbook surface, which by #3745's own principle should be bounded by its own load cost
   — a per-file or per-directory cap on `references/**` — not by a global pool that also
   counts `AGENTS.md` and every host adapter.
2. **Re-arm the global ratchet** anyway, accepting the renegotiation treadmill #3745
   documented. If chosen, `test_total_reduction_is_noop_when_unconfigured`'s comment needs
   rewriting so it stops citing #3745 as current policy.

Option 1's concrete form is a per-surface cap and is a design choice, not a mechanical edit,
so it is not being made unilaterally here.

The reporting half of the problem — the banner presenting a non-measurement as `0% reduction`
— is true regardless of which option wins and is handled separately.

### LINKED ISSUE #4463: Decide whether the skills map's Method references table is duplication too
Raised by the independent adversarial review of PR #4462, which observed that the PR applies
its own principle inconsistently inside one file. Filing rather than extending that PR's
scope.

PR #4462 deleted the `Covers` and `Read it when the task touches` columns from
`.agents/skills/README.md` because they were `routing.md`'s trigger column with the columns
swapped. The **Method references** table (`README.md:119-128`) survived untouched. The
question is whether that is a principled line or an arbitrary one.

## It is genuinely two different cases, but only for half the table

Four of the eight rows describe files the routing table does **not** have a trigger row for
— `routing`, `delegation`, `roles`, `heuristics` are reached from the entrypoint's prose,
not from the table. Their README rows summarise what each file *contains*. There is no
authoritative trigger column being duplicated, and a map does need some way to say what a
file is.

The other four rows have the same defect that was deleted:

| README | routing.md |
| --- | --- |
| `verification-gap lens` — "A worked method for finding changed behaviour that would break without any check noticing." | "Reviewing a diff for behavior that no check would catch" |
| `orchestrator bootstrap` — "The opening sequence when a session holds the main thread: gather live state, queue by priority, ticket, dispatch." | "Holding main thread at session start" |
| `work GitHub playbook` — "Taking one issue from filed to merged: scope grounding, tracking issues, grouped PRs, review, and close-out." | "One issue through a merged PR" |
| `graphify` — "How to ask the structural index what calls what, and what to do when it is unavailable." | "What calls or depends on this?" (knowledge table) |

So the line drawn by #4462 is not "method references are different" — it is "these thirteen
playbooks and ten chapters were the ones I looked at."

## The residual risk on the other four rows

A contents summary drifts too, and drifts silently. The `delegation` row asserts "the
four-agent concurrency cap"; if `delegation.md` ever changes that number, the README keeps
claiming four and nothing fails. That is the same failure mode as the trigger duplication,
one step removed — and worse in one respect, because a wrong number reads as authoritative
where a wrong trigger only misroutes.

## Options

1. Delete the four rows that duplicate a routing trigger; keep the four that summarise a
   file with no trigger row. Consistent with #4462's stated principle, and it leaves the map
   able to say what `routing`/`delegation`/`roles`/`heuristics` are.
2. Delete the whole column, matching the playbook and mastery treatment exactly. Cheapest to
   state, and the map becomes a pure inventory.
3. Keep all eight and accept the drift, on the grounds that a contributor-facing map with no
   prose at all is not worth having. If this wins, it is worth saying so in the file so the
   next agent does not delete it as an oversight.

I lean to option 1, but this is a judgement about contributor-facing docs rather than about
agent behaviour, and #4462 already carries one such trade for review. Not doing it
unilaterally.

Related: whichever option wins, the underlying gap is that `NoDuplicationTest` cannot see a
paraphrase at all — its own docstring concedes this. Nothing here makes that check stronger.

### LINKED ISSUE #4466: Test pins the absence of reduction_baseline_bytes, which #4458 may legitimately add
From the independent adversarial review of PR #4461, on a line that PR introduces.

`tests/scripts/test_validate_agent_setup.py:292` asserts:

```python
self.assertNotIn("reduction_baseline_bytes", budget)
```

That turns a legitimate, reversible configuration choice into a test failure — and **#4458 is
open proposing exactly that choice.** If the owner decides to re-arm the total-guidance
ratchet, this line goes red for a reason that has nothing to do with what the test is named
for.

## Why it is there, and why that is not a good enough reason

It guards the test's own premise. `test_unconfigured_reduction_is_reported_as_absent_not_as_zero_percent`
asserts `metrics["guidance_reduction_percent"] is None`, which is only meaningful while no
baseline is configured; without the guard the test would silently become vacuous if one were
added. So it is not gratuitous — but a premise guard should skip or reconfigure, not fail.

The sibling at `:296-301` (`test_configured_baseline_still_reports_a_real_reduction`) already
covers the configured path directly against `reduction_percent`, so the *behaviour* is
covered either way.

## Suggested scope

Drive the unconfigured case from a synthetic budget rather than the repository's real one, so
neither test depends on which choice #4458 lands on. Failing that, make the premise a
`skipIf` with a message naming #4458.

Small and self-contained. Filed rather than folded into #4461 so that PR's diff stays about
what it was reviewed for.

MohabMohie:
## 📋 Description

Resolves the two open decisions on the guidance harness and the test line that
would have forbidden one of them. The total-guidance ratchet stays retired
(#3745 stands); what was genuinely unmetered — the on-demand reference surface
— is now bounded by its own per-task load cost instead of by a shared pool.

## 🔗 Related Issue(s)

- Closes #4458
- Closes #4463
- Closes #4466
- Related to #4452

## 🗂️ Type of Change

- [x] 🔧 CI/CD or infrastructure change
- [x] 📖 Documentation update
- [x] ✅ Test coverage improvement

### #4458 — do not re-arm the global ratchet; cap the on-demand surface per file

#4452 read the absent `reduction_baseline_bytes` / `minimum_reduction_percent`
as "the leak through which every future 'just one more reference file'
arrives". They are a deliberate repair, made by the closed owner directive
#3745. Its objection is structural, not a matter of the number being wrong:
under a shared pool, the only way to fund one file's growth is to delete an
unrelated file's prose. That is why it was renegotiated three times in one week
(68→55→53%) and why it "today forced deleting good prose over 21 bytes".

The real question was whether anything catches unbounded growth without
reinstating that. Four candidates were considered:

| Candidate | Verdict |
| --- | --- |
| Leave as-is, say so on the record | Half right — the diagnosis holds, but a surface really is unmetered |
| Non-blocking advisory of total drift | **Already shipped.** `guidance_bytes` is printed on every run and gates nothing. Needs no code |
| Per-surface cap on the on-demand surface | **Chosen** |
| Ratchet that fails only past a generous band | Rejected — a loose pool is still a pool; it just defers the first unfair deletion |

`file_budgets` keys may now be globs, each matched file bound on its own, and
`.agents/skills/act-as-mohab/references/**/*.md` is capped at **16384 bytes** —
the ceiling `max_always_loaded_body_chars` already uses, leaving the largest
file (14838, `work-github-playbook.md`) ~10% of headroom.

> **Corrected after review.** The first version of this PR justified the cap by
> claiming routing sends a task to *one* reference, so the per-task cost is the
> largest single file. That is false, and `routing.md:12-16` refutes it — every
> dispatch goes through `delegation.md` and `roles.md` before a row is read. A
> realistic dispatch task loads ~28639 bytes of references, 1.75x the cap. See
> the review comment below; the claim is gone from the config note and the
> docstring.

**So this bounds the worst single file, not a session total, and that is
deliberate.** The argument is about who pays when a ceiling binds. Under a pool
the way to fund one file's growth is deleting a neighbour's prose — which lowers
nothing any session loads, because the neighbour was not in it. Under a per-file
cap the file that grew is the file that pays, and the fix is to split it —
though a split half is usually reached from the other, so that relocates load
rather than removing it. **Incentive locality is the whole claim.** #3745's own
per-skill caps have never been renegotiated; the pool was renegotiated three
times in one week.

> **Corrected after review, second time.** The clause "which does lower what a
> task loads" was also false, and `work-github-playbook.md:107` refutes it on
> the very file the cap binds: "…are covered in [delegation](delegation.md) and
> are not restated here" — that split already happened, in #4462, and its
> outcome was a pointer to another reference that gets loaded anyway. Worse, a
> split grants each half its own 16384. The clause is gone from the config note
> and the docstring. See the note in the report below on why this keeps
> happening.

**The argument against my own choice.** A per-file cap is also a ceiling someone
renegotiates the day it binds — the treadmill #3745 named — and at 16384 it
meters exactly one file in 31 while permitting 507904 bytes in total, 5.5x
today. That is a real weakness: the other 30 files can grow many-fold in
silence. It is accepted because the alternative that catches it is the pool, and
the pool's failure is worse and unfixable by tuning.

**What would change the answer.** If a host ever preloaded a reference
directory, the aggregate would become a real ceiling rather than a tax, and a
bounded pool over that surface — never one spanning `AGENTS.md` and every host
adapter — would be justified. Growth by *file count* rather than file size is
the other vector a per-file cap misses; that one is gated by deliberateness
today, since adding a reference requires a routing-table row and a skills-map
entry, both checked.

`validate_total_reduction` is left in place and still no-ops when unconfigured:
the decision is meant to stay reversible.

### #4463 — the Method references table was duplication in **all eight** rows

#4463 argued four rows duplicate a routing trigger and four describe files with
no routing row. The second half does not hold. The entrypoint itself describes
all four:

| README row | Authority it restates |
| --- | --- |
| `routing` | `SKILL.md` § Route — "maps the deliverable in front of you to the one surface that owns it… also orders knowledge retrieval" |
| `delegation` | `SKILL.md` — "defines the levels and the review gate" |
| `roles` | `SKILL.md` — "defines role boundaries" |
| `heuristics` | `SKILL.md` — "only for deeper investigation, risk analysis, or review" |

So option 1 would have deleted the milder four and kept the worse four —
#4463 says so itself: a contents summary asserts numbers ("the four-agent
concurrency cap", "the five portable roles") that read as authoritative once
they drift, where a stale trigger only misroutes.

Collapsed to an inventory, matching the playbook and mastery treatment exactly.
The line is now stateable: **the map describes a surface only where nothing it
links to already describes it.** Written into the file so the next agent does
not restore it as an oversight.

Per the `CONTRIBUTING.md` lesson from #4462, claims about this file were checked
before changing it. `CONTRIBUTING.md:109` promises the map "explains how the
skills fit together and how each host imports them" — the four diagrams and the
Importing section still deliver both, so it stays true and is not edited. One
claim inside the file itself was already false before this PR and is fixed here:
the heading `## What each surface does` sat over three inventory lists after
#4462. It now reads `## The surfaces`.

`10424 → 9857` bytes. Total guidance `126723 → 126156`.

### #4466 — the test no longer forbids the choice

Both reduction cases now run against a fixture budget, so neither depends on
which way #4458 went. Nothing was deleted or weakened: the existing unit test at
`:295` is untouched, and the measured branch gains end-to-end coverage through
`collect_metrics` that it did not have before.

`scripts/ci/validate_agent_setup.py` is **not** touched — only its test file.

## 🧪 How to Test

1. `py -3 scripts/ci/validate_agent_setup.py --skip-external`
2. `py -3 -m unittest tests.scripts.test_agent_router_contract tests.scripts.test_agent_harness_portability tests.scripts.test_guard_nul_corruption tests.scripts.test_guard_lifecycle tests.scripts.test_validate_agent_guidance tests.scripts.test_validate_agent_setup tests.scripts.test_validate_skills tests.scripts.test_sync_user_harness tests.scripts.test_shaft_skills_content tests.scripts.test_worktree_hygiene tests.scripts.test_maven_retry`
3. `py -3 scripts/agents/guard.py --self-test`
4. Add `reduction_baseline_bytes` to the budget and re-run step 2 — it stays green (that is #4466).

## Evidence

**RED, observed before the code existed** — the glob budget key was silently ignored:

```
FAIL: test_a_glob_file_budget_caps_every_file_the_pattern_matches
AssertionError: Items in the second set but not the first:
'.agents/skills/example-bridge/references/big.md'
```

**The armed cap binds, proven by mutation** (16384 lowered to 14000 on a scratch copy):

```
size-budget: .agents/skills/act-as-mohab/references/work-github-playbook.md: 14838 bytes exceeds configured maximum 14000
exit=1
```

**#4466 reproduced on the pre-fix line**, after adding the key to the real budget:

```
FAIL: test_unconfigured_reduction_is_reported_as_absent_not_as_zero_percent
  File "tests/scripts/test_validate_agent_setup.py", line 292
    self.assertNotIn("reduction_baseline_bytes", budget)
AssertionError: 'reduction_baseline_bytes' unexpectedly found in {...}
```

**After the fix, the same mutation is green:**

```
Ran 21 tests in 4.134s
OK
```

**Full harness suite** (the eleven `agent-guidance` modules):

```
Ran 231 tests in 40.734s
OK
```

**Validator:**

```
Agent setup is valid: 126156 guidance bytes, 336 memory objects.
```

**`git diff --check`:** clean, no output.

## Documentation Site

- Documentation not required because: this changes agent-facing guidance
  configuration and a contributor-facing map inside this repository. No public
  SHAFT behaviour, API, or usage changes.

## 📝 Additional Notes

Auto-merge deliberately **not** armed — independent adversarial review first.

Reviewers should push hardest on two things. First, whether 16384 is the right
number rather than merely a defensible one: it binds `work-github-playbook.md`
at ~10% headroom, and if that file is expected to keep growing the honest answer
is to split it now rather than raise the cap later. Second, whether the
inventory README still earns its place for a human — the claim being made is
that the diagrams carry the understanding and the tables were only ever an
index.

Not attempted: `NoDuplicationTest` still cannot see a paraphrase, so nothing
here prevents the next contents summary from being added. That gap is real and
stays open rather than being papered over.

**Worth recording.** Two separate causal claims about load were made in this PR
and both were refuted by this repository's own structure — first "a task loads
one reference" (`routing.md:12-16`), then "splitting lowers what a task loads"
(`work-github-playbook.md:107`, where the split already happened and produced a
pointer to another reference). The decision survived both times and only the
reason failed. A reason that keeps failing while the conclusion keeps holding
means the conclusion never rested on it. What it actually rests on is
**incentive locality** — who pays when a ceiling binds — which is a claim about
authorship behaviour, not about bytes loaded. The instinct to re-derive it as a
load argument is the thing to distrust here, and the config note now says so in
the place an agent reads when the cap binds.


MohabMohie:
## Independent adversarial review: one rationale refuted, three decisions held

Reviewed by an agent that did not write the change, prompted to refute it, read-only. Its
findings are recorded here in full rather than summarised into a pass, because the interesting
one was a refutation.

### What it refuted

**The load-cost premise was false.** The cap was justified by "routing sends a task to ONE
reference, so what it pays is the largest single file". The reviewer checked that against this
repository's own routing table instead of accepting it:

- `references/routing.md:12-16`, "Always, before any row" — *"use delegation before dispatch
  and roles for role boundaries."* Two references, mandated before a row is read.
- `SKILL.md:150` puts `routing.md` itself (6171 bytes) on every task.
- `SKILL.md:154` — *"Load one surface, finish its deliverable, then return here for the next."*

Measured worst realistic dispatch task: routing 6171 + delegation 4040 + roles 1911 +
orchestrator-bootstrap 1679 + work-github-playbook 14838 = **28639 bytes, 1.75x the cap**.

Worse, the PR body's own stated reversal criterion — *"if routing began requiring several
references per task, the aggregate would become the real load cost"* — was **already satisfied
on the day it was written**. Shipping that would have left the next agent reasoning from a
premise `routing.md` contradicts: exactly the two-files-asserting-different-things failure this
PR is otherwise about.

Fixed in c03881ca56. The conclusion stands, because the argument that carries it never needed
the false claim: under a pool the way to fund one file's growth is deleting a neighbour's
prose, which lowers nothing any session loads; under a per-file cap the file that grew is the
file that pays and the fix is to split it. Both the config note and the docstring now say that
and explicitly state that a session total is *not* bounded.

Also corrected from the same review:

- **Stale appeal to precedent.** "The figure #3745 set for a substantive skill body" — no live
  skill-body budget is 16384 (`.agents/skills` is 20000, `.claude/skills` 1024, `.github/skills`
  4096). The only live 16384 is `max_always_loaded_body_chars`, which is what the note cites now.
- **A contradiction the new prose created.** `README.md:38`'s flowchart has routing reaching
  `heuristics`; `grep heuristics references/routing.md` returns zero. The diagram error was
  pre-existing, but with the table gone the diagram became the README's only statement about
  those files, and the new sentence made it a contradiction inside one file. Added the
  entrypoint edge the prose describes.
- **A false implication.** "The rest are reached from routing" implied exclusivity;
  `work-github-playbook` is reached from both. It says "described by" now, which is the claim
  actually being made.
- **A latent crash I introduced.** A literal `file_budgets` key containing `..` would have hit
  `ValueError` from `relative()` (which resolves before subtracting the root) where the
  pre-PR code produced a clean issue. Literal keys report the configured string verbatim
  again; verified:
  ```
  [{'code': 'size-budget', 'path': '../outside.md', 'message': '100 bytes exceeds configured maximum 10'}]
  ```
  No test added: the case requires a config shape that should never exist, and a test for it
  would be speculative. Recording the choice rather than implying coverage.

### What it confirmed by mutation, not by reading

| Mutant | Result |
| --- | --- |
| Glob branch reverted to literal-only | `test_a_glob_file_budget_caps_every_file…` **RED** |
| Glob branch kept, empty glob `continue`s (fails open) | `…matching_nothing_is_reported_as_missing` **RED** |
| `baseline = None` in `collect_metrics` | `…reports_a_real_reduction_end_to_end` **RED** (`None != 50.0`) |
| `reduction_percent(...) or 0` (restore the pre-#4459 bug) | `…absent_not_as_zero_percent` **RED** (`0 is not None`) |

So the "matching nothing" test is load-bearing after all — it stays green under the revert, but
it is the only thing that kills the likelier regression, a glob that silently enforces nothing.
And the #4466 end-to-end test is not tautological.

It also verified #4463 row by row: **all eight** deleted rows have an upstream description
(`SKILL.md:83, 150-157, 199-201`; `routing.md:27, 54, 55, 56`). None had zero. The only genuine
loss is "the five portable roles", compressed by `SKILL.md:200` to "defines role boundaries" —
and that count was the exact drift risk #4463 named. `CONTRIBUTING.md:109` stays true;
`git grep "What each surface does"` returns no hits, so nothing depended on the old heading.

### Filed rather than folded in

- **#4480** — `.agents/skills/README.md` is in no intent check's globs, so the rule this PR
  writes into it ("the map describes a surface only where nothing it links to already describes
  it") is stated-but-unchecked, the category #4452 says gets traded away.
- **#4481** — the references glob is written out in five places and only the new one fails
  loudly when the path moves.

### Checks after the fixes

```
Ran 230 tests in 35.234s
OK
Agent setup is valid: 126197 guidance bytes, 336 memory objects.
```

`git diff --check` clean. CI on the first push: Agent Guidance Gate passed on both
`ubuntu-22.04` and `windows-2025`, no failures.

Auto-merge still deliberately not armed.

Issue #3965, opened 2026-07-21
shaft-intellij: agentOutput/failureOutput stderr dumps bypass Verbose gating (4th path from #3918, deferred by PR #3958)


MohabMohie:
Follow-up from PR #3958 (closed #3918), which fixed 3 of the 4 raw-output-leaks-in-non-verbose-mode paths named in #3918.

## What was deferred and why

Issue #3918 named four paths where unformatted content still surfaces in non-verbose mode:
1. Non-JSON stdout passthrough rendered as milestone bubbles — **fixed by PR #3958**.
2. Buffered CLIs' (Copilot/custom) raw lines rendered regardless of Verbose — **fixed by PR #3958**.
3. Killed partial raw buffer (`stopLocalAgentStreaming`) — **fixed by PR #3958**.
4. stderr fenced dumps at terminal for non-parser/buffered runs (`agentOutput` around `ShaftAssistantPanel.java:1913-1928`, `failureOutput` around `:763-775` as of #3918's filing — re-locate exactly, line numbers have shifted since, including through PR #3961's changes to this file) — **not fixed**.

PR #3958's body says: "The issue's 4th path... is not fixed here. See 'Adjacent findings' below for why and what's needed" — but no "Adjacent findings" section was actually included in the final PR body, so the specific reasoning was never captured. Filing this issue to make sure the 4th path isn't silently dropped.

## Scope

Apply the same fix pattern PR #3958 used for the other three paths: gate `agentOutput`'s and `failureOutput`'s stderr-dump-into-terminal-answer behavior behind live Verbose state (`verboseLocalAgentOutput()`), consistent with how the sibling paths were fixed — content must remain reachable via Verbose, never silently dropped, only withheld from the compact non-verbose rendering.

## Sequencing note

Touches `ShaftAssistantPanel.java`, which is currently being modified by the in-flight sequential #3923 chain (#3921 in progress as of this filing, #3922 to follow). Pick this up once that chain lands, same as #3955, to avoid a merge conflict.

MohabMohie:
Status update: PR #3977 (merging now) implements the gating mechanism correctly and is well-tested — `agentOutput`/`failureOutput` in `AssistantLocalAgentRunner.java` now correctly withhold stderr from the compact terminal answer when `verbose=false`, threaded cleanly through a new `verbose` parameter down the whole `start(...)`/`run()` call chain.

[...]

PR #4493, opened 2026-08-04
Fix #4424: surface sandbox bootstrap failures on a successful CLI exit

### LINKED ISSUE #4424: Bug: SHAFT Assistant does not complete a requested browser recording after setup passes
Reproduced during #4299 live IntelliJ execution on 2026-08-03 after MCP, workspace, and Codex CLI setup all reported Done.\n\nWith the unsupported local-model override cleared, the Assistant accepted a request to record DuckDuckGo -> search shaft_engine -> assert the first result contains shafthq -> generate a test and page object. It did not open a managed browser and generated no recording, test, or page object. The transcript contained sandbox-setup.exe ... Access is denied (os error 5) and then only environment-instruction processing.\n\nExpected: the Assistant either drives the configured SHAFT capture tools through the requested flow or surfaces a precise actionable tool/sandbox failure.\n\nActual: the run completes without the requested artifacts or a usable result.

MohabMohie:
## Summary

`Closes #4424`. A structured-stream local-agent run (Claude/Codex) that exits 0 with a terminal event, or a buffered/custom-command run, could silently discard its stderr even when that stderr was the only account of what actually happened -- most concretely, Codex's sandbox helper failing to launch (`sandbox-setup.exe ... Access is denied (os error 5)`) before any tool could run. The transcript ended as a confident bare confirmation with no trace of the failure. This went through three review rounds; this body describes the final shipped state, not the intermediate ones (see the PR comments for that history).

## The rule being shipped

**A zero-exit run that attempted zero tool calls, touched zero files, and hit zero denials, but did emit stderr, surfaces that stderr (or a notice pointing at it) regardless of the Verbose toggle.** A recognized sandbox-bootstrap-failure shape surfaces on its own too, even alongside real tool activity (a partially-degraded run a user cannot otherwise diagnose). This is not ordinary noise (contrast the Verbose-gated stderr fence for failure paths, issue #3965): if nothing else happened, stderr is the only account of what went wrong, so withholding it makes the whole request silently impossible without a trace.

### Structured-stream path (`StructuredStreamParser`, `AssistantLocalAgentRunner.java`)

- `finalOutput(String stderr)` overload (the no-arg `finalOutput()` stays for the `terminalAnswerConsumer` side channel, where stderr is out of scope): surfaces stderr when the run attempted **zero tool calls** (new `toolCallObserved` flag) **and** touched zero files **and** hit zero denials, or independently when the stderr matches a recognized sandbox-bootstrap shape (`sandboxBootstrapFailureHeadline`) -- the recognized shape is enrichment (a nicer, specific headline) layered on the general rule, never its gate, so an unrecognized shape still surfaces via a generic headline.
- `toolCallObserved` is fed by every tool-bearing event on both mappers: Claude's `tool_use`, `server_tool_use`, and `tool_result` blocks; Codex's `command_execution`/`mcp_tool_call`/`collab_tool_call`, `web_search`, and `file_change` items. This is what keeps the rule from becoming "always surface any stderr": a run that attempted a tool call but had nothing file/denial-worthy to report still withholds incidental stderr noise exactly as issue #3965 established.
- `activitySummary()`'s "no silent bare confirmation" guard renders an explicit "no files, no tool calls" line for the same silent-run shape instead of staying blank.
- `run()`'s success branch never hands raw NDJSON stdout to the buffered composer below: a structured CLI that exits 0 without ever emitting a terminal event has only machine NDJSON in `rawStdout`, mirroring the guard the failure branch a few lines below already has.

### Buffered/custom-command path (`bufferedSuccessOutput`, new)

- `run()`'s buffered success answer (Copilot's default command, or any hand-typed custom command) now goes through a dedicated `bufferedSuccessOutput(stdout, stderr, verbose)` instead of `agentOutput`. `agentOutput` itself is untouched -- its success-path short-circuit is a separate, deliberately-tested contract for its other caller, `runCompactPreamble`'s low-stakes auxiliary status line.
- `bufferedSuccessOutput` surfaces stderr regardless of Verbose when stdout is blank (the buffered analogue of `toolCallObserved == false`: nothing else happened). When stdout has real content, stderr stays Verbose-gated exactly as before -- there is no reliable signal in this path to distinguish a real failure from ordinary chatter once the CLI has already produced real output, so Verbose off keeps withholding it there.

## Test plan

Every fix went through a RED-first cycle; the exact assertion failures for each round are recorded in the PR comments below (round 2's four general-case tests plus round 3's nine tests covering F1-F4). Summary:

- `AssistantLocalAgentRunnerVerboseStreamTest`: 68 tests, 0 failed.
- Full `AssistantLocalAgentRunner*` wildcard suite plus `AssistantCommandTest` (the test an earlier attempt at the buffered-path fix broke, then fixed by splitting `bufferedSuccessOutput` out of `agentOutput` instead of changing `agentOutput`): green.
- Byte-identical regression for an ordinary successful run (`successfulRunWithoutSandboxBootstrapStderrStaysByteIdenticalToTodaysOutput`): still passes, exact literal `"plain answer\n\n{\"usage\":{\"input_tokens\":1,\"output_tokens\":1}}"`.
- Full module `./gradlew test`: 1309 tests, 21 skipped. Re-run multiple times across every round; each run fails a different 2-3 tests, always in `ShaftPanelSetupTest`/`ToolApprovalPromptPanelTest`, always the same `threadUtil.kt:32` (`SwingTimerWatcherExtension`) / `ThreadLeakTracker.java:206` Swing-timer / thread-leak signature, never touching the changed files -- confirmed pre-existing, non-deterministic test-infrastructure flakiness (tracked separately by the reviewer, not part of this PR).

Environment note for reviewers: `shaft-intellij` is a Gradle module outside the Maven reactor (`mvn` will not run these tests); the module also requires JDK 21 for its Gradle daemon (JDK 25 crashes it on Windows, per its own `settings.gradle.kts` check).

Not arming auto-merge -- waiting for the next independent review pass.

### DIFF
--- a/shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantLocalAgentRunner.java
+++ b/shaft-intellij/src/main/java/com/shaft/intellij/ui/AssistantLocalAgentRunner.java
@@ -687,7 +687,7 @@ final class AssistantLocalAgentRunner {
                 if (success) {
                     output = streamParser != null && streamParser.hasTerminalEvent()
                             ? streamParser.finalOutput(rawStderr)
-                            : agentOutput(true, rawStdout, rawStderr, "", verbose);
+                            : bufferedSuccessOutput(rawStdout, rawStderr, verbose);
                 } else if (streamParser != null) {
                     // A structured (stream-json / --json) CLI writes machine NDJSON to stdout, so on a
                     // failed run the raw stream must never reach the transcript verbatim -- rerun/codegen
@@ -883,6 +883,12 @@ final class AssistantLocalAgentRunner {
         // after it), then carried through composeOutput's trailing metadata line so
         // AssistantLocalAgentRunner.parseQuestion can recover it, mirroring planProposal above.
         private AssistantQuestion structuredQuestion;
+        // Whether any tool call was ever attempted, regardless of its outcome (issue #4424 round 2):
+        // distinct from filesTouched/permissionDenialsByTool, which only record a tool call's *result*.
+        // A run that attempted a (successful, non-mutating) tool call must not be treated as "silent"
+        // by finalOutput(String) just because it had nothing file/denial-worthy to report -- only a run
+        // that never got to attempt anything at all is that shape.
+        private boolean toolCallObserved;
 
         StructuredStreamParser(Format format) {
             this.format = format;
@@ -926,27 +932,39 @@ final class AssistantLocalAgentRunner {
         }
 
         /**
-         * Same as {@link #finalOutput()}, plus consulting {@code stderr} for an environment/sandbox
-         * bootstrap failure (issue #4424). {@code run()}'s success branch used to call the no-arg
-         * overload unconditionally, discarding {@code rawStderr} entirely once a terminal event had
-         * arrived -- so a CLI that exited 0 after its sandbox helper failed to launch (Codex's
-         * {@code sandbox-setup.exe ... Access is denied (os error 5)}) produced a confident-looking
-         * answer with no trace that no tool had actually run. That is not ordinary noise (contrast the
-         * Verbose-gated stderr fence in {@link #failureOutput}, issue #3965): it means the entire
-         * request was silently impossible, so {@link #sandboxBootstrapFailureNotice} surfaces one
-         * actionable line regardless of the Verbose toggle. A {@code null}/blank/non-matching {@code
-         * stderr} leaves {@code core} untouched, so this overload is byte-identical to {@link
-         * #finalOutput()} for the overwhelmingly common case. This overload is deliberately separate
-         * from (rather than replacing) {@link #finalOutput()}, which stays in use where {@code
-         * terminalAnswerConsumer} recovers a cancelled/killed run's answer with no stderr in scope.
+         * Same as {@link #finalOutput()}, plus consulting {@code stderr} (issue #4424). {@code run()}'s
+         * success branch used to call the no-arg overload unconditionally, discarding {@code rawStderr}
+         * entirely once a terminal event had arrived. The general rule (round 2 of this issue, after an
+         * adversarial review showed a single-string {@code sandbox-setup} match was both incomplete --
+         * an unrelated MCP-startup failure hit the identical silent shape -- and fragile -- a bare
+         * naming-convention drift such as {@code SandboxSetup.exe} defeated it): <b>a zero-exit run that
+         * attempted zero tool calls, touched zero files and hit zero denials, but did emit stderr, must
+         * surface that stderr -- or a notice pointing at it -- regardless of the Verbose toggle.</b> That
+         * is not ordinary noise (contrast the Verbose-gated stderr fence in {@link #failureOutput}, issue
+         * #3965): if literally nothing else happened, the stderr is the only account of what went wrong,
+         * so withholding it makes the whole request silently impossible without a trace. {@code
+         * toolCallObserved} is what keeps this from becoming "always surface any stderr": a run that DID
+         * attempt a tool call but simply had nothing file/denial-worthy to report keeps withholding
+         * incidental stderr noise exactly as issue #3965 established. {@link
+         * #sandboxBootstrapFailureHeadline} still recognizes the sandbox-bootstrap shape for a more
+         * specific, actionable headline, but only as an enrichment on top of the general rule -- never
+         * as its gate -- so an unrecognized shape still surfaces via the generic headline. A {@code
+         * null}/blank stderr, or any tool-call/file/denial activity, leaves {@code core} untouched, so
+         * this overload is byte-identical to {@link #finalOutput()} for the overwhelmingly common case.
+         * This overload is deliberately separate from (rather than replacing) {@link #finalOutput()},
+         * which stays in use where {@code terminalAnswerConsumer} recovers a cancelled/killed run's
+         * answer with no stderr in scope.
          */
         synchronized String finalOutput(String stderr) {
             String core = answer.strip();
-            String notice = sandboxBootstrapFailureNotice(stderr);
-            if (notice != null) {
+            boolean silentRun = filesTouched.isEmpty() && permissionDenialsByTool.isEmpty() && !toolCallObserved;
+            String stderrText = stderr == null ? "" : stderr.strip();
+            boolean surfaceStderr = silentRun && !stderrText.isBlank();
+            if (surfaceStderr) {
+                String notice = stderrNotice(stderrText);
                 core = core.isBlank() ? notice : core + "\n\n" + notice;
             }
-            return composeOutput(core, notice != null);
+            return composeOutput(core, surfaceStderr);
         }
 
         /**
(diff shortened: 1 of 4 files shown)


MohabMohie:
## Round 2: independent adversarial review findings, addressed

The reviewer's round-1 verdict confirmed RED/GREEN/the overload split/the pre-existing flakiness claim, but proved the round-1 scoping decision (narrowing `activitySummary`'s widening) was wrong in a more fundamental way: the fix itself was fitted to one observed string. Three findings, all fixed except one deliberately narrowed sub-case (explained below).

**Finding 1 -- the general bug was still open.** Probe: Codex, zero tool calls, stderr `Error: MCP server 'shaft-tools' failed to start (ECONNREFUSED 127.0.0.1:5173)`, Verbose ON -> silent `ok\n\n{"usage":...}`. Same shape as #4424, different cause.

**Finding 2 -- detection fitted to one string.** Probe: `SandboxSetup.exe failed to launch: Access is denied (os error 5)` (hyphen removed) -> no notice, despite the exact `Access is denied (os error 5)` text being present verbatim.

**Finding 3 -- the buffered/custom-command path was unreachable by the round-1 fix.** `agentOutput` returns `stdout.strip()` on success without ever consulting stderr, regardless of Verbose. Probe: `customCommandInvocation()`, stdout `ok`, matching `sandbox-setup` stderr, Verbose ON -> `ok`, zero trace.

### Fix

Stated the actual rule instead of pattern-matching the failure: **a zero-exit run that attempted zero tool calls, touched zero files, and hit zero denials, but did emit stderr, surfaces that stderr (or a notice pointing at it) regardless of Verbose.** The `sandbox-setup` shape is now purely an enrichment (a nicer headline) on top of that rule, never its gate -- an unrecognized shape still surfaces via a generic headline, closing finding 2 without needing broader string-fuzzing.

- New `toolCallObserved` flag on `StructuredStreamParser`, fed by both `ClaudeStreamEventMapper`/`CodexStreamEventMapper` on every tool-call event regardless of outcome. This is what keeps the rule from becoming "always surface any stderr": a run that attempted a tool call but had nothing file/denial-worthy to report still withholds incidental stderr noise, exactly as issue #3965 established. Pinned by new boundary tests on both mappers (`successfulCodexRunWithAToolCallButNothingToReportWithholdsIncidentalStderrWhenVerboseIsOff`, `successfulClaudeRunWithAToolCallButNothingToReportWithholdsIncidentalStderrWhenVerboseIsOff`).
- `sandboxBootstrapFailureNotice` -> `stderrNotice` + `sandboxBootstrapFailureHeadline`: the specific "Environment setup failed" headline is now enrichment-only; a non-matching stderr still surfaces via a generic "no tool ran, but the CLI reported this on stderr" headline, with the raw stderr always included as a fenced block either way. `sandboxBootstrapFailureHeadline` also matches "sandbox"+"setup" as separate substrings now (order/separator-agnostic) instead of the literal `sandbox-setup` token, closing finding 2's specific naming-drift case too, though it's enrichment, not the gate that matters.
- Finding 3: `agentOutput`'s short-circuit (bare `stdout` on success, unconditionally) turned out to be a **deliberate, separately-pinned contract** for its other caller, `runCompactPreamble`'s low-stakes auxiliary status line (`AssistantCommandTest#directLocalAgentOutputPrefersStdoutOnSuccessAndIncludesStderrOnFailure`, which I broke and had to revert). Rather than changing `agentOutput` itself, `run()`'s success branch now calls a new, dedicated `bufferedSuccessOutput(stdout, stderr, verbose)` that folds `stdout` and `verbose`-gated `stderr` together consistently, closing the actual reported gap without touching the other caller's tested contract.

### Scoping decision on finding 3's Verbose=false sub-case (flagging per "say so before changing course")

The reviewer's probe used Verbose ON. I fixed that. I did **not** extend `bufferedSuccessOutput` to ungate Verbose=false, and did not file a followup issue for it either -- on reflection this isn't a demonstrated gap, it's a deliberate choice: unlike the structured-stream path, which has `toolCallObserved` as a strong, protocol-level "nothing else happened" signal, the buffered path (Copilot's default command, or *any* hand-typed custom command) has no such signal -- only two raw text blobs. Unconditionally surfacing stderr there whenever Verbose is off would defeat issue #3965's noise-suppression for arbitrary, unpredictable custom CLIs that routinely log benign chatter to stderr. Verbose=false keeps the exact, already-tested #3965 behavior in this path.

### The two "stays clean" tests

`finalOutputStaysCleanWhenNothingWasWrittenAndNothingWasDenied` and `finalOutputStaysCleanForCodexWhenNothingWasWrittenAndNothingFailed` were updated deliberately, not weakened: they now build their stub via `successfulRun(..., "", true)` (explicit blank stderr, Verbose ON) instead of the shorter `finalOutput` helper, so the "genuinely nothing to report" precondition is textual, not an implicit stub default. Both still assert the run stays clean -- they pin the other side of the distinction from the new "silently lost activity" tests.

### RED for the general case (observed before any production change)

```
successfulRunWithNonBootstrapStderrAndZeroToolCallsStillSurfacesItRegardlessOfVerbose() FAILED
  AssertionFailedError: Any stderr on an otherwise-silent successful run must surface, not only the
  sandbox-setup shape: ok

  {"usage":{"input_tokens":1,"output_tokens":1}} ==> expected: <true> but was: <false>

successfulRunWithNonBootstrapStderrSurfacesItEvenWithVerboseOff() FAILED
successfulRunWithASandboxSetupNamingVariantStillSurfacesTheStderr() FAILED
successfulRunWithNoAnswerTextAtAllStillSurfacesTheStderrNoticeCleanly() FAILED
  (all four: actual output was "ok\n\n{"usage":...}" / "\n\n{"usage":...}" -- silent)

56 tests completed, 4 failed
```

RED for finding 3, buffered path (observed after the general fix, before the `bufferedSuccessOutput` fix):
```
agentOutputSurfacesStderrAlongsideNonBlankStdoutWhenVerboseIsOn() FAILED
  expected: <ok

  sandbox-setup.exe failed to launch: Access is denied (os error 5)> but was: <ok>

customCommandSurfacesStderrAlongsideNonBlankStdoutWhenVerboseIsOn() FAILED
  A custom command's stderr must surface alongside its stdout when Verbose is on: ok ==> expected: <true> but was: <false>
```

### GREEN

`AssistantLocalAgentRunnerVerboseStreamTest`: 59 tests, 0 failed. Full `AssistantLocalAgentRunner*` suite green. `AssistantCommandTest` (the test my first attempt at finding 3 broke, then fixed by scoping to `bufferedSuccessOutput` instead of `agentOutput`) green again.

### Both review probes re-run against the new build

- MCP-server probe (finding 1), Verbose ON: `successfulRunWithNonBootstrapStderrAndZeroToolCallsStillSurfacesItRegardlessOfVerbose` -- observed `result.output()`:

  > ok
  >
  > **Note:** no tool ran for this request, but the CLI reported this on stderr:
  >
  > &#96;&#96;&#96;
  > Error: MCP server 'shaft-tools' failed to start (ECONNREFUSED 127.0.0.1:5173)
  > &#96;&#96;&#96;
  >
  > ---

- `SandboxSetup.exe` naming-variant probe (finding 2), Verbose OFF: `successfulRunWithASandboxSetupNamingVariantStillSurfacesTheStderr` -- observed `result.output()`:

  > ok
  >
  > **Environment setup failed:** the sandbox helper could not launch, so no tool could run for this request. Ask an administrator to grant it access, then try again.
  >
  > &#96;&#96;&#96;
  > SandboxSetup.exe failed to launch: Access is denied (os error 5)
  > &#96;&#96;&#96;
  >
  > ---

Both now produce the notice, as required.

### Byte-identical regression still holds

`successfulRunWithoutSandboxBootstrapStderrStaysByteIdenticalToTodaysOutput` still passes unchanged: exact literal `"plain answer\n\n{\"usage\":{\"input_tokens\":1,\"output_tokens\":1}}"`.

### Full module Gradle run

`./gradlew test`: 1300 tests, 21 skipped. Re-run 3 times across this round; each run failed a **different** 2-3 tests, always in `ShaftPanelSetupTest` or `ToolApprovalPromptPanelTest`, always the same `threadUtil.kt:32` / `ThreadLeakTracker.java:206` Swing-timer-disposal signature, never `AssistantLocalAgentRunner*` or `AssistantCommandTest`. Confirmed pre-existing, non-deterministic, unrelated test-infrastructure flakiness (same conclusion as round 1, now reinforced across 5 total runs with 5 different failing sets).

Not arming auto-merge -- per instructions, waiting for the next independent review pass.

MohabMohie:
## Round 3: OPUS re-review findings, addressed

Thanks for the precision here -- F1 in particular was a real, shippable regression (successful runs with Verbose off getting noise re-injected), not a limitation to document. All four findings fixed; PR body rewritten (F6) to describe the final state instead of round 1's reversed reasoning.

### F1 (blocking) -- hoisted `recordToolCallObserved()` to every tool-bearing variant

Added to Codex's `web_search` and `file_change` items (including the no-`changes`-array edge case you probed) in `CodexStreamEventMapper.java`, and to Claude's `server_tool_use` block (previously entirely unrecognized) and `tool_result` block (proves a call happened even if the requesting event was missed) in `ClaudeStreamEventMapper.java`.

### F2 -- fixed, not documented

`bufferedSuccessOutput` now checks `stdoutText.isBlank()` and surfaces stderr unconditionally in that case -- the buffered analogue of `toolCallObserved == false`, exactly as you pointed out `sections.isEmpty()` already computed. The narrower, correct claim (non-blank stdout + Verbose off keeps withholding stderr) is what's now documented; the PR body no longer claims "no reliable signal" for the whole path.

### F3 -- your ruling implemented as stated

`finalOutput(String stderr)` now computes `recognizedBootstrapFailure` independently of `silentRun` and surfaces stderr when either is true. `silentRun` still gates the general unread-stderr rule; the recognized shape fires on its own alongside it, closing the partially-degraded-run gap.

### F4 -- guard carried across

`run()`'s success branch now passes `""` instead of `rawStdout` to `bufferedSuccessOutput` whenever `streamParser != null` (a structured CLI), mirroring the failure branch's guard three lines below. A genuinely buffered/custom command (`streamParser == null`) is unaffected.

### RED observed before each fix (9 new tests, all against the round-2 committed sources)

68 tests completed (59 existing + 9 new), 9 failed:

- `codexFileChangeItemWithNoChangesArrayStillCountsAsAToolCallObserved()` -- actual output: `ok` + the generic "no tool ran for this request" notice + a fenced `some incidental progress chatter` + the "No files were crea..." footer (the exact false-positive shape).
- `claudeToolResultWithNoPriorToolUseEventStillCountsAsAToolCallObserved()` -- same shape, `done` + the notice.
- `claudeServerToolUseBlockCountsAsAToolCallObserved()` -- same shape, `done` + the notice.
- `codexWebSearchItemCountsAsAToolCallObservedSoIncidentalStderrStaysWithheld()` -- same shape, `ok` + the notice.
- `customCommandWithBlankStdoutSurfacesStderrEvenWithVerboseOff()` -- `expected: <true> but was: <false>`, actual output was empty string.
- `bufferedSuccessOutputSurfacesStderrWhenStdoutIsBlankRegardlessOfVerbose()` -- `expected: <sandbox-setup.exe failed to launch: Access is denied (os error 5)> but was: <>`.
- `successfulCodexRunWithoutATerminalEventNeverLeaksRawNdjson()` -- actual output was the two raw NDJSON lines verbatim (`{"type":"thread.started",...}` / `{"type":"turn.started"}`).
- `successfulCodexRunWithoutATerminalEventStillSurfacesStderrInsteadOfRawNdjson()` -- `expected: <false> but was: <true>` (the raw-NDJSON-contains check), output was the raw lines with no stderr.
- `recognizedSandboxBootstrapFailureSurfacesEvenAlongsideSuccessfulFileActivity()` -- actual output was `ok` + the file-activity footer, no trace of the sandbox stderr at all.

### GREEN after the fixes

`AssistantLocalAgentRunnerVerboseStreamTest`: 68 tests, 0 failed. `AssistantLocalAgentRunner*` wildcard suite + `AssistantCommandTest`: green. Full module `./gradlew test`: 1309 tests, 21 skipped, 2 failed -- `ShaftPanelSetupTest.directCodegenToolDispatchArmsSameStickyReviewGateAsRecordFlow` and `ShaftPanelSetupTest.assistantCancelButtonArmsKillForActiveInvocation`, both `threadUtil.kt:32`, a different pair than any prior run -- the same pre-existing flakiness you're filing separately, not touching any changed file.

Not arming auto-merge.

Issue #4538, opened 2026-08-04
Every task should start with a draft PR and push status-update commits every 5 minutes


MohabMohie:
Owner standard, set 2026-08-04 after it was tried once and worked: **every new task opens a draft pull request at the start, and pushes to it every 5 minutes with an explicit status-update commit message.**

## What prompted it

A delegate working an eight-issue batch had been running 25 minutes with nothing pushed. Everything it had completed existed in one worktree on one machine. Told to push and open a draft immediately rather than wait for completion, it did — and five of the eight issues became visible and recoverable within a minute, along with which three remained and why.

The general point: a branch is recoverable only by whoever can see the machine it lives on. A pushed draft is recoverable by anyone. Waiting until work is presentable optimises for the reviewer's first impression and trades away the only property that matters when an agent dies mid-run. Three of eight done and visible beats seven of eight done and stranded.

## The standard

1. **Open the draft PR at the start of the task**, not when the work looks finished. Carry the `Closes` lines from the outset so the scope is on record even before the code is.
2. **Push at least every 5 minutes.** A delegate has a clock; this is a real interval, not a figure of speech.
3. **The commit message states status, not just content** — what is done, what remains, what is next, and what is blocking. `WIP: #4509 done (git-view counting + 3 tests); next #4498, blocked on re-measuring after #4524 merge` is the shape. A bare `wip` is not, because an unexplained pushed commit is nearly as bad as an unpushed one: the next agent cannot tell whether a half-finished edit was deliberate.
4. **Draft status is what makes this safe.** Frequent pushes do not signal readiness, and the independent adversarial review still gates arming. Nothing about this shortens the review.

## What needs deciding rather than assuming

- **Where it belongs.** `delegation.md` governs dispatch and is the natural home, but its size counts toward the always-loaded budget that #4498 is re-deriving — so the wording has to earn its bytes.
- **Whether it is enforceable or only stated.** #4530 and #4536 both record rules that failed precisely because their only enforcement was an agent remembering. A stated interval with no trigger is the same shape a third time. Worth deciding once for all three rather than three times: is there a check that can observe a delegate's push cadence, or is the honest answer that this one is dispatch-prompt boilerplate rather than a check?
- **Squash-merge interaction.** The repository squash-merges, so a stream of status commits collapses to one message on `main`. That is fine — the commits exist for the resumption case, not the history — but it should be stated so nobody later "tidies" the practice away as noise.

## Relationship to the rest of the group

[...]

Issue #4536, opened 2026-08-04
Findings must be ticketed, but nothing requires the state needed to resume, so a lost agent loses the run


MohabMohie:
Owner requirement: every finding and enough state must be on GitHub continuously, so that when an agent runs out of tokens another agent picks the work up. Standing, not a one-off.

## What already works

`AGENTS.md` → Working rules: *"Search before filing every non-blocking adjacent finding as an issue; do not silently expand scope or leave it only in chat."*

That rule holds and was honoured. Fourteen findings became issues during the 2026-08-04 run rather than chat lines.

## What has no rule, and demonstrably failed

Findings are covered. **Decisions and in-flight state are not.** Nothing in tracked guidance requires an agent to record:

- **Owner decisions made in session.** Measured: #4504 had *zero* comments while an agent was actively implementing the owner's choice to split the playbook deliberately. #4498 had one comment, and it was a measurement, not the decision. Both choices existed only in a delegate's dispatch prompt and in a conversation. Losing that conversation loses the decision, and the next agent re-asks a question the owner already answered — or worse, guesses differently.
- **What is in flight.** Branch name, which issues are batched into it, whether a PR exists yet, what was deliberately excluded from scope.
- **Sequencing constraints.** Which files are held by which open PR, and therefore what order the work must follow. This is the state most expensive to rediscover, because rediscovering it wrongly produces a conflict rather than an error.
- **Standing constraints given in session.** Instructions that bind every agent but arrived mid-run, so they are in no tracked file — this session produced four, including one that forbids full test suites and one that sets the check-in cadence (#4530).

The asymmetry is the point: a *finding* is about the code and has a natural home. A *decision* is about the work and has no home at all unless someone invents one.

## Why the obvious fix is not enough

"Comment on the issue when you decide something" is a recall-based rule, which is the same shape as the cadence gap in #4530 — and that one failed within the same session it was noticed. The rule needs a defined trigger, not diligence.

Trigger points worth considering, all of which are already moments an agent stops and does something:

- On dispatching a delegate: the ticket gets the branch, the batched scope, and the exclusions.
- On receiving an owner decision: it goes on the affected issue before the work starts, not after it lands.
- On discovering a sequencing constraint: it goes on the tracker.
- On a session ending or an agent being stopped: current state goes on the tracker.

[...]

Issue #4535, opened 2026-08-04
POLICY_RECORD_ALLOWLIST entries carry no enforced reason and no expiry


MohabMohie:
Carved out of #4516, which is closed by the PR that fixes its two named
residuals (path-form citations, and `background agent` in the subject list).
This third item from that issue is a design decision rather than a defect, so
it is recorded here rather than decided by the implementer.

## The gap

`POLICY_RECORD_ALLOWLIST` in `tests/scripts/test_agent_harness_portability.py`
is `dict[reason, frozenset[object-id]]`. An entry is one opaque id in a set.
Two properties it does not have:

- **no enforced reason.** The prose justifying each entry lives in a comment
  above it, by convention. A comment is not a field, so nothing fails when one
  is absent, wrong, or copied from the entry above.
- **no expiry.** `test_every_allowlist_entry_is_individually_load_bearing`
  fails an entry whose object stopped tripping the scan, which is real
  staleness protection, but an entry whose *justification* went stale while the
  prose still trips looks identical forever.

Why it matters more than an ordinary nit: from the outside there is no
difference between an honest narrative escape and an agent quietly exempting a
policy restatement to get to green. Both are the same one-line diff. The
allowlist is the whole reason #4484 could drop the grammar -- it is what makes
a false positive cheap to clear without weakening the check, which iron law 4
forbids -- so its own trust boundary is load-bearing.

The list is now two entries rather than one (#4516 added the second), which is
what makes this concrete rather than hypothetical.

## Shape of the fix, if taken

`dict[reason, dict[object-id, why]]` reads the same at every existing call site
(`identifier not in permitted.get(reason, {})` is unchanged for a dict), so the
change is the literals plus one assertion that each `why` is non-empty. Roughly
ten lines and five fixture edits.

Worth deciding at the same time: whether the reason should have to name the
sentence in the object that trips the scan, which is what makes a stale
justification visible.

Refs #4516, #4484, #4519

MohabMohie:
Queued on `ChaosEngine/portability-scan-gaps-4534-4535`, cut from `origin/main`, grouped with #4534.

[...]

Issue #4534, opened 2026-08-04
Numbering check cannot see a rule deleted from the end of a pinned list


MohabMohie:
Measured and disclosed during the fix for #4490 (PR #4524), which closed the repeat-of-`1` and `1)` delimiter holes. This one survives.

## The surviving exploit

Deleting the **last** item of an ordered list in a pinned section leaves the numbering contiguous — `1, 2, 3` becomes `1, 2` — so `numbering_defects` reports nothing. The writer's own mutation table records it:

| Exploit | Result on the fixed head |
| --- | --- |
| Paren-delimited + **middle** rule deleted | RED, `rule numbering is [1, 3], not [1, 2]` |
| Paren-delimited + **last** rule deleted | **still green** |

It is now stated in the docstring, which is the right thing to have done rather than leaving it silent. But #4490's thesis is that a weakening shape nothing catches is a weakening shape, and the honest disclosure does not change what an agent under pressure can do to the file.

## Why it is narrower than it first looks

Deleting a **pinned clause** from the end is still caught, by `test_deleting_any_pinned_clause_is_reported` and the entrypoint pin tests — a different mechanism with a different failure signature. The exposed surface is an ordered item in a pinned section that is not itself a pinned clause. That is a smaller target than "any last rule", and it is why this is a follow-up rather than a merge blocker for #4524.

It is not empty, though. The pinned-clause set is a list someone maintains, so the gap is exactly the set of ordered items that nobody has pinned yet — which is the normal state of a newly added rule.

## What a fix has to reckon with

Numbering alone cannot detect a truncation; `1..N-1` is a valid `1..N`. So this needs either an expected-length signal per pinned section, or the pinned-clause mechanism extended to cover ordered items, or an explicit decision that end-truncation is covered by clause pinning and the numbering check is not the right instrument. Any of those is defensible; picking one silently is not.

Worth deciding alongside #4531, which asks the same question one level up — what defines the set being checked — and #4532, which is the same scan gap in another file.

Grouped under #4533 with the rest of the open harness work: one branch, one PR, after the current batch.

Related: #4490, #4524, #4531, #4532, #4533.

MohabMohie:
Two corrections from the verification pass on PR #4524, both narrowing this issue and both in the writer's favour. The original framing above is wrong on each.

[...]

PR #4539, opened 2026-08-05
One skill, whole harness: close the reachability boundary and the harness-governance batch

### LINKED ISSUE #4529: Workflow-timeout test asserts on four files but only guards two, so three timeouts can be lowered back green
Found by the adversarial review of PR #4523 (#4494), which cleared that PR to merge. None of the three below blocks it; all are cheaper as separate issues than as a re-opened PR.

## 1. Verification gap — the timeout assertion is not bound for three of its four files

`.github/workflows/pr-gate.yml`, the path filter feeding the new test.

The test added by #4523 asserts timeouts on four workflow files. Three of those four match neither the `agent_guidance` nor the `infra` path filter, so **editing them does not run the test that guards them**. Concretely: lowering `publish-intellij-plugin.yml`'s timeout back to 10 minutes would land green, which is precisely the regression the test was written to prevent.

This is the verification-gap shape the harness cares about — a check whose pass is independent of the thing it verifies, for three quarters of its own subject matter.

Fix: the assertion belongs in the `Workflow Timeout Guard` job in `pr-gate.yml`, which already runs on the right trigger, rather than in a job gated behind path filters that do not cover the files it asserts on.

## 2. The PR body's hazard argument is inverted

#4523's body attributes a "silent version drift" hazard to caching. It is not a property of caching — it is a property of the **un-pinned status quo**, which caching neither creates nor worsens.

The second half of that argument, the "wrong IDE" risk, is unreachable: `shaft-intellij/build.gradle.kts` pins both. Worth correcting in the record, since a wrong rationale attached to a right change tends to get cited later as if it were established.

Overlaps #4525, which covers the unpinned Plugin Verifier itself.

## 3. The cache dismissal overclaims

"Outside the Gradle cache" was treated as "uncacheable." It is not. The verifier's own download directory is `actions/cache`-able exactly the way `cache-maven-repo` already handles `~/.m2`.

Whether it *should* be cached is a separate question given #4526's finding that Actions cache is already near its budget — but the reason to decline should be the budget, not an incorrect claim that it cannot be done.

## What the review confirmed, for the record

The change itself is correct. Five separate attacks found nothing: no surviving `maven_retry` references anywhere in the tree; the publish split verified against the IPGP 2.17.0 jar, where `PublishPluginTask`/`SignPluginTask` pull only `buildPlugin` and never `verifyPlugin` or `check`, so the signed zip is the verified zip; no retry-masking, since the original job log shows `check` passing before `:verifyPlugin` fails and Gradle is fail-fast with no `--continue` in the repo; the timeout arithmetic fits, measured across 38 recent jobs at 5.1-6.3 minutes against a 20-minute budget; and 3 of the 4 new tests fail on revert.

Related: #4494, #4523, #4525, #4526.

### LINKED ISSUE #4531: Harness reachability is bounded by a hand-maintained glob list and enumerates three vendors, so "any agent on any machine" is not yet guaranteed
Owner requirement, stated directly: *any machine running any model from any agent and any manufacturer must be able to access 100% of the harness by loading the single act-as-mohab skill.*

The mechanism to do this already exists and is good. `tests/scripts/test_agent_harness_reachability.py` walks markdown links from `.agents/skills/act-as-mohab/SKILL.md` and reports unreachable elements; `harness_reachability.element_globs` in `scripts/ci/agent_guidance_budget.json` lists 34 globs covering guidance, adapters, roles, hooks, guards, MCP config, the enforcement scripts, the tests that police them, the plugin manifest, and `pr-gate.yml` itself. `exemptions` is empty and deliberately so — both categories that had a case for exemption were made reachable instead. `EXPECTED_ELEMENT_COUNT` is 100.

This issue is not a redesign. Three specific things stop that mechanism from delivering the requirement as stated.

## 1. The boundary is a hand-maintained list, so a new harness surface is invisible rather than unreachable

The test's own docstring says it:

> `element_globs` bounds the whole question. A harness surface added under a path no glob covers is not an element, so it cannot be reported as an [unreachable one] ... the failure mode most likely to bite next.

A disclosed gap is better than a hidden one, but it is still the gap. Nothing fails when someone adds a harness surface under an uncovered path — the count simply does not move, and `EXPECTED_ELEMENT_COUNT` is an equality assertion against a number that only rises when a human remembers to raise it.

This is not hypothetical. #4530 proposes a check-in cadence mechanism; wherever it lands, no current glob is guaranteed to cover it, so the harness would grow a component that the reachability guarantee cannot see.

An audit of every tracked path that falls under the wide umbrella of *harness*, diffed against what `element_globs` actually matches, is the first deliverable. Each uncovered path then gets either a glob or a stated exemption — the exemption mechanism exists and is currently unused, which is the honest place to record a deliberate no.

## 2. The walk is not run by the command agents are told to run

Also from the docstring: the walk is not run by `validate_agent_setup.py`. Only its parity row is. So

```
py -3 scripts/ci/validate_agent_setup.py --skip-external
```

— the check `AGENTS.md` names for guidance and memory work, and the one an agent on a machine with no CI will actually run — exits 0 on a fully unreachable harness.

That is the verification-gap shape: a check whose pass is independent of the thing it verifies. It matters more here than usual because the owner's requirement is explicitly about *any machine*. A guarantee enforced only by a GitHub Actions job is not a guarantee for a machine that never opens a pull request.

## 3. The design enumerates vendors rather than generalising over them

`host_skill_metadata_globs` names exactly three: `codex`, `claude`, `grok`. `element_globs` likewise spells out `.claude/**` patterns, `.codex/*` patterns, and `.github/copilot-instructions.md` one by one.

The requirement says *any manufacturer*. A fourth vendor dropping its own adapter in gets no coverage until someone edits the list — and per gap 1, nothing tells them to. Whether that is acceptable is a real decision: enumerating vendors is defensible if the enumeration is enforced somewhere, and indefensible if it silently degrades. It should be decided rather than inherited.

## Deliverables

1. The uncovered-path inventory, with a per-path decision: new glob, or stated exemption with a reason.
2. Close gap 2 — the walk, or an equivalent, must run under the local command, so a machine with no CI can learn its harness is unreachable.
3. A stated decision on gap 3: vendor-general, or vendor-enumerated with the enumeration itself enforced.

A read-only audit covering all three is running now and will supply the inventory and mutation evidence.

## Sequencing

`element_globs` lives in `scripts/ci/agent_guidance_budget.json`, which is currently held by the in-flight harness batch (#4498 re-derives the always-loaded ceiling in that same file). This work follows that batch rather than racing it. #4530 also lands after, for the same reason.

Related: #4485, #4488, #4489, #4481, #4490, #4498, #4530.

### LINKED ISSUE #4540: The retrieval discipline fires too late: its trigger lives behind the load it is meant to precede, and no skill defines depth
Owner directive, 2026-08-05: **Graphify, MemPalace and `.memory` must all be source-controlled and used in every session, with the extent of use scaled to the size of the change — defined in a custom retrieval (RAG) skill that is part of the harness.**

## The defect, proved by the session that found it

The rule is not missing. `references/routing.md` carries a **Knowledge before discovery** table with a per-store trigger:

| Question | Store | Query it when |
| --- | --- | --- |
| Has this constraint or gotcha already bitten us? | native Memory | Required before any non-trivial change |
| What happened around this before, and what does it touch? | MemPalace | The change spans entities, sessions, or a history you were not part of |
| What calls or depends on this? | Graphify | Blast radius is unknown |

It even closes with: *"Skipping a store whose trigger fired, without saying so, is the failure this table exists to prevent."*

**The table is behind a load it is supposed to precede.** `routing.md` is reached only when the entrypoint routes a deliverable to a surface. Session-start retrieval is not a deliverable, so nothing pulls the file in before discovery begins. The entrypoint itself carries only one prose sentence — *"Routing also orders knowledge retrieval. Query the applicable stores before broad manual discovery"* — which names no store, no trigger, and no depth.

Measured, not hypothesised: the session that filed this loaded the entrypoint, ran triage, and went straight to `gh issue list`, `git ls-files` and `rg` — textbook broad manual discovery — without loading `routing.md`, because no routing row is needed to start looking. The owner caught it, not the harness. When the stores were finally queried, MemPalace returned in one call the fact that an earlier entrypoint carried an enforced `## Skill routing` section with these exact triggers inline, and `.memory` holds `constraint.memory-and-graphify-are-a-mandatory-session-entry-point`. Both were available before any of the manual discovery, and both would have shortened it.

This is the same shape as #4530, #4536 and #4538: a rule whose only enforcement is an agent remembering. It is the fourth instance, which is the argument for deciding the trigger question once for all of them rather than four times.

## What the skill has to define

1. **Depth by change size**, not by reflex. Triage already computes blast radius and reversibility in the entrypoint; retrieval depth should be a function of that, so a one-file change pays a smaller retrieval cost than a public-contract change. Querying all three stores for a typo is the failure that gets the whole rule ignored.
2. **The completion half.** The recovered earlier entrypoint paired session start with a completion step — `memory remember` for new gotchas, flag a Graphify refresh on structural change, mine the session into MemPalace. The current Learning loop table covers where a learning goes but not that the indexes themselves need refreshing.
3. **Degraded mode.** Which stores are optional on a machine that has none of the MCP servers, since the requirement is any agent on any machine. Source-controlled *configuration* is portable; a locally-built index is not, and `AGENTS.md` already forbids committing runtime indexes.
4. **Where the trigger fires.** Whatever the answer, it cannot be a file reached only by routing.

## Source-control status, measured

- `mempalace.yaml` — tracked, and already a reachability element.
- `.graphifyignore` — tracked, **not** an element.
- `.memory/**` — 741 tracked files, **not** elements, though `.memory/**` is already listed in `pr-gate.yml`'s `agent_guidance` path filter. The repository classified the store as agent guidance before the reachability check did — the same argument that brought `.github/skills/**` inside the boundary.

So all three are source-controlled; none of Graphify's or the memory store's surfaces are inside the harness boundary. That part is being handled in #4531 / PR #4539.

Grouped under #4533. Related: #4530, #4531, #4536, #4538.

### LINKED ISSUE #4542: The Stop hook blocks on 'pending work' without ever checking for the PR that would clear it, so a delivered branch has no reachable green state
Found live, three times in one session, on branch `ChaosEngine/harness-one-skill-4533` with PR #4539 open over exactly its commits.

## Measured

`_worktree_report` for the current worktree:

```
{'path': 'C:/Users/Mohab/IdeaProjects/SHAFT_ENGINE',
 'branch': 'ChaosEngine/harness-one-skill-4533',
 'uncommitted_files': 0, 'corrupt_files': 0,
 'ahead': 3, 'behind': 0, 'unique_commits': 3,
 'open_pull_requests': None,
 'state': 'pending'}
```

`open_pull_requests: None` is not "no open PR". It is **the lookup was never performed** — `worktree_hygiene.py` only runs it under `--check-pull-requests`, and the Stop hook does not pass it. `run_stop` then maps `state == "pending"` to a block: *"Current branch still carries pending work. Re-read the act-as-mohab Completion section..."*

## Why this is a defect rather than a strict guard

The branch is clean, pushed, and covered by an open pull request carrying `Closes` lines for nine issues. Every preservation and delivery step the block asks for has already been taken. Walk the actions available:

- commit — nothing to commit, `git status` is clean;
- push — `@{u}..HEAD` is empty;
- open a PR — #4539 is open over these exact commits, and the hook does not look;
- merge it — it is a draft mid-RED-GREEN, and merging to silence a hook would be shipping to satisfy a check.

**No reachable state clears the block** short of merging. This is the exact shape recorded in `gotcha.a-check-whose-healthy-end-state-is-unreachable-is-a-check-that-will-be-weakened`: individually defensible guards that leave no legal configuration at the system's intended end state, where *"the cheapest exit an agent will see is deleting the guard"* — which iron law 4 forbids. A long-running session that legitimately holds an open draft PR across iterations hits this on every single turn.

It also fails the rule the owner stated on #4541: the guard **requires** the agent to resolve a condition it never **does** the work to evaluate. One `gh pr list` would settle it.

## Fix

Have the Stop hook perform the pull-request lookup for the current worktree, so a branch whose commits are covered by an open PR resolves to delivered rather than pending. `worktree_hygiene.py` already implements it behind `--check-pull-requests`; this is wiring, not new capability.

Constraints the fix has to respect, or it trades one failure for another:

1. **Fail open, never closed.** No `gh`, no auth, no network, rate-limited, or slow — the hook must not block. `test_worktree_hygiene.py` already covers a failing lookup on the orphaned-branch path (`test_a_failing_pull_request_lookup_does_not_break_the_report`); the same contract applies here.
2. **Respect the Stop timeout.** One bounded call for the current worktree only, not a survey of all ten.
3. **Keep `pending` meaningful.** Commits ahead with a confirmed *zero* open PRs is still a real finding — that is the case the hook exists for. The bug is treating "did not look" identically to "looked and found none". Those must become distinct states.
4. **Draft PRs count as delivered for this purpose.** The work is recoverable by anyone, which is what the check is protecting (#4538). Readiness is the review gate's business, not the Stop hook's.

Point 3 is the general lesson worth carrying beyond this file: an unknown must not be silently collapsed into a negative. `open_pull_requests: None` and `open_pull_requests: 0` mean opposite things about whether an agent should be stopped.

Related: #4541 (require-vs-do registry), #4538 (draft PR as the recoverability mechanism), #4450 (PR-less branch hygiene), #4540.

MohabMohie:
Owner goal, restated as the acceptance bar: **one singular `act-as-mohab` skill that any LLM or agent, on any machine, from any vendor, can load and thereby reach 100% of the harness, with zero orphaned components** — guidance, skills, hooks, triggers, timelines, workflows, and governance rules, all source-controlled.

Grouped per the owner directive in #4533: one branch, one PR.

## Where it landed

**Reachability is closed and pinned.** 123 elements, 0 orphans, 0 broken links, 0 stale named paths, 0 exemption problems. `validate_agent_setup.py --skip-external` runs the walk itself, so the boundary is enforced by the same command CI already runs rather than by a separate opt-in check.

The check derives the boundary from the guidance rather than from a list: a reachable file that tells an agent to run an instrument has already claimed that instrument as harness, so excluding it is a contradiction the repository detects by itself. That is why it could not be satisfied by editing `element_globs` — the failing token originates in the guidance.

The largest single finding was that **`.memory/**` sat entirely outside the boundary** — 741 tracked files, including objects under `.memory/memory/constraints/` that are literally governance rules (`one-branch-one-worktree-one-pr-per-session`, `every-pr-body-must-use-github-closing-keywords`, `memory-and-graphify-are-a-mandatory-session-entry-point`). None was reachable from the entrypoint, and none could ever have been *reported* as an orphan, because no glob covered it. Resolved with coverage for the store's contract files plus a stated `memory_store_note`, rather than by pretending 741 data objects can each carry a markdown link.

**Eleven required-action rules gained a mechanism.** The guard compiled only prohibitions before this branch — never-do-X — and left every always-do-X-first rule as prose, which is the class the literature measures as decaying with context (arXiv 2604.20911) and which long binding policy documents fail to hold at 36.2% strict adherence even at maximum reasoning effort (arXiv 2607.25398). R12–R19 plus the session ledger convert eight of them into gates, and SessionStart carries the constraint titles in rather than asking for them.

## Independent adversarial review

One round, by a separate agent instance in its own worktree at the PR tip, prompted to refute rather than to confirm. Recorded here because nothing else records it — that gap is #4545.

**Verdict: REQUEST CHANGES, 2 confirmed blockers.** Both are fixed in this branch and both were reproduced before being fixed:

1. **The live-state defect had shipped twice, identically.** The first fix named R18. R17 reads `gh` for a reviewed pull request nobody armed, was left live, and turned the same five Stop tests red the moment this PR received its own review. Fixed by isolating *every* Stop rule and pinning the list with `StopRuleIsolationIsCompleteTest`, which fails until a newly added rule is classified. CI structurally cannot catch this shape: a fresh checkout has nothing unpushed and no credentials to ask about reviews.

2. **R13 and R18 blocked delivered work.** This repository squash-merges and GitHub deletes the head branch, so every original commit of a fully delivered branch reports as existing on no remote. R13 therefore refused `git branch -D` — the entrypoint's own cleanup step 2 — on correct work, and named a remedy, `git push -u origin <branch>`, that would have re-created the remote branch for already merged work. Reproduced in a fixture repository. `git cherry` cannot substitute: it compares patch ids, and a squash of two commits matches neither, verified in the same fixture. Recoverability is now judged by content over the files the branch introduced, which is what this repository's own hygiene advisory already prescribed.

Also fixed from the same review:

3. **The shared subprocess budget was claimed and never implemented.** One PreToolUse invocation was measured issuing 7 subprocesses at 4s each — 28s against a 10s hook timeout. Overrunning does not produce a slow decision, it produces *no* decision, because a killed PreToolUse hook fails open and skips every rule in the file. `main` now opens one window per invocation and all 8 subprocess sites draw from it.

4. **`Closes #4529` overclaimed** — see below.

5. **R14 offered a remedy R8 refuses.** "Stash deliberately" is blocked outright here. Not a deadlock, since the other two exits are legal, but it is the near miss `decision.check-every-new-guard-pairwise-against-the-guards-already-shipped` exists for. Removed and pinned by a test that runs each remedy through the neighbouring rules.

What the review attacked and could not break: the `EXPECTED_ELEMENT_COUNT` equality pin (verified by running — 123, honest), reachability absorption (wildcards excluded, tokens must be exact and tracked, prose ignored), all eight new rules bound to a hook, `ModuleNotFoundError` from a foreign cwd, multi-reason Stop starvation, hook latency on the widened `Write|Edit` matcher, and pairwise deadlocks across R12–R19 × R1–R11 — no third deadlock found. Full required suite 290 tests OK at review time, 483 after these fixes.

Minor findings not fixed here, filed instead so the fix is not lost: `--self-test` covers only R1/R9/R10/R11 and exercises none of R12–R19; `ledger_record` is a non-atomic read-modify-write and parallel tool calls can lose an event; ledger files are never reaped; and R17/R18/R19 read `cwd` raw while R10/R11/R14 use `_hook_working_directory`.

## Closing keywords, applied honestly

The review found `Closes #4529` claimed an issue whose stated defect was **already false on `origin/main`** — the `workflows` filter matched `.github/workflows/**` and the timeout job already ran the guard's test, verified directly against main. This PR fixes a real but *different* gap: the guard's own two files were in no filter. Items 2 and 3 of that issue are record corrections, and nothing in this diff touches them.

Applying the same standard to every line rather than only the one that was caught:

**Fully resolved, closed here:**

Closes #4531
Closes #4540
Closes #4542

**Partial by intent, left open with the remainder stated:**

- **#4529** — the filter gap is fixed; the issue's own item 1 was already untrue on main, and items 2–3 are untouched record corrections.
- **#4530 / #4538** — R18 gates the recoverability half: commits existing on no remote at the end of a turn. The five-minute interval itself is not gated, because no hook can observe an interval without a wall clock the agent shares. Unpushed-at-turn-end is the failure the interval exists to prevent, and it is observable exactly when it matters.
- **#4541** — eleven of twelve required-action rules now carry a mechanism, but the registry that records performed/gated/evidenced per rule is not built. Closing it would leave that work with no ticket.

#4532, #4534, #4535 and #4536 are untouched by this branch and carry no keyword; they are regrouped on follow-up branches per #4533.

#4536 and iron laws 2 and 5 will not reach *gated*, and that is reported rather than relabelled: a hook can see a comment exists but never whether it captured the decision, and no hook observes an assertion. **10 of 12 gated or performed, 2 evidenced.**

## Verification

- `py -3 -m unittest` over the 15 agent-guidance modules the gate runs — **483 tests OK**
- `py -3 scripts/agents/guard.py --self-test` — exit 0
- `py -3 scripts/ci/validate_agent_setup.py --skip-external` — exit 0, and it runs the reachability walk
- Both blockers reproduced before fixing; both fixes mutation-proven (unlisted Stop rule caught, phantom Stop rule caught, unwired hook budget caught)
- The delivered-branch helper exercised against a real fixture repository, not only under mocks

https://claude.ai/code/session_017uBrtSmdU6Xvwmkyajqkuy


MohabMohie:
## Independent adversarial review — round 1, findings applied

Separate agent instance, own worktree at the PR tip, prompted to refute. Recorded here because nothing in the harness records that a review happened — that is #4545 defect 3.

**Verdict: REQUEST CHANGES, 2 confirmed blockers.** Both reproduced before being fixed.

| # | Finding | Status |
| --- | --- | --- |
| 1 | R17 repeats the live-state defect R18 just fixed; five Stop tests go red once this PR receives a review | fixed, `a2fffdf` |
| 2 | R13/R18 block delivered squash-merged branches, and the remedy resurrects merged branches | fixed, `6793caf` |
| 3 | The shared subprocess budget was claimed in a comment and never implemented; 7 calls x 4s against a 10s hook | fixed, `002d5f4` |
| 4 | `Closes #4529` overclaims — item 1 was already false on main, items 2-3 untouched | fixed in the PR body |
| 5 | R14 offers "stash deliberately", which R8 refuses outright | fixed, `002d5f4` |
| 6 | Four minor: self-test coverage, ledger atomicity, ledger reaping, cwd asymmetry | filed as #4551, #4552, #4553 |

Two things worth keeping from this round:

**Finding 1 is the same defect twice in one branch.** The first fix named R18 rather than the class R18 belonged to, so R17 walked straight into it. The generalisation now is an equality pin — `StopRuleIsolationIsCompleteTest` asserts the isolated list equals the rules `run_stop` actually calls, in both directions, and fails in the commit that adds a rule. CI structurally cannot catch the underlying shape: a fresh checkout has nothing unpushed and no credentials to ask about reviews, so the environment that runs the suite is exactly the one where the bug is invisible.

**Finding 2 defeats the obvious fix.** `git cherry` compares patch ids, and a squash of two commits matches neither — verified in a fixture repository where both commits came back `+`. Only a content comparison answers the question R13 and R18 are actually asking, which is what this repository own hygiene advisory already prescribed: "diff its tip against main over exactly the files it touched."

**Applied to every closing keyword, not just the one caught.** #4529, #4530, #4538 and #4541 are now stated partials with no keyword; only #4531, #4540 and #4542 are closed. #4541 in particular would have closed with its main deliverable, the registry, unbuilt.

What the review attacked and could not break: the element-count equality pin (verified by running), reachability absorption, all eight new rules bound to a hook, foreign-cwd imports, multi-reason Stop starvation, hook latency on the widened matcher, and pairwise deadlocks across R12-R19 x R1-R11 — no third deadlock found.

**Verification after the fixes:** 483 agent-guidance tests OK, `guard.py --self-test` exit 0, `validate_agent_setup.py --skip-external` exit 0. Both fixes mutation-proven (unlisted Stop rule caught, phantom Stop rule caught, unwired hook budget caught), and the delivered-branch helper exercised against a real fixture repo rather than only under mocks.

**Not independently reviewed:** the three commits that fix these findings. Stated rather than glossed.

PR #4597, opened 2026-08-08
test: inspect rendered Stop messages

### LINKED ISSUE #4563: The interrupts-once invariant reads source literals, so a message assembled at runtime evades it
`_bare_interruption_promises` in `tests/scripts/test_guard_lifecycle.py` asserts that no guard message promises a single interruption. It reads the module's source with `ast`, merges adjacent string literals, lowercases, collapses whitespace, and requires every `interrupts once` to be followed by `per turn`.

It inspects **source string literals only**. It never sees a check function's rendered return value.

## Reproduced

An adversarial reviewer built four messages whose *rendered* output contains the bare promise. The checker reported zero violations for all four:

| technique | rendered value | checker |
|---|---|---|
| f-string split by an interpolation | `...This interrupts  once.` | `[]` |
| `.format()` splitting the phrase | `This interrupts  once.` | `[]` |
| `%`-formatting splitting the phrase | `This interrupts  once.` | `[]` |
| runtime `+` of two module-level constants | `...This interrupts once.` | `[]` |

Not exploited today — the five live `interrupts once per turn` messages in `guard.py` are all plain adjacent literals and are caught correctly. This is a latent gap, and the reason to fix it is that R17's message already carried the phrase split across two source lines once, which is how the previous version of this check was defeated.

Also minor and opposite in direction: `interrupts once, per turn` would false-positive, because the lookahead requires the literal ` per turn` immediately after. Low impact; contributors will match the required phrase.

## The history, which is the actual argument for fixing it properly

This is the **third** time the same check has been the wrong width, each time in a different disguise:

1. First version asserted `assertNotIn("This interrupts once.", source)` — one literal spelling. Defeated by lowercase in R16 and by implicit string concatenation in R17, where a quote and a newline sit inside the phrase so the substring is not in the file's text at all.
2. Second version, the ast-based one, fixed adjacent-literal splitting and casing. Defeated by any message assembled at runtime.
3. Its docstring then claimed it catches a bare promise "in any phrasing" and that "a future rule that adds a third bare-promise variant fails this the same way the first two would have" — false for all four techniques above. That claim has been corrected to state the real limit; the gap itself remains.

Each fix closed the technique that had just been observed and left the class open. That pattern is the finding, more than any individual evasion.

## Proposed fix

**Assert over rendered values, not over source.** Call each `Stop`-dispatched check with inputs that make it return its message, collect the returned strings, and apply the invariant to those. A rendered string cannot be split by f-strings, `.format`, `%`, or concatenation — the technique disappears rather than being enumerated.

Constraints for whoever implements it:

- The checks are already exercised in both directions by `guard.py --self-test`, so the machinery for driving each rule to its firing state exists. Reuse it rather than building a parallel harness.
- Keep it independent of live state. `StopTestsAreIndependentOfLiveStateTest` runs the Stop suite twice with every subprocess refused and demands identical outcomes.
- The invariant must fail when a **new** rule is added whose message carries a bare promise, without that rule being enumerated anywhere. If the implementation needs a list of rules to check, pin that list by equality against `run_stop`'s dispatch, the way `ISOLATED_STOP_RULES` is pinned — otherwise the next rule is invisible to it and this recurs a fourth time.
- Source scanning may remain as a cheap second layer, but it must not be the only one, and its docstring must keep saying what it does not cover.

Found during the third adversarial review of #4554. Related: the wrong-width trap is item 3 in the consolidated recurring-traps record. Not started.

MohabMohie:
## Summary

- render one reporting message from every Stop-dispatched rule using deterministic guard stubs
- pin the renderer set to `run_stop` dispatch so future rules cannot escape coverage
- catch bare promises assembled at runtime as well as split literals
- preserve an observed RED commit before the green implementation commit

## Validation

- `py -3 -m unittest tests.scripts.test_guard_lifecycle.InterruptsOncePromiseIsHonestTest`
- `py -3 -m unittest tests.scripts.test_guard_lifecycle` (189 tests)
- `py -3 scripts/agents/guard.py --self-test`
- `git diff --check`

Closes #4563
Supersedes #4596

### DIFF
--- a/tests/scripts/test_guard_lifecycle.py
+++ b/tests/scripts/test_guard_lifecycle.py
@@ -2674,6 +2674,10 @@ class InterruptsOncePromiseIsHonestTest(unittest.TestCase):
             violations, "a split-literal bare promise must be caught, not missed"
         )
 
+    def test_a_runtime_assembled_bare_promise_is_caught(self):
+        message = f"This interrupts {'once'}."
+        self.assertTrue(_bare_interruption_promises([message]))
+
     def test_the_honest_phrasing_raises_no_violation(self):
         """The check must not flag the wording it is meant to require."""
         synthetic = 'MESSAGE = "This interrupts once per turn: retry proceeds."\n'
--- a/tests/scripts/test_guard_lifecycle.py
+++ b/tests/scripts/test_guard_lifecycle.py
@@ -2590,24 +2590,46 @@ class RunStateStopGateTest(unittest.TestCase):
 
 
 
-def _bare_interruption_promises(source: str) -> list[str]:
-    """Every "interrupts once" in `source` not immediately followed by "per turn".
+_STOP_RULE_RENDERERS = {
+    "check_r16_learning_loop": lambda: guard._with_stubs(
+        {"ledger_events": lambda payload: {"commit"}},
+        lambda: guard.check_r16_learning_loop({"session_id": "s"}),
+    ),
+    "check_r17_unarmed_pull_request": lambda: guard._with_stubs(
+        {"_unarmed_reviewed_pull_request": lambda cwd, payload=None: "1"},
+        lambda: guard.check_r17_unarmed_pull_request({"cwd": "."}),
+    ),
+    "check_r18_unpushed_work": lambda: guard._with_stubs(
+        {
+            "_current_branch": lambda cwd: "feature",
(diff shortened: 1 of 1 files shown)


MohabMohie:
Superseded by a corrected RED-before-GREEN history that moves the deterministic renderer into guard.py; no force-push used.

PR #4811, opened 2026-08-12
test: run mobile recording on a compatible Appium provider

### LINKED ISSUE #4751: Establish honest real-provider acceptance for Appium screen recording
## Incident

The dedicated real-Android recording acceptance in #4740 reached a BrowserStack App Automate Pixel 7 / Android 13 session, then failed on the first public SHAFT recording start in workflow run https://github.com/ShaftHQ/SHAFT_ENGINE/actions/runs/31535484748.

Observed provider response:

org.openqa.selenium.WebDriverException: Command is not supported
Command: startRecordingScreen {options={timeLimit=30}}
Driver: io.appium.java_client.android.AndroidDriver
Session: BrowserStack App Automate, Pixel 7, Android 13

This is not a SHAFT selector, model, lifecycle, or report-guard failure. The pinned Appium Java client exposes CanRecordScreen on AndroidDriver, but the real BrowserStack endpoint rejects that Appium command. BrowserStack documents its own always-on session video, which is a separate provider artifact and does not exercise SHAFT's mobile().recording().start/stop/stopAndSave path.

## Required investigation/outcome

- Confirm whether any BrowserStack Appium version/device configuration supports the Appium screen-recording endpoint on Android and iOS.
- If supported, pin the exact provider configuration and add guarded real-device acceptance.
- If unsupported, preserve the provider exception unchanged and define an honest real-provider negative acceptance for BrowserStack.
- Establish a real Appium provider/lab acceptance that actually exercises successful bounded start, stop, restart, and stopAndSave; do not substitute BrowserStack's session-video REST artifact for the public SHAFT namespace.
- Keep exact testcase/no-skip guards so an unavailable provider cannot create a vacuous GREEN.
- Document the provider boundary without inferring support solely from the Java driver's implemented interface.

Related to #4740 and #4727.

MohabMohie:
## Summary
- preserve the existing BrowserStack fixture as the default
- add an explicit fail-closed compatible-provider opt-in using `executionAddress` and `mobile_app`
- bind configuration precedence, missing-input rejection, and ambient JVM-property isolation
- execute the existing positive recording acceptance on a local Appium Android provider

## Verification
- `mvn -pl shaft-engine -am test "-Dtest=MobileTestConfigurationTest,MobileRecordingActionsTest,MobileNamespaceTest" "-Dallure.automaticallyOpen=false" "-Djacoco.skip=true"` — 35/0/0/0
- exact live `AndroidBasicInteractionsTests#screenRecordingShouldReturnAndSaveBoundedMedia` against Appium 3.6.0 / UiAutomator2 7.6.2 / Android 36 — 1/0/0/0
- `git diff --check` — clean apart from checkout line-ending notices

## Provider boundary
The opt-in is explicit and test-fixture-only. It does not infer support from the provider, does not replace the BrowserStack negative coverage, and does not substitute provider session video for Appium recording-command bytes.

Closes #4751

### DIFF
--- a/shaft-engine/src/test/java/testPackage/appium/MobileTest.java
+++ b/shaft-engine/src/test/java/testPackage/appium/MobileTest.java
@@ -7,7 +7,14 @@ import org.openqa.selenium.Platform;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 
+import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
+
 public abstract class MobileTest {
+    private static final String COMPATIBLE_PROVIDER = "shaft.mobile.compatibleProvider";
+    private static final String EXECUTION_ADDRESS = "executionAddress";
+    private static final String MOBILE_APP = "mobile_app";
     public static final ThreadLocal<SHAFT.GUI.WebDriver> driver = new ThreadLocal<>();
 
     @BeforeMethod(onlyForGroups = {"NativeAndroidDemo"})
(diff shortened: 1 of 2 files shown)


MohabMohie:
Security-review bot capacity was unavailable, so this was not treated as a green review. An independent read-only adversarial review inspected the exact pushed two-file diff, configuration precedence, provider boundary, JVM-property isolation, retained 35/0 aggregate, and exact live-provider 1/0 XML; it reported zero blocking and zero nonblocking findings. CodeQL and Codacy remain required remote gates.

Issue #4860, opened 2026-08-13
Wire managed local AI configuration, transparent provisioning UX, and diagnostics


MohabMohie:
Parent: #4851. Depends on runtime and hardware-tier issues. Research contract: #4852.

## Outcome
A user who explicitly enables managed local AI needs no runtime, model manager, container, service, endpoint, credential, or administrator setup; SHAFT owns setup and recovery.

## Scope
- Disabled-by-default properties following SHAFT precedence: system > project custom > default/classpath, including thread-local behavior.
- Default transparent provisioning after enablement and an advanced explicit provision action.
- Continuously visible runtime/model/license, bytes/percentage, storage class, phase, cancellation, retry, disk/RAM exclusion, cache health and cleanup/update/rollback status.
- IntelliJ, CLI/MCP, and Doctor-facing status through one service contract; no GUI launch or machine-wide installer.

## Invariants
Local enablement grants local-processing consent only, never remote consent. Cancellation leaves no active partial artifact. Configuration never bypasses AiExecutionService policy or deterministic fallback.

## Acceptance
- Clean-host acceptance proves first enablement through ready inference with mocked artifact boundaries and one controlled real matrix.
- Offline, cancellation, contention, antivirus lock, corrupt cache, low resources, upgrade and rollback flows are actionable and idempotent.
- Public behavior receives companion documentation in a separate shafthq.github.io PR.

MohabMohie:
## Executable specification and implementation decision (2026-08-13)

### Objective and proof of done

A clean desktop/headless host sets one managed-local enablement property (or the equivalent SHAFT thread-local setter), thereby granting **local processing consent only**. The first approved `AiExecutionService` request transparently performs reviewed provisioning, launches the SHAFT-owned authenticated loopback runtime, returns schema-constrained inference, and exposes the same immutable lifecycle snapshot to direct Java, Doctor, MCP/CLI, and IntelliJ callers. A mocked clean-host flow must reach `READY` without external runtime/model-manager/container/service/credential/admin setup; normal builds and disabled inspection must perform no network or cache mutation.

### Resolved caller matrix

| Site | Effective cwd/path | Runtime/version/platform | Permissions/trust | Configuration precedence | Input existence |
| --- | --- | --- | --- | --- | --- |
[...]

PR #4883, opened 2026-08-13
feat(ai): configure managed local inference

MohabMohie:
## Summary
- add disabled-safe SHAFT properties for managed local inference and transparent provisioning
- expose thread-isolated overrides with explicit system/project/default precedence
- select the managed-local Pilot provider only when explicitly enabled, granting local consent without changing remote or on-prem consent

## Verification
- 6/6 managed property and initialization tests passed after syncing main
- 6/6 Pilot provider configuration tests passed after syncing main
- independent adversarial review approved this behavior slice with no HIGH/MEDIUM findings

Related to #4860


MohabMohie:
## Independent review round 3 — blocking

Exact reviewed state:
- base HEAD: `b51ed20ec69470afa8cfb078e90b624960975dba`
- six modified managed-local-AI files
- diff SHA-256: `DF248CF590A50B983CD4FF44F5404326F08FCFF82AEF388B2B9D52ADB97B46CB`
- affected low-memory/no-fork suite: 22 tests, 0 failures/errors/skips

Two blocking findings remain:

1. `ManagedLocalAiProcess.launch` gives each of three attempts the same full remaining duration and adds termination waits. A single request can therefore exceed its end-to-end deadline by roughly three times plus cleanup.
2. If transparent provisioning exceeds the request deadline, timed `get` returns `TIMEOUT` but the `ManagedLocalAiOperation` is discarded without cancellation. Its worker can continue artifact/cache mutation after the caller returns.

Review rounds 1 and 2 found and drove fixes for seven earlier blockers. ChaosEngine's three-round ceiling is now reached. Explicit owner instruction on this PR is required before a fourth repair/review round.

Separate surfaced tooling defect: #4908 tracks the host safety layer rejecting verified exact-target cleanup. The generated root `src/` artifact remains; no deletion occurred.

MohabMohie:
Fourth independent review remains BLOCKING (exact six-file diff SHA-256 `C67659E00B468D0B0B731B76D5CDA9BEB2B9FA88F9366B90EF4A0B203638468D`):

1. `ManagedLocalAiProcess.requestIdentity` uses independent fixed 2-second HTTP request and response-body timeouts, so an attempt with less time remaining can exceed the shared launch deadline.
2. `ManagedLocalAiProcess.terminate` reuses the supplied timeout across parent wait, descendant wait, and forced-parent wait, allowing cleanup to consume roughly 3x the remaining budget.
3. `ManagedLocalAiProvider.execute` captures remaining time before synchronous `service.provision()` preflight and reuses that stale duration for `completion().get(...)`, excluding preflight from the request deadline.

All three are confirmed blocking `wrong-width-check` defects. The provisioning cancellation signal itself is now correctly wired and observed. PR is not approved for merge.

Separately, cleanup of the generated root `src/` remains blocked by the host execution layer and is tracked visibly in #4908.

MohabMohie:
Next independent review remains BLOCKING (exact six-file diff SHA-256 `C270A31CA251781E486C462B5D531F75DDD99CC43E4FB853DF7F71DD09355D69`). The previous three deadline blockers are confirmed fixed. Two new blockers remain:

1. `ManagedLocalAiProvider.ServiceLifecycle.execute` is a synchronized method and starts the request deadline only after monitor acquisition. A short-timeout request queued behind a long provisioning/inference request can wait without its timeout running.
2. When replacing a stale or alias-mismatched session, `closeSession()` uses the prior session's stored launch/shutdown timeout before checking the current request's remaining deadline. A short request can therefore block for the previous request's much longer budget.

No current test covers concurrent `execute` queue time or current-vs-prior shutdown timeout during session replacement. PR remains unapproved pending RED/GREEN repair and another independent pass.

The unrelated generated-root cleanup failure remains visible in #4908; `src/` has not been deleted.

MohabMohie:
Next independent review remains BLOCKING (exact diff SHA-256 `1914B25F31A146993571E5A90D2BB10626EC7C17C258284F1BCDD080E50CCD1B`). Prior queue-time and prior-session-budget findings are fixed. Three lifecycle blockers remain:

1. Interrupted execution restores the interrupt flag before cleanup. `terminate` can immediately abort its waits, skip force-kill, and orphan the already-detached live process.
2. Expired cleanup detaches the session and `close(Duration.ZERO)` can throw while force-kill is asynchronous, masking the intended `TIMEOUT` as public `PROVIDER_UNAVAILABLE`.
3. The JVM shutdown hook waits indefinitely on `executionLock.lock()` and cannot cancel/interrupt an in-flight provisioning or inference operation, so shutdown can hang.

The current tests do not cover pre-interrupted cleanup, asynchronous force-kill with zero budget, or shutdown-hook lock contention. PR remains unapproved pending TDD repairs and another adversarial pass.

Generated-root cleanup remains separately blocked and visible in #4908; `src/` is still present.

MohabMohie:
Next independent review remains BLOCKING (exact diff SHA-256 `8D5570A00AFF568B55A6BE3F5B682A81377999F12C397982F2E632AFD4C2DFDD`). The previous interrupt/suppression/active-cancellation findings are fixed. Three shutdown/ownership blockers remain:

1. If shutdown's two-second lock attempt times out, it returns without issuing termination to an existing live session.
2. Shutdown sets no terminal state, so a racing request can acquire the lock afterward and launch a new child after the sole hook cleanup completes.
3. Cleanup restores ownership only when the parent remains alive; an exited parent with an asynchronously surviving descendant loses every lifecycle owner.

Current tests cover cooperative shutdown and synchronously dying descendants, not these paths. PR remains unapproved pending TDD repair and another independent pass.

Generated-root cleanup remains separately blocked and visible in #4908; `src/` remains present.

MohabMohie:
Next independent review remains BLOCKING (exact diff SHA-256 `D1D0F60E00B80F98A7D148C24209C1D92C129706B7661A78A366B7DAF48B9E10`). Three process-tree ownership gaps remain:

1. Descendant handles are first captured during cleanup. If the parent exits and the OS reparents its child before first cleanup, the session cannot rediscover or own that child. The current exited-parent fake is not realistic because it exposes the child through the exited parent.
2. Request cleanup and shutdown fallback can call `Session.close` concurrently. The retained-descendant snapshot/clear/repopulate transaction is not atomic, so one closer can erase a survivor captured by the other.
3. Launch-failure cleanup uses a temporary retained set. A surviving failed-attempt process tree is discarded before retry/final failure, knowingly orphaning the handles.

PR remains unapproved. I am replacing late discovery with live tree ownership, serializing session termination, and retaining failed-attempt trees until confirmed dead under focused RED coverage.

Generated-root cleanup remains separately blocked and visible in #4908; `src/` remains present.

MohabMohie:
Blocking review remains open; implementation is not approved yet.

1. Shutdown registration race: `execute` checks terminal state before publishing `activeExecutionThread`, so shutdown can miss a request that then continues provisioning/launch.
2. Launch deadline reset: the relative timeout is captured before entering synchronized launch, but launch starts a new deadline after queueing/reaper inspection; queue time can exceed the public `AiProvider` timeout.
3. Security/privacy — port theft: the current reserve/release port flow lets a loopback squatter receive the bearer identity probe, spoof the predictable alias, and then receive approved evidence. This breaks the invariant that evidence is sent only to the verified SHAFT-launched runtime.

Chosen repair: atomically publish execution ownership before the post-lock terminal check; carry one absolute deadline through a deadline-aware launch lock; launch the pinned llama.cpp with `--port 0` and discover the child-owned bound port from its sanitized startup stream before any authenticated request. Upstream b10400 supports `bind_to_any_port` for port 0 and logs its resolved listening address.

RED proof will cover the shutdown race, launch-lock queue time, and a real loopback squatter that must receive no bearer or evidence. Normal loopback launch/inference remains the positive control.

The unrelated filesystem deletion failure is still explicitly tracked in #4908: the host execution layer blocked deletion of the untracked root `src/`; it remains visible in `git status` and will not be buried or described as cleaned up.

MohabMohie:
Superseded by clean draft PR #4921, replayed from current main. All uncommitted lifecycle work is preserved on remote recovery commits f1e845b34812c0d5cdc1eeaa5af06382a0198b32 and 6939f69cf82d9400a66f5bfcdb76c47762f4285a; review blockers will be fixed there before replay.

PR #4998, opened 2026-08-15
Tolerate element timeouts that have no root cause

Mochxd:
﻿## Summary
- Element identification logged timeoutException.getCause().getMessage() with no null check
- Selenium TimeoutException often has a null cause, so the real element-not-found path crashed with an NPE instead of returning the missing-element payload
- Guard the cause before reading its message; fall back to the timeout message

## Tests
- [x] ElementsHelperCoverageUnitTest (22/22), including waitForElementPresenceShouldTolerateTimeoutWithoutCause
[2026-08-15_22-54-01-135_AllureReport.html](https://github.com/user-attachments/files/31107527/2026-08-15_22-54-01-135_AllureReport.html)

### DIFF
--- a/shaft-engine/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
+++ b/shaft-engine/src/main/java/com/shaft/gui/element/internal/ElementActionsHelper.java
@@ -494,8 +494,11 @@ public class ElementActionsHelper {
             if (resolution != null) {
                 return recoveredElementInformation(driver, elementLocator, resolution);
             }
-            var causeMessage = timeoutException.getCause().getMessage();
-            causeMessage = !causeMessage.isBlank() && causeMessage.contains("\n") ? timeoutException.getMessage() + " || " + causeMessage.substring(0, causeMessage.indexOf("\n")) : timeoutException.getMessage();
+            var cause = timeoutException.getCause();
+            var causeMessage = cause == null ? null : cause.getMessage();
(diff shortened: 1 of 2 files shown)


Mochxd:
@MohabMohie small heads up from this file, nothing blocking.

The same getCause().getMessage() pattern shows up in two other spots and I left them out to keep this PR focused:

ElementActionsHelper.scrollToFindElement - its wait lambda returns false instead of throwing, so FluentWait never records a lastException and the cause is always null on timeout. That line would NPE every time it's reached. I couldn't find a production caller though, only a unit test, so it's latent for now.

WebDriverListener.beforeFindElement (line 60) - same pattern, and no contains("\n") guard either, so substring(0, -1) is possible too. It rethrows right after, so a crash there would hide the original timeout.

While checking whether that second one was reachable I hit something I'm unsure about. The only place the listener gets attached is SHAFT.java:307, and it's skipped when the driver is an AppiumDriver or a RemoteWebDriver. Since ChromeDriver, FirefoxDriver and EdgeDriver all extend RemoteWebDriver, that check seems to catch local browsers as well, so the branch commented "local execution" never runs as far as I can tell. Was that check meant to target remote sessions only? I didn't touch it, since flipping it would turn the whole listener on.

I am going to fix the two null cause spots here or in a follow up, whichever you prefer..

MohabMohie (approved):

PR #5112, opened 2026-08-18
fix(chaos-engine): accept healthy Memory v4 and exclude-first MemPalace YAML

### LINKED ISSUE #5107: Portable installer refuses a healthy adopter Memory v4 store and MemPalace YAML
## Problem Statement

A real official install into iTestFlow/iTestFlow-Agent on unmodified `main` failed before any payload was published:

```
invalid Memory configuration
```

`hosts.py` `validate_memory_config` requires Memory v5 exact keys. The adopter already had a healthy Memory **v4** store (extra `git` / `saveContextPacks`, `version: 4`) and a valid custom `mempalace.yaml` with exclude-patterns first and unindented lists. `validate_mempalace_config` also rejects that YAML shape.

After reshaping only `.memory/config.json` to v5, `memory status/check` still failed against the project-local **v4** `config.schema.json` (required `git` and `saveContextPacks`). Mixed v5 config + v5 `config.schema.json` + remaining v4 object schemas/objects then passed `memory check`.

`.gitignore` also ignored all of `.claude` and `.codex`; the installer un-ignore block is required for adapters to be source-controlled.

## Proposed Solution

Make official install succeed on an existing healthy Memory v4 + custom MemPalace taxonomy without a silent object conversion:

- accept or migrate `.memory/config.json` v4 to v5 without deleting objects
- keep or upgrade project-local schemas so `memory check` matches the config version
- accept existing MemPalace YAML that already has a wing, rooms, and exclude patterns
- keep the gitignore un-ignore block (already present)

## Alternatives Considered

- Archive v4 objects and start an empty v5 store: preserves files but drops live Memory. Rejected.
- Remap v4 types into v5 enums: silent conversion of user knowledge. Rejected.
- Leave the adopter to hand-edit config/schema/YAML before install: what we had to do; this ticket exists so the next adopter does not.

## Use Case & Impact

Any adopter that already used `@aictx/memory` 0.1.55 / schema v4, or a custom `mempalace.yaml`, cannot run the documented one-liner until this is fixed.

## Acceptance criteria

- [ ] Official bootstrap on a fixture with iTestFlow-like v4 `config.json` + v4 object store + unindented `mempalace.yaml` reaches doctor healthy without deleting objects
- [ ] `memory status --json` and `memory check --json` succeed after install
- [ ] Focused installer/hosts tests cover the v4 config and YAML shapes
- [ ] No silent rewrite of Memory object types

## Proof plan

- Add a fixture matching the iTestFlow v4 config and YAML
- Prove RED on current `main`
- Prove GREEN after the installer change
- Re-run the iTestFlow official install without the manual config/schema/YAML preflight

## Duplicate search

- Query: portable installer Memory v4 config mempalace.yaml
- Open: none
- Closed: #4965 (old ShaftHQ/iTestFlow-Agent 404; not this validator collision), #5105 (brand/research payload, separate)

Evidence: iTestFlow/iTestFlow-Agent#188 / PR iTestFlow/iTestFlow-Agent#189

MohabMohie:
Closes #5107

## Summary

Official host install now keeps a healthy adopter Memory v4 store and a custom MemPalace YAML in place.

`validate_memory_config` accepts v5 exact keys or healthy v4 exact keys (`version`, `project`, `memory`, `git` with `saveContextPacks` / `trackContextPacks`). `validate_mempalace_config` accepts `wing` + `rooms` + `exclude_patterns` in any order and 0-or-more-space list items. Incomplete v4 and empty YAML still fail before mutation. Fresh installs still seed Memory v5. Existing object files and project-local v4 schemas are not rewritten.

## Checks

- RED: `py -3 -m unittest tests.scripts.test_chaos_engine_hosts.ChaosEngineHostsTest.test_healthy_adopter_v4_memory_and_unindented_mempalace_installs` raised `ValueError: invalid Memory configuration` on current `hosts.py` before the validator change
- GREEN: same test plus `test_invalid_retrieval_configs_fail_before_mutation` and `test_complete_host_harness_installs_inventory_roles_hooks_and_plugin` (3 tests, OK)
- Live SHAFT checkout `.memory/config.json` (v4) and exclude-first `mempalace.yaml` now pass `retrieval_configs_healthy`

## Continuation

Head: 57575a3cb614cef74add599511f38e324b0d717b
State: implementation committed and pushed; independent review then arm auto-merge
Blockers: none
Next action: independent adversarial review of this exact diff, then `gh pr merge --auto --merge` and watch checks to confirmed merge

### DIFF
--- a/chaos-engine/hosts.py
+++ b/chaos-engine/hosts.py
@@ -119,30 +119,60 @@ def project_identity_name(project: Path) -> str:
     return project.name
 
 
+def _memory_project_valid(project_config: object) -> bool:
+    return (
+        isinstance(project_config, dict)
+        and set(project_config) == {"id", "name"}
+        and isinstance(project_config.get("id"), str)
+        and re.fullmatch(r"project\.[a-z0-9][a-z0-9-]*", project_config["id"]) is not None
+        and isinstance(project_config.get("name"), str)
+        and bool(project_config["name"].strip())
+    )
+
+
+def _memory_options_valid(memory_options: object, required: set[str]) -> bool:
+    budget = memory_options.get("defaultTokenBudget") if isinstance(memory_options, dict) else None
+    return (
+        isinstance(memory_options, dict)
+        and set(memory_options) == required
+        and isinstance(memory_options.get("autoIndex"), bool)
+        and isinstance(budget, int)
+        and not isinstance(budget, bool)
+        and 501 <= budget <= 50000
+        and (
+            "saveContextPacks" not in required
+            or isinstance(memory_options.get("saveContextPacks"), bool)
+        )
+    )
+
+
 def validate_memory_config(content: bytes) -> None:
     try:
         config = json.loads(content)
     except (UnicodeDecodeError, json.JSONDecodeError) as error:
         raise ValueError("invalid Memory configuration") from error
-    project_config = config.get("project") if isinstance(config, dict) else None
-    memory_options = config.get("memory") if isinstance(config, dict) else None
-    if (
-        not isinstance(config, dict)
-        or set(config) != {"version", "project", "memory"}
-        or config.get("version") != 5
-        or not isinstance(project_config, dict)
-        or set(project_config) != {"id", "name"}
-        or not isinstance(project_config.get("id"), str)
-        or re.fullmatch(r"project\.[a-z0-9][a-z0-9-]*", project_config["id"]) is None
-        or not isinstance(project_config.get("name"), str)
-        or not project_config["name"].strip()
-        or not isinstance(memory_options, dict)
-        or set(memory_options) != {"autoIndex", "defaultTokenBudget"}
-        or not isinstance(memory_options.get("autoIndex"), bool)
-        or not isinstance(memory_options.get("defaultTokenBudget"), int)
-        or isinstance(memory_options.get("defaultTokenBudget"), bool)
-        or not 501 <= memory_options["defaultTokenBudget"] <= 50000
-    ):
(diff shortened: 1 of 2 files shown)


MohabMohie (commented):
Adversarial review of `57575a3cb614cef74add599511f38e324b0d717b` (this exact PR head). Refute pass against the recorded #5107 spec.

Spec first
- Root owner is the two keep-existing validators. Diff does not touch `install.py`/`bootstrap.py` `source_files()`, so it does not re-ship origin brand/research.
- Fresh v5 exact-key contract remains; only a healthy v4 exact-key set is added. Incomplete v4 (`memory: {}`, `git: {}`) still fails before receipt.
- MemPalace now accepts any order and 0+ space list items, which is the iTestFlow shape. Empty `rooms:` / `exclude_patterns:` still fail.
- Objects and existing schemas stay unmanaged/kept. Test asserts pre/post bytes, including all five schema files and `type: architecture`.
- Second `install()` stays `installed` and does not rewrite the v4 payload.

Quality / gaps (non-blocking)
- Copilot review request was 403 from this integration; this comment is the recorded independent pass over the GitHub diff.
- Focused test does not run live `memory status --json` / `memory check --json`. That matches the recorded proof map (approach A: consistent v4 store, no schema overlay). Residual: if `@aictx/memory@0.2.1` later hard-rejects `version==4`, doctor runtime would still fail after these validators pass. That is the rejected approach-B risk, not a hole in this diff.
- Room `description:` and exclude `-` matches are no longer `$`-anchored. Real YAML comments/extra tokens can now pass. Recorded negatives still fail.

No blocking finding. Safe to arm.

Issue #5141, opened 2026-08-18
Tracking: Soup follow-ups and owner-gated train


MohabMohie:
## Summary

Owner asked for remaining Soup work plus owner-gated Batch 4, under the same orchestrator rules as #5124: this host session orchestrates; one implementer per batch; the writer stops when the PR exists; Learning Loop at the end.

#5124 is closed. This tracker owns only the leftover follow-ups and Batch 4.

## Orchestration

- The host session orchestrates.

- One implementer per batch.

- The writer stops when the PR exists.

- Independent review on the exact PR SHA before merge.

- Merge with `--merge` (squash forbidden).

- Close that writer after its PR exists.

## Batches (sequential)

1. Checker follow-ups: #5132 + #5133 via PR #5148 (`6e69f6c822b8195fb5af5381522f9be4c7ccfd12`).

2. Memory index + Soup Learning Loop objects: #5136 + #5142 via PR #5153 (`123c4395261851beb0bfdd7c58b8fc49011b03ca`).

3. Owner-gated train / GGUF / new Ollama name + one Aider Java flow: #5143 + #5144.

## Tracking

- [x] #5132 Pin `source_endpoint` negative test

- [x] #5133 Accept quoted OpenAPI `operationId`s

- [x] #5136 Native Memory project-scoped `task` field

- [x] #5142 Commit Soup trainer decision and Python 3.12 uv gotcha

- [ ] #5143 Owner-gated `soup train` + GGUF + new Ollama name

- [ ] #5144 One Aider Java flow against the new name + harvest row

## Out of scope

- `soup serve`, `soup ui`, `soup loop watch`, `soup mcp serve`.

- Vendoring Soup.

- Changing managed llama.cpp / AiExecutionService (#4851/#4852).

- Overwriting the existing workstation Ollama model `qwen2.5-coder:7b`.

- Committing GGUF, weights, `target/`, or `D:\AI` artifacts.

- Auto-train from harvest.

Follow-up from #5153 review: #5156 (`validate_agent_setup` must fail project-scoped `task`).

### User Story 1 - Deliver the stated acceptance (Priority: P1)

**Independent Test**: Prove the acceptance criteria and proof notes already on this issue.

**Acceptance Scenarios**:

1. **Given** the current codebase state named in this issue, **When** the proposed fix lands, **Then** the Acceptance section below (or Success Criteria) holds.

## Edge Cases

[...]

PR #5148, opened 2026-08-18
test(tools): pin source_endpoint and accept quoted operationIds

### LINKED ISSUE #5132: Pin a negative test for soup-delegate-poc source_endpoint
Follow-up from PR #5129 review (F1). Not a child of #5125 after merge; related to #5124.

## Problem Statement

`tests/scripts/test_soup_delegate_poc.py` deletes `tool` and `messages` but never `source_endpoint`. Dropping `source_endpoint` from `REQUIRED_ROW_KEYS` still leaves all five tests green.

## Proposed Solution

Add one negative test that deletes `source_endpoint` and asserts a blocker. Keep the fixture valid.

## Alternatives Considered

Leave it: the checker already rejects the key, but the suite cannot see a regression.

## Use Case & Impact

Anyone changing the row contract. Low frequency, high value as a sibling-left pin.

## Acceptance

`py -3 -m unittest tests.scripts.test_soup_delegate_poc` fails if `source_endpoint` stops being required.

### LINKED ISSUE #5133: Accept quoted OpenAPI operationIds in soup-delegate-poc validate_spec
Follow-up from PR #5129 review (F2). Related to #5124.

## Problem Statement

`validate_spec` uses `operationId:\s*(\S+)` so `operationId: "run_focused_test"` (valid YAML) is treated as missing.

## Proposed Solution

Strip optional quotes from the captured id, or parse with a small YAML subset. Keep stdlib-only. Do not add PyYAML unless already required.

## Alternatives Considered

Require unquoted ids forever: current fixture is unquoted, but OpenAPI commonly quotes strings.

## Use Case & Impact

Future harvest/synth specs. Current shipped spec still passes.

## Acceptance

Quoted and unquoted operationIds both satisfy `validate_spec`. Existing unquoted fixture still passes.

MohabMohie:
## Summary

Pins a `source_endpoint` missing-key regression and accepts quoted YAML `operationId`s in `validate_spec`. Existing unquoted `mechanical-tools.yaml` still passes. Fixture stays valid.

Closes #5132
Closes #5133

Related to #5141.

## Checks

- RED: `py -3 -m unittest tests.scripts.test_soup_delegate_poc` — Ran 11 tests, 1 failure. `test_spec_accepts_quoted_and_unquoted_operation_ids` reported missing quoted `read_file` and `run_focused_test`. `test_row_missing_source_endpoint_fails` already passed because production already required the key.
- GREEN: same command after stripping optional quotes from captured ids — Ran 11 tests in 0.015s, OK.

## Continuation

Head: 5860995f30668b26786aee2ba8ed1b5ab713537c
State: Batch 1 implementation complete. Writer stops here for independent review.
Blockers: none
Next action: Independent adversarial review on this SHA, then merge with `--merge`. Do not start batch 2/3.

### DIFF
--- a/tests/scripts/test_soup_delegate_poc.py
+++ b/tests/scripts/test_soup_delegate_poc.py
@@ -71,6 +71,12 @@ class SoupDelegatePocRowTest(unittest.TestCase):
         blockers = MODULE.validate_row(row)
         self.assertTrue(any("messages" in item.lower() for item in blockers))
 
+    def test_row_missing_source_endpoint_fails(self):
+        row = valid_row()
+        del row["source_endpoint"]
+        blockers = MODULE.validate_row(row)
+        self.assertTrue(any("source_endpoint" in item.lower() for item in blockers))
+
     def test_assistant_tool_name_mismatch_fails(self):
         row = valid_row("read_file")
         row["messages"][1]["tool_calls"][0]["function"]["name"] = "replace_file"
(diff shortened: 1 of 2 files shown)


MohabMohie (commented):
Independent adversarial review on `5860995f30668b26786aee2ba8ed1b5ab713537c`. Zero blocking findings.

Spec check:
- #5132: `test_row_missing_source_endpoint_fails` deletes `source_endpoint` and asserts a blocker. Mutating `REQUIRED_ROW_KEYS` to `("messages", "tool")` yields `blockers=[]`, so the suite fails if the key is dropped. Restoring the tuple returns `['missing source_endpoint']`.
- #5133: `validate_spec` strips optional quotes from captured ids. Mixed quoted/unquoted tempfile fixture returns `[]`. Shipped unquoted `mechanical-tools.yaml` still returns `[]`. Without the strip, `read_file` and `run_focused_test` are missing. Stdlib only; no PyYAML in the two files.
- Files: `tests/scripts/test_soup_delegate_poc.py`, `tools/soup-delegate-poc/soup_delegate_poc.py` only.
- Closes #5132 and #5133. Does not close tracker #5141.
- Tests assert. Observed `py -3 -m unittest tests.scripts.test_soup_delegate_poc` at this SHA: Ran 11 tests in 0.014s, OK.

No F<n> blocks. Loop can stop.

Issue #5240, opened 2026-08-19
[Bug]: Sibling private escapeJson copies still omit RFC 8259 control-character escapes


MohabMohie:
Related to #5213 and #5209. Duplicate search on 2026-08-19 found no open ticket for these sibling serializers; #5213 covers `FailureTraceReporter.escapeJson` only.

## ✅ Pre-submission Checklist

- [x] Searched existing issues. #5213 / #5209 own the timeline/HTML `JSON.parse` failure via `FailureTraceReporter.escapeJson`. This ticket is the remaining copies.
- [x] Observed on `main` at `4a09f00435` while implementing #5213.
- [x] Not a docs-only request.

## 🐛 Describe the Bug

`FailureTraceReporter.escapeJson` is being fixed to emit `\u00xx` for U+0000–U+001F. Four other private `escapeJson` methods still use the old `\\` `"` `\n` `\r` `\t` chain:

| Class | File:line |
|---|---|
| `TraceEventRecorder` | `shaft-engine/src/main/java/com/shaft/tools/io/internal/TraceEventRecorder.java:724` |
| `BrowserObservabilityRecorder` | `shaft-engine/src/main/java/com/shaft/tools/io/internal/BrowserObservabilityRecorder.java:805` |
| `FailureDiagnosticsReporter` | `shaft-engine/src/main/java/com/shaft/tools/io/internal/FailureDiagnosticsReporter.java:413` |
| `FailureBriefReporter` | `shaft-engine/src/main/java/com/shaft/tools/io/internal/FailureBriefReporter.java:411` |

`TraceEventRecorder` and `BrowserObservabilityRecorder` JSON is spliced into the same trace payload via `rawObject` / `rawArray`. An ESC in an action message, console line, or network field would still make `JSON.parse` throw after the timeline fix. Diagnostics/brief JSON has the same RFC 8259 hole in sibling artifacts.

## 🔁 Steps to Reproduce

1. Put `\u001B` in an action result message or a browser console payload that `TraceEventRecorder` / `BrowserObservabilityRecorder` serializes.
2. Render a failure trace.
3. Strict-parse `shaft-trace.json` or the HTML `#trace-data` payload.

## ✅ Expected Behavior

Every SHAFT-owned JSON serializer that feeds the trace viewer or diagnostics JSON emits RFC 8259 escapes for U+0000–U+001F.

## ❌ Actual Behavior

Those four copies leave raw control bytes in strings.

[...]

Issue #5214, opened 2026-08-19
Native Memory index fails closed on blank events.jsonl lines


MohabMohie:
## Problem

`shaft-memory` `search_memory` failed closed during #5207 with:

```
MemoryIndexUnavailable / MemoryInvalidJsonl: Blank JSONL lines are not allowed.
path: .memory/events.jsonl:1176
```

Line 1176 is a blank line between two otherwise valid `memory.updated` / `memory.created` events (timestamps `2026-08-18T20:21:32+03:00` and `2026-08-18T14:40:29+03:00`). Retrieval and remember are unavailable until the index can load.

## Not a duplicate of

- #4137 — GitHub `CONFLICTING` on `merge=union` for this file
- #5136 — project-scoped gotcha `task` field (closed)

## Suggested repair

- Reader: skip blank lines instead of fail-closed, or
- Writer/merge: never emit a blank line into `events.jsonl`

Ordinary tasks should keep treating the store as degraded until this is fixed. Do not rewrite history.

Related to #5207.

Issue #5212, opened 2026-08-19
Tracking: open issues except #4307 (2026-08-19 orchestration)


MohabMohie:
Owner asked this session to work every open SHAFT issue except #4307, keep grouping to the fewest PRs, and keep working until in-scope delivery.

Main session orchestrated only. Merge used `--merge`. This tracker is complete.

## Landed (this session)

- [x] #5207 Local E2E HTML fixtures — PR #5215 `919c6c90e9`
- [x] #5221 IntelliJ plugin release timeout
- [x] #5208 Allure one step per action — PR #5219
- [x] #5220 Allure action-step sweep leftovers
- [x] #5204 Nightly Local E2E tracker
- [x] #5224 / #5225 install URL + STANDALONE.md — PR #5228 `3f5956ca82`
- [x] #5229 source-tree marker — PR #5230 `ed6951c751`
- [x] #5227 Marketplace listing lag — PR #5235 `e12ada50bb`
- [x] #5209 / #5213 trace ANSI JSON.parse — PR #5241 `196b221810`
- [x] #5216 #5217 #5218 Local E2E fixture leftovers — PR #5244 `9671291303`
- [x] #5210 #5246 #5248 #5249 #5232 #5233 #5234 #5253 ChaosEngine harness — PR #5252 `7d5381004e`
- [x] #5240 #5242 #5243 #5239 #5211 #5214 #5245 #5260 product/tooling — PR #5259 `e514556e9e`
- [x] #5247 #5261 IntelliJ Assistant + Agentic Tools — PR #5251 `cd433a4f71` (companion docs shafthq.github.io#998)
- [x] #5151 IntelliJ screenshots — PR #5262 `a997b08fc5` (companion docs shafthq.github.io#1000)
- [x] #5263 unselected agent null CLI — PR #5264 `c6edff1f26`

## Closed without a product PR

- [x] #5144 not planned
- [x] #5141 completed
- [x] #5223 after children
- [x] #5231 duplicate of #5229

## Out of scope

- #4307 (owner exclude)

Delivery loop scheduler `01a01bb0714a` is stopped. Remaining open issue is only the excluded tracker #4307.

MohabMohie:
Dispatching first writer for #5207 (nightly Local E2E root cause). Serial. Next after that PR merges: #5208, #5209, #5211, #5210. Blocked remain #5144/#5141/#5151. Excluded: #4307.

MohabMohie:
[...]

PR #5241, opened 2026-08-19
fix: escape RFC 8259 control chars in trace JSON

### LINKED ISSUE #5209: [Bug]: Offline trace report HTML shows no archive data and ignores clicks
Related to #5204. Child of the nightly Local E2E tracker. Do not close #5204 from this ticket.

Prior closed viewer work this ticket must not re-litigate: #4695, #4787, #4789, #4802, #4804, #4817, #4824, #5131. Those shipped the offline viewer. This ticket is a **solidification / regression** pass because the shipped HTML does not actually show collected data when a human opens it.

## ✅ Pre-submission Checklist

- [x] I have searched existing issues. The mature-viewer work (#5131 and parents) is closed. No open ticket says the generated HTML is empty and its buttons do not respond.
- [x] Observed while investigating #5204: the trace archive contains data; the HTML viewer shows none of it; buttons are nonresponsive.
- [x] This is not a missing-archive problem. The archive has data.

## 🐛 Describe the Bug

Opening the SHAFT trace report HTML from the #5204 investigation shows a shell with no visible collected data. Buttons do not respond. The matching archive/artifact does contain data.

That breaks the product promise of the offline viewer: one self-contained HTML a human can click through (Allure attachment, MCP/IntelliJ open, or file open) and see the session that was recorded.

`TraceViewerBrowserAcceptanceTest` and related archive tests exist, but they did not catch "page loads, controls dead, data invisible." This ticket is a full sweep and solidification round: load, bind data, click every control, assert the corresponding panel shows the archive's events.

## 🔁 Steps to Reproduce

1. Produce or download a SHAFT trace archive from a real GUI test (the #5204 Allure/trace artifacts, or a local run of a browser test with tracing on).
2. Open the generated offline trace report HTML in a browser (the same file Allure/IntelliJ/MCP would open).
3. Observe the empty canvas / empty panels.
4. Click the viewer's tabs, filmstrip frames, action list, and any header buttons.
5. See error: clicks do nothing; none of the archive's collected data is rendered.

## ✅ Expected Behavior

- The HTML loads the archive's data on first paint (or after a documented explicit load that actually runs).
- Every interactive control responds to a real click/keyboard activation.
- Each panel shows the data that exists in the archive: actions, screenshots/filmstrip, network, console, artifacts/integrity, omissions.
- If a slice is absent from the archive, the panel shows an explicit empty/omitted state, not a dead control.
- Keyboard and pointer both work. The page is usable without a live server if the current contract is self-contained HTML.

## ❌ Actual Behavior

The HTML shell appears. Archive data is not shown. Buttons are nonresponsive. A human cannot use the report to inspect the #5204 failure. Existing unit/acceptance tests still pass.

## 💻 Minimal Reproducible Code

No new product API. Repro is:

1. Run any traced GUI test (or reuse the #5204 artifact).
2. Open the generated viewer HTML.
3. Click each control; inspect DOM vs `shaft-trace.json` / archive entries.

Existing tests to extend, not delete:

- `shaft-engine/src/test/java/com/shaft/tools/io/internal/TraceViewerBrowserAcceptanceTest.java`
- `TraceArtifactManifestTest`, `TraceArchiveWriterTest`, IntelliJ `ShowTraceViewerActionTest`

## 🌍 Environment

| Field | Value |
|---|---|
| SHAFT_ENGINE version | `main` as used by #5204 |
| Last working SHAFT_ENGINE version | Unknown; viewer features landed via #4817 / #5131 and now regress in the real HTML |
| Target platform | Offline HTML viewer (Allure attachment + IntelliJ/MCP open) |
| Execution environment | Local file open and CI-generated artifact |

## 📋 Console Logs

Owner observation from the #5204 Allure/trace artifact: archive has data; HTML shows none; buttons dead.

Related implementation surfaces (sweep these; do not assume one file):

- Offline viewer HTML/JS bundled with the trace archive
- `FailureTraceReporter.convergeTraceArchive`
- `PlaywrightTraceArchiveLoader`
- `ShowTraceViewerAction` (opens the file; if it opens a stale empty copy, that is in scope)
- `TraceService` MCP extraction path

### User Story 1 - Data from the archive is visible (Priority: P1)

**Independent Test**: Build a fixture archive with known actions, one screenshot, one network row, one console line. Open the generated HTML in headless Chrome. Assert those strings/nodes are in the rendered DOM (not only in the zip).

**Acceptance Scenarios**:

1. **Given** an archive whose `shaft-trace.json` lists N actions, **When** the HTML is opened, **Then** the action list renders N identifiable rows.
2. **Given** an archive with a screenshot/filmstrip frame, **When** the HTML is opened, **Then** that image is visible (not a blank token-colored box with no source).
3. **Given** an archive that omitted a slice, **When** the HTML is opened, **Then** that panel shows the omission reason already stored in the archive (#5131), not an empty dead tab.

### User Story 2 - Every control responds (Priority: P1)

**Independent Test**: Scripted clicks (Playwright/Selenium against the file URL or a local static server if file-origin blocks modules) on every tab and header button. Each click changes a measurable DOM state.

**Acceptance Scenarios**:

1. **Given** the viewer is open, **When** the user activates each `data-tab` / panel control, **Then** that panel becomes selected and its contents are displayed.
2. **Given** a filmstrip or action row, **When** the user clicks it, **Then** the detail/snapshot view updates to that item.
3. **Given** a control that should be disabled, **When** it is shown, **Then** it is visibly disabled and explained, not silently inert.

### User Story 3 - Solidification sweep (Priority: P1)

**Independent Test**: A checklist test walks every shipped panel (actions, filmstrip, network, console, artifacts/integrity, mobile if present). Missing panel = fail. Dead panel = fail.

**Acceptance Scenarios**:

1. **Given** the inventory of panels shipped by #4787, #4789, #4802, #4804, #5131, **When** the sweep runs against a fixture that contains data for each, **Then** every panel both opens and shows that data.
2. **Given** the same inventory, **When** a mutation removes the click handler or the data-binding for one panel, **Then** the new acceptance test fails.

## Edge Cases

- `file://` vs `http://127.0.0.1` origin. If ES modules or fetch of sidecar JSON fail on `file://`, either bundle the data into the HTML or document and test the supported open path (and make Allure/IntelliJ/MCP use that path).
- Do not require a live Playwright server for the SHAFT offline report.
- Do not clone the upstream Playwright viewer. Reuse remains pinned (#4824).
- Empty archive: show a clear empty state, not a clickable void.
- Large archives already have budgets (#4826). Viewer must still render what was kept.

## Functional Requirements

- **FR-001**: Opening the generated trace HTML shows the archive's collected data without a manual extra load that users cannot discover.
- **FR-002**: Every interactive control has a passing click/keyboard test.
- **FR-003**: Headless browser acceptance covers render + click + visible data, not only zip integrity.
- **FR-004**: Keep taxonomy labels valid (one primary, one lifecycle, at least one subsystem/module).

## Success Criteria

- **SC-001**: A reviewer can open the HTML from a known fixture and see actions/screenshots/network/console that exist in the archive.
- **SC-002**: Clicking each tab/button changes the view; a mutation that removes a handler fails CI.
- **SC-003**: #5131 integrity/omission surfaces still work; this ticket does not revert that design.

## Assumptions

- The archive writer is storing data correctly. The bug is viewer bind/load/event handling (confirm during implementation; if the writer is dropping fields, fix the writer in this ticket only when the viewer cannot show data that was never written — then add a writer test).
- IntelliJ/MCP should open the same HTML Allure attaches.

## Out of scope

- Replacing public E2E websites with local HTML (sibling ticket).
- Allure step-per-action reporting (sibling ticket).
- New viewer features or a JS framework rewrite.

## 📝 Additional Context

Found while opening the trace report during the #5204 investigation. Sibling of the local-HTML fixture ticket and the Allure one-step-per-action ticket.

### LINKED ISSUE #5213: [Bug]: Trace report / Allure trace viewer renders blank — unescaped ANSI bytes (0x1B) in embedded trace JSON break JSON.parse
### Summary

`SHAFT Trace Report.html` renders blank (header + tab bar only, all panels empty) because the trace JSON it embeds contains **unescaped ANSI escape bytes** (`0x1B`), which makes the report's own `JSON.parse` throw.

Since failure trace archives are written only on failure, this breaks exactly when the report is most needed.

### Root cause

The report embeds its payload and parses it client-side:

```html
<pre hidden id="trace-data">{ ...trace JSON... }</pre>
```
```js
const trace = JSON.parse(document.getElementById('trace-data').textContent);
```

Every panel (`#trace-summary`, `#action-list`, `#details`, `#tab-content`) is an empty container in the HTML, populated only by that script.

The `timeline` array captures SHAFT's own colorized console banners verbatim, so raw `0x1B` bytes land inside JSON strings:

```
"timeline": ["<ESC>[1;36m===============...", ..., "<ESC>[0m", ...]
```

(where `<ESC>` is a literal `0x1B` byte, not the two characters `\` + `u`)

Per RFC 8259 section 7, characters below `U+0020` MUST be escaped inside JSON strings (as ``). Because they are not, `JSON.parse` throws on the first line of the script and no panel is ever filled.

The same defect makes `shaft-trace.json` unreadable by strict parsers — Python's `json.load()` raises `Invalid control character`; only `strict=False` parses it.

### Reproduction

Any failing web test. Minimal case:

```java
public class TraceBugRepro {
    private SHAFT.GUI.WebDriver driver;

    @BeforeClass public void setUp() { driver = new SHAFT.GUI.WebDriver(); }

    @Test public void failOnPurposeToEmitATrace() {
        driver.browser().navigateToURL("https://example.com");
        driver.element().click(By.id("no-such-element-trace-bug-repro"));
    }

    @AfterClass(alwaysRun = true) public void tearDown() { if (driver != null) driver.quit(); }
}
```

Then open `target/shaft-traces/<testId>/shaft-trace.zip` -> `SHAFT Trace Report.html`.

**Observed:** the page renders the header and the ten tab buttons; summary, action list, details and every tab are empty. The browser console shows a `SyntaxError` from `JSON.parse`.

**Expected:** panels populated. The data is present and correct in the payload — only the parse fails.

Verifying the payload:

```python
import re, json
h = open('SHAFT Trace Report.html', encoding='utf-8').read()
p = re.search(r'<pre hidden id="trace-data">(.*?)</pre>', h, re.S).group(1)
print([(i, repr(c)) for i, c in enumerate(p) if ord(c) < 0x20 and c not in '\n\r\t'])
json.loads(p)   # JSONDecodeError: Invalid control character
```

On a freshly generated trace this prints 3 occurrences of `'\x1b'` and the parse fails at the first one.

### Environment

| | |
|---|---|
| SHAFT | `10.3.20260817` (also reproduced on `10.3.20260702`) |
| Coordinate | `io.github.shafthq:shaft-engine` |
| Java | 25 |
| Selenium | 4.45.0 |
| OS | macOS 15.1.1 (aarch64) |
| Browser | Chrome, headless |

### Impact

Affects any test whose timeline captures SHAFT's banner output — which appeared in every run we observed — so most failure traces are unopenable. The capture itself is fine; only the viewer breaks. Teams relying on the trace report for offline triage (attaching a single self-contained HTML to a ticket) cannot use it.

### Suggested fix

Escape control characters when serializing the trace payload, emitting `` rather than a raw byte. Stripping ANSI sequences from `timeline` entries before serialization would also make the timeline more readable.

As a local workaround, rewriting the embedded payload with control characters escaped makes the existing report render correctly with no other change.

MohabMohie:
## Summary

`FailureTraceReporter.escapeJson` now emits `\u00xx` for remaining RFC 8259 control characters (U+0000–U+001F), including ANSI ESC (`\u001b`). Timeline log lines with SHAFT color banners no longer make `shaft-trace.json` or the embedded `#trace-data` payload invalid for strict `JSON.parse`.

This is the same root as the blank/dead-click viewer: `JSON.parse` was the first statement of the inline script, so a throw skipped handler binding. No viewer rewrite.

Closes #5213
Closes #5209

Related to #5204
Related to #5212
Related to #5240 (sibling private `escapeJson` copies left in place)

## Checks

RED (before the `escapeJson` change):

```
mvn -pl shaft-engine "-Dtest=FailureTraceReporterTest#ansiEscapeInTimelineMustProduceStrictlyValidJson" test "-DheadlessExecution=true" "-Dallure.automaticallyOpen=false"
```

Surefire: 1 test, 1 failure: `Unescaped RFC 8259 control character U+001B at index 2387`

GREEN (after the change): same command. Surefire `TEST-com.shaft.tools.io.internal.FailureTraceReporterTest.xml`: tests=1 failures=0.

Nearest regression:

```
mvn -pl shaft-engine "-Dtest=FailureTraceReporterTest" test "-DheadlessExecution=true" "-Dallure.automaticallyOpen=false"
```

Surefire: tests=72 failures=0 errors=0 skipped=0. Did not run `TraceViewerBrowserAcceptanceTest` (no click/DOM/viewer JS change).

## Continuation

Head: `b86eb97358cce7a2e94b2bbbf06aee8b2108fe95`
State: implementation committed and pushed; awaiting independent review and PR-gate.
Blockers: none for this chunk. Native Memory `remember` degraded (`MemoryInvalidJsonl` blank `.memory/events.jsonl:1176`, already #5214, out of scope).
Next action: independent adversarial review, then arm auto-merge after the review/comment gate. Sibling serializers tracked as #5240.

### DIFF
--- a/shaft-engine/src/main/java/com/shaft/tools/io/internal/FailureTraceReporter.java
+++ b/shaft-engine/src/main/java/com/shaft/tools/io/internal/FailureTraceReporter.java
@@ -3092,12 +3092,26 @@ public final class FailureTraceReporter {
     }
 
     private static String escapeJson(String value) {
-        return value(value)
-                .replace("\\", "\\\\")
-                .replace("\"", "\\\"")
-                .replace("\n", "\\n")
-                .replace("\r", "\\r")
-                .replace("\t", "\\t");
+        String source = value(value);
+        StringBuilder escaped = new StringBuilder(source.length());
+        for (int i = 0; i < source.length(); i++) {
+            char c = source.charAt(i);
+            switch (c) {
+                case '\\' -> escaped.append("\\\\");
+                case '"' -> escaped.append("\\\"");
+                case '\n' -> escaped.append("\\n");
+                case '\r' -> escaped.append("\\r");
+                case '\t' -> escaped.append("\\t");
+                default -> {
+                    if (c < 0x20) {
+                        escaped.append(String.format("\\u%04x", (int) c));
+                    } else {
+                        escaped.append(c);
+                    }
+                }
+            }
+        }
+        return escaped.toString();
     }
 
     private static String escapeHtml(String value) {
--- a/shaft-engine/src/test/java/com/shaft/tools/io/internal/FailureTraceReporterTest.java
+++ b/shaft-engine/src/test/java/com/shaft/tools/io/internal/FailureTraceReporterTest.java
@@ -193,6 +193,41 @@ public class FailureTraceReporterTest {
         }
     }
 
+    @Test(description = "ANSI ESC in a timeline log line must be RFC 8259-escaped so strict JSON.parse can load the payload")
+    public void ansiEscapeInTimelineMustProduceStrictlyValidJson() throws Exception {
+        TestExecutionInfo failingInfo = info("ansiTimelineScenario", failure());
+        Path traceDirectory = FailureTraceReporter.traceDirectory(failingInfo);
+        try {
+            deleteDirectory(traceDirectory);
+            SHAFT.Properties.reporting.set().traceEnabled(true).traceMode("failure");
+            String ansiBanner = "\u001B[1;36m=============== Starting execution... ===============\u001B[0m";
+            FailureTraceReporter.attachOnFailure(failingInfo, ansiBanner, List.of());
+
+            try (ZipFile zip = new ZipFile(traceDirectory.resolve("shaft-trace.zip").toFile())) {
+                String json = readZipEntry(zip, "shaft-trace.json");
+                assertRfc8259ControlCharsEscaped(json);
+                JsonNode root = JSON.readTree(json);
+                Assert.assertEquals(root.path("timeline").path(0).asText(), ansiBanner, json);
+                Assert.assertTrue(json.contains("\\u001b") || json.contains("\\u001B"), json);
+
+                String html = readZipEntry(zip, "SHAFT Trace Report.html");
+                String marker = "<pre hidden id=\"trace-data\">";
+                int payloadStart = html.indexOf(marker);
+                Assert.assertTrue(payloadStart >= 0, html);
+                payloadStart += marker.length();
+                int payloadEnd = html.indexOf("</pre>", payloadStart);
+                String encoded = html.substring(payloadStart, payloadEnd);
+                String decoded = encoded.replace("&lt;", "<").replace("&gt;", ">").replace("&amp;", "&");
+                assertRfc8259ControlCharsEscaped(decoded);
+                Assert.assertEquals(JSON.readTree(decoded).path("timeline").path(0).asText(), ansiBanner);
+            }
+        } finally {
+            TraceEventRecorder.clear();
+            deleteDirectory(traceDirectory);
+            Properties.clearForCurrentThread();
+        }
+    }
+
     @Test(description = "Rendered trace JSON should redact sensitive values and keep source fallback frame")
     public void traceJsonShouldRedactSecretsAndKeepFallbackFrame() throws Exception {
         RuntimeException throwable = failure();
@@ -2597,6 +2632,16 @@ public class FailureTraceReporterTest {
     private static String readZipEntry(ZipFile zip, String entryName) throws Exception {
         try (var input = zip.getInputStream(zip.getEntry(entryName))) {
             return new String(input.readAllBytes(), StandardCharsets.UTF_8);
+        }
+    }
+
+    private static void assertRfc8259ControlCharsEscaped(String json) {
+        for (int i = 0; i < json.length(); i++) {
+            char c = json.charAt(i);
+            if (c < 0x20 && c != '\n' && c != '\r' && c != '\t') {
+                Assert.fail("Unescaped RFC 8259 control character U+" + String.format("%04X", (int) c)
+                        + " at index " + i);
+            }
         }
     }
 


MohabMohie (commented):
Independent review of `b86eb97358`. Zero blocking. Merge may arm after remaining checks.

F1 (non-blocking, vacuous-check): `ansiEscapeInTimelineMustProduceStrictlyValidJson` stays green if only NUL/BS escaping is reverted. Fixture is ESC-only. Ticket follow-up.

F2 (non-blocking, sibling-left): other private `escapeJson` copies still feed `#trace-data`. Already #5240.

F3 (non-blocking, wrong-width-check): Jackson parse of extracted HTML is not a browser `JSON.parse` of ANSI `#trace-data`. `TraceViewerBrowserAcceptanceTest` was not run with ESC.

PR #5377, opened 2026-08-25
test(installer): verify MCP initialize handshakes

MohabMohie:
## Summary
- replace shallow MCP startup checks with real JSON-RPC initialize handshakes
- validate exact protocol negotiation and server metadata
- cover both Memory and MemPalace launch contracts, including closed connections and malformed responses

## Validation
- `python3 -m unittest tests.scripts.test_chaos_engine_live_installer_acceptance`
- real fresh installation of the repository distribution
- real Memory and MemPalace initialize probes through the installed `.chaos-engine/tool.py`
- `git diff --check`

## Notes
A fresh installation succeeded before this change, confirming the reported local failure came from missing machine-local runtime state rather than the installer transaction. This change strengthens the live installer acceptance boundary so the same client-visible handshake failure cannot pass CI through `--help` probes.

### DIFF
--- a/scripts/ci/chaos_engine_live_installer_acceptance.py
+++ b/scripts/ci/chaos_engine_live_installer_acceptance.py
@@ -338,12 +338,39 @@ def probe_mcp(command: list[str], project: Path, *, popen=subprocess.Popen) -> N
         raise RuntimeError(
             f"MCP connection closed during initialize: {sanitize(stderr or stdout)}"
         ) from error
-    if process.returncode or response.get("id") != 1 or not isinstance(
-        response.get("result"), dict
+    result = response.get("result")
+    server = result.get("serverInfo") if isinstance(result, dict) else None
+    if (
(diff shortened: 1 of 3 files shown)


MohabMohie (commented):
Independent adversarial review completed in two bounded rounds against exact committed revisions.

Round 1 confirmed and fixed:
- F1 (wrong-width-check): malformed initialize results could pass.
- F2 (sibling-left): unit coverage did not prove both Memory and MemPalace remained wired into acceptance.

Round 2 confirmed and fixed:
- F1 (wrong-width-check): JSON boolean `true` compared equal to numeric ID `1`, and arbitrary protocol versions passed.

Final implementation requires JSON-RPC 2.0, an integer request ID, exact negotiated protocol, capabilities, and named/versioned server metadata. Both MCP siblings have explicit coverage. Scoped tests and live fresh-install handshakes pass.

PR #5436, opened 2026-08-28
fix(chaos-engine): align Codex lifecycle triggers

MohabMohie:
## Summary

- Align tracked Codex hooks with Codex 0.150.1 native trigger names: add PreCompact and remove unsupported PostToolUseFailure.
- Keep SHAFT tracked hooks on repository guard while generated portable hooks trust only owned portable guards.
- Preserve supported lifecycle behavior across Codex, Claude, and Claude-compatible Grok without inventing unsupported Codex triggers.
- Document one provider limit: Codex exposes neither PostToolUseFailure nor exec exit status in PostToolUse, so native failed-exec reflection has no exact Codex equivalent.

## Checks

- Focused RED/GREEN trigger, installer, matcher, fallback, native payload, and foreign-guard regressions passed.
- Consolidated balanced check: 322 passed, 2 skipped.
- Agent setup validation passed.
- Two terminal adversarial review rounds completed; findings fixed or documented as provider limits.
- Exact-head CI and final acceptance must pass before merge.

## Continuation

No deferred implementable work. Codex-only PermissionRequest, PostCompact, SubagentStart, and Interrupt remain unregistered because ChaosEngine owns no lifecycle duty at those trigger points; adding inert handlers would create false parity. Exact failed-exec reflection parity requires a future Codex failure event or exit-status field.


MohabMohie (commented):
Terminal adversarial review found two confirmed issues:

1. Native Codex 0.150.1 serializes shell/exec tool_response as a JSON string. Guard only classified dictionary failure metadata, so nonzero exits could skip failure recording and reflection checkpoints. Regression used a synthetic dictionary shape instead of native payload.
2. Generated installed locator preferred any repository scripts/agents/guard.py, allowing an adopter unrelated file at that path to bypass owned portable guard.

Both findings require fixes and exact-head revalidation before merge. Failed-tool reflection remains unproven until native payload regression passes.

MohabMohie (commented):
Round-2 terminal review found one blocking issue: pinned Codex source emits plain command output for native exec PostToolUse, while exit_code exists only in a separate internal code-mode result. The added JSON-object-string fixture was not native, so failed-shell reflection parity remained unproven. This provider limitation must be stated honestly; fabricated decoding and claims must be removed before merge. Foreign repository guard isolation passed.

PR #5800, opened 2026-09-15
fix(allure): always use managed Allure 3 CLI (#5798)

### LINKED ISSUE #5798: Allure: decide 2.x user-CLI compat vs always-portable Allure 3 after BOM 3.0
## Context

Opened from Dependabot PR https://github.com/ShaftHQ/SHAFT_ENGINE/pull/5793 (`io.qameta.allure:allure-bom` 2.35.5 → **3.0.0**).

That bump is the **Java adapter/BOM** (listeners, model, lifecycle). Report **HTML generation** still shells out to an Allure **CLI** via `AllureManager` (`shaft-engine/.../AllureManager.java`), which is a separate resolution path from the Maven BOM.

## Current behavior (investigate before changing)

`AllureManager` already documents a dual path:

1. **Default today:** `allure.forceConfiguredCliVersion=true`  
   - Skips system `allure` on `PATH` (including Allure 2.x compatibility detection).  
   - Uses managed Allure 3 only: `npx --yes allure@<SHAFT.Properties.internal.allure3Version()>`  
   - If Node/`npx` missing → downloads portable Node.js under `~/.m2/repository/nodejs/` and retries.

2. **When `allure.forceConfiguredCliVersion=false`:**  
   - If `PATH` has `allure` reporting **2.x** → Allure 2 compatibility mode (`cachedAllureCommandPrefix = "allure"`, different generate command shape).  
   - If `PATH` has a 3.x/`unknown` binary → may use it (unless force requires exact version match).  
   - Else fall through to npx / portable Node.

Unit coverage already exists for “build Allure 3 and Allure 2 commands” (`AllureManagerCoverageUnitTest.getCommandToCreateAllureReportShouldBuildAllure3AndAllure2Commands`).

## Problem to decide

With the **Java stack on Allure 3.0.0**, keeping “prefer/activate user-installed Allure **2.x** CLI” is increasingly a footgun:

- Results written by allure-java 3 may not round-trip cleanly through an Allure 2 CLI.
- Operators with an old global `allure` and `forceConfiguredCliVersion=false` get a silent compat path that may look “supported” but is fragile after #5793.
- We already have a portable Allure 3 path (npx + portable Node) — the question is whether **user PATH binaries** should remain first-class.

## Options (pick one in delivery)

| Option | Summary |
| --- | --- |
| **A. Keep 2.x PATH compat** | Leave dual mode; add/strengthen tests that a real/fake Allure 2.x on PATH still generates when `forceConfiguredCliVersion=false`, and document that BOM 3 + CLI 2 is best-effort only. |
| **B. Ignore user-installed binaries** | Always use managed Allure 3 (`npx` / portable Node); remove or hard-deprecate PATH/`allure` 2.x activation (keep property only as no-op or fail-closed). |
| **C. Maven-provisioned portable CLI** | Add a dependency/plugin (or engine bootstrap) that vendors/installs a pinned Allure 3 CLI binary into the project/user cache without relying on npx/network at report time; never call user `PATH` allure. |

**Recommendation (non-binding):** lean **B**, with optional later **C** if offline/CI air-gap needs stronger guarantees than npx. Keeping **A** requires an explicit compatibility matrix test in CI (BOM 3 results × Allure 2 CLI) — do not claim support without that.

## Acceptance

1. Written decision in the PR/issue (A/B/C) with rationale tied to #5793.
2. Tests matching the decision:
   - If A: regression that 2.x PATH mode still builds/runs generate command under `forceConfiguredCliVersion=false`.
   - If B/C: assert managed 3 path wins even when a stub Allure 2 is first on `PATH`; update docs/properties javadoc.
3. Docs: `Allure.java` / user guide note on `allure.forceConfiguredCliVersion` and what operators should install (or not).
4. Link back to #5793 in the delivery PR body.

## Out of scope

- Reverting the allure-bom 3.0.0 bump itself (tracked by #5793).
- ChaosEngine installer work.

MohabMohie:
## 📋 Description

**Decision: Option B** — drop Allure 2 CLI support completely after `allure-bom` 3.0.0 (#5793).

SHAFT report generate/open/serve now **always** use the managed Allure 3 path already provisioned by the engine (`npx --yes allure@<allure3Version>`, or portable Node under `~/.m2/repository/nodejs/`). User-installed `allure` on `PATH` (including Allure 2.x) is never selected.

`allure.forceConfiguredCliVersion` remains as a **deprecated always-on alias** (default `true`). Setting it `false` is a no-op that logs a discrete deprecation note; PATH allure stays ignored.

## 🔗 Related Issue(s)

- Closes #5798
- Follows #5793 (`821e9d3` allure-bom 3.0.0)

## 🗂️ Type of Change

- [x] 🚀 New feature (non-breaking change that adds functionality)
- [x] 🔨 Refactor / code cleanup

Release-note classification label: `enhancement`

## ✅ Pre-Submission Checklist

- [x] Behavior changes have focused automated coverage.
- [x] Affected tests pass: `mvn -pl shaft-engine -am test -Dtest=AllureManagerUnitTest,AllureManagerCoverageUnitTest -Dallure.automaticallyOpen=false -DheadlessExecution=true` (62 tests, 0 failures).
- [x] I have not broken backward compatibility for configs that already set `forceConfiguredCliVersion=true` (still the effective/default behavior).

## 🧪 How to Test

1. Run the focused unit tests above.
2. Optionally place a stub `allure` 2.x binary early on `PATH` and confirm resolution still uses `npx --yes allure@<version>` (covered by unit tests).

## Evidence

```
Tests run: 62, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

## Documentation Site

- Documentation not required because: no dedicated Allure user-guide page for this switch; `Allure.java` property javadoc documents the migration. Operators do not need to install a global Allure CLI.

## 📝 Additional Notes

- **Option C** (Maven-provisioned / pinned Allure 3 CLI binary for stronger offline/air-gap guarantees) remains a later optional follow-up — out of scope for this PR.
- Removed `cachedIsAllure2` and Allure 2 generate/serve command shapes from `AllureManager`.
- HTML attachment-preview CSS comments that mention Allure 2 class names are unchanged (they describe DOM selectors, not CLI support).


MohabMohie:
Deferred Option C (Maven-pinned Allure 3 CLI) tracked as https://github.com/ShaftHQ/SHAFT_ENGINE/issues/5801

